PR #3, opened 2020-09-14
Explain the meta processing logic in some detail.

jeremyroman:
@domenic PTAL?

I've tried to capture my current thinking about how the details of this ought to work.


domenic (approved):
LGTM. It might just be me, but I had an intuition that you'd want this before anything that causes a resource fetch (like `<link>`). It could be worth explaining why that's not necessary.

The answer is, IIUC, because this prescanning will happen before any resource fetching, so if the meta is not found, then we bail out without performing any problematic resource fetches.

jeremyroman:
Expanded the discussion of `<link>` a little more; ptal?

domenic (approved):
Haha, I would just omit bgsound, but LGTM :)

PR #4, opened 2020-10-05
Initial draft of outlining the larger problem space

domenic:
The biggest uncertainty I have about this is how I've doubled down on prerendering, whereas Jeremy's original doc seems to want to treat prerendering and prefetching on almost-equal footing. Thoughts welcome.

I highly recommend reviewing this with the "Split" diff mode, or better yet looking at the [rendered output](https://github.com/domenic/alternate-loading-modes/blob/bigger/README.md), since the README.md diff is just a mess.

/cc @domfarolino @yoavweiss @kinu @ianvollick

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,15 @@ The tradeoff is that we now require opt-in from the referring page. The user age
 
 Cross-origin prerendering, on the other hand, can be done without such triggers, because it is so much more constrained and requires opt-in from the content itself. The tradeoff is that, until the ecosystem starts preparing itself for prerendering via the opt-in and associated upgrade code, such prerenders are unlikely to succeed.
 
+Here's a summary:
+
+|                          |Opt-in location  |Restrictions on disruptive APIs |Restrictions on credentials/storage/etc. |
+|--------------------------|-----------------|--------------------------------|-----------------------------------------|
+|Same-origin prerendering  |Referring page   |Yes                             |No                                       |
+|Cross-origin prerendering |Destination page |Yes                             |Yes                                      |
+
+_Aside: it's probably safe to also allow same-origin prerendering with only a destination-side opt-in, as long as all of the same restrictions are applied (e.g., no credentials or storage access). But, this complicates the model a good deal, for both implementers and web developers. For now, we're concentrating on the model described above._
+
 Finally, we'll note that browser-initiated prerenders fall somewhere in between these cases. In particular, the user typing `https://example.com/` in the URL bar, even before they press <kbd>Enter</kbd>, might serve as a reasonable prerender trigger, and perhaps even the prerendering could be done with credentials. The need to prevent user annoyance is still present, so the prerendering browsing context concept is important. But, what if the user types `https://example.com/logout`? Our thinking is still evolving in this area.
 
 ## Prefetching
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ We envision prerendering having several related pieces:
 
 * [**Prerendering fetching modes**](./fetch.md), which modify the way in which cross-origin documents and subresources are fetched in order to preserve privacy and avoid side effects.
 
-  Closely related to the previous bullet, this covers the mechanics of how a document is fetched in order to check for the opt-in, and provide it with no identifying information. The most obvious technique here is omitting credentials, but others could include using a proxy server (for IP privacy), or using a previously-fetched response in "memory cache".
+  Closely related to the previous bullet, this covers the mechanics of how a document is fetched in order to check for the opt-in, and provide it with no identifying information. The most obvious technique here is omitting credentials, but one could also consider using a proxy server (for IP privacy), or using a previously-fetched response in "memory cache".
 
 * [**Prerendering browsing contexts**](./browsing-context.md), which are special browsing contexts that are not displayed to the user, and within which content is constrained to not perform disruptive or side-effecting operations.
 
--- a/README.md
+++ b/README.md
@@ -58,7 +58,9 @@ Finally, we'll note that browser-initiated prerenders fall somewhere in between 
 
 ## Prefetching
 
-Although these explainers focus largely on prerendering, we expect some of the work they produce to be useful for _prefetching_ as well. TODO more.
+Although these explainers focus largely on prerendering, we expect some of the work they produce to be useful for _prefetching_ as well. Prefetching currently exists in [`<link rel="prefetch">`](https://w3c.github.io/resource-hints/#dfn-prefetch), but as with prerendering, it is underspecified, and its current implementations have potential privacy issues which will require some work to address.
+
+In particular, the [triggers](./triggers.md) and [opt-in](./opt-in.md) can be designed in a generic way, so that they can also be used to trigger and opt-in to prefetching (of documents, in particular). Similarly, the [prerendering fetching infrastructure](./fetch.md) will likely be used for modernized prefetching.
 
 ## Example scenarios
 
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ If the opt-in is present, then the resulting document is loaded into a cross-ori
 function afterPrerendering() {
   // grab user data from cookies/IndexedDB
   // update the UI
-  // maybe ask for camera access
+  // maybe ask for notifications access
 }
 
 if (!document.loadingMode || document.loadingMode.type === 'default') {
@@ -172,6 +172,6 @@ newsArticleLink.onclick = async e => {
 
   e.preventDefault();
   await animateToFullViewport(newsArticlePortal);
-  newsArticlePortal.activate()
+  newsArticlePortal.activate();
 };
 ```
--- a/README.md
+++ b/README.md
@@ -131,6 +131,7 @@ If the opt-in is present, then the resulting document is loaded into a cross-ori
 function afterPrerendering() {
   // grab user data from cookies/IndexedDB
   // update the UI
+  // maybe ask for camera access
 }
 
 if (!document.loadingMode || document.loadingMode.type === 'default') {
@@ -142,6 +143,15 @@ if (!document.loadingMode || document.loadingMode.type === 'default') {
     }
   });
 }
+```
+
+Alternately, if the page only cares about storage access (and not other facets of prerendering, such as ability to autoplay or trigger permission prompts), they could use a [proposed storage access API extension](https://github.com/privacycg/storage-access/issues/55):
+
+```js
+document.storageAccessAvailable.then(() => {
+  // grab user data from cookies/IndexedDB
+  // update the UI
+});
 ```
 
 ### Cross-origin news aggregator with previews
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ The space is full of challenges which have not been comprehensively tackled: mos
 * avoiding user-visible side effects and annoyances from running script on the destination site; and
 * fulfilling the privacy objectives of the user and the referring site.
 
-This repository contains a set of explainers and (eventually) specifications which, combined, give a rigorous model for performing such prerendering and prefetching of content, in an interoperably-implementable way.
+This repository contains a set of explainers and (eventually) specifications which, combined, give a rigorous model for performing such prerendering of content, in an interoperably-implementable way. Each piece is designed to be composable and reusable; for example, [some contribute to prefetching](#prefetching), independent of prerendering, and the [opt-in](./opt-in.md) is designed to be usable by other types of alternate loading modes, such as [fenced frames](https://github.com/shivanigithub/fenced-frame/).
 
 ## Pieces of the solution
 
@@ -58,7 +58,7 @@ Finally, we'll note that browser-initiated prerenders fall somewhere in between 
 
 ## Prefetching
 
-Although these explainers focus largely on prerendering, we expect some of the work they produce to be useful for _prefetching_ as well. Prefetching currently exists in [`<link rel="prefetch">`](https://w3c.github.io/resource-hints/#dfn-prefetch), but as with prerendering, it is underspecified, and its current implementations have potential privacy issues which will require some work to address.
+Although these explainers focus largely on prerendering, we expect some of the work they produce to be useful for _prefetching_ as well. Prefetching currently exists in [`<link rel="prefetch">`](https://w3c.github.io/resource-hints/#dfn-prefetch), but as with prerendering, it is underspecified, and its current implementations have potential privacy issues for cross-origin prefetching, which will require some work to address.
 
 In particular, the [triggers](./triggers.md) and [opt-in](./opt-in.md) can be designed in a generic way, so that they can also be used to trigger and opt-in to prefetching (of documents, in particular). Similarly, the [prerendering fetching infrastructure](./fetch.md) will likely be used for modernized prefetching.
 
--- a/README.md
+++ b/README.md
@@ -1,168 +1,156 @@
-# Alternate Loading Modes
+# Prerendering, revamped
 
-(Thanks to @domenic, @kinu, @yoavweiss, @clelland, and others, for their thoughts which contributed to this proposal.)
+In order to making the experience of loading on the web faster, user agents employ prefetching and prerendering techniques. However, they have historically been [underspecified](https://w3c.github.io/resource-hints/#prerender) and [inconsistently implemented](https://caniuse.com/link-rel-prerender).
 
-In order to making the experience of loading on the web faster, user agents employ prefetching and prerendering techniques. However, making cookies and other credentials available to the origin server or script may be inconsistent with the privacy objectives of the user or of the referring site.
+The space is full of challenges which have not been comprehensively tackled: most notably,
 
-First, consider the __fetch__ of the resource. User agents would ideally prefetch the content in a way that does not identify the user. For example, the user agent could:
+* avoiding user-visible side effects and annoyances from running script on the destination site; and
+* fulfilling the privacy objectives of the user and the referring site.
 
-* send a request without credentials (e.g., no `Cookie` or `Authorization` request header)
-* establish the connection from a different client IP address (e.g., using a proxy server or virtual private network, if available)
-* use a previously fetched response, including one previously fetched by a third party if it can be authenticated
+This repository contains a set of explainers and (eventually) specifications which, combined, give a rigorous model for performing such prerendering and prefetching of content, in an interoperably-implementable way.
 
-However, because this (intentionally) obscures the user's identity, the response document cannot be personalized for the user. If it is used when the user navigates, the user will notice that they are not logged in (even if they should be), and other surprising behavior. A page designed with this in mind could "upgrade" itself when it loads, by personalizing the page based on data in unpartitioned storage and by fetching personalized content from the server.
+## Pieces of the solution
 
-Second, consider __prerendering__ the page. User agents would ideally allow HTML parsing, subresource fetching, and script execution in a way that does not identify the user or cause user-visible annoyance. For example, the user agent could:
+We envision prerendering having several related pieces:
 
-* apply mitigations as above to subresource and scripted fetches
-* deny scripted access to unpartitioned storage, such as cookies and IndexedDB
-* deny permission to invoke `window.alert`, autoplay audio, and other APIs inappropriate at this time
+* [**Prerender triggers**](./triggers.md), where a referring site indicates to the user agent what content can or should be prerendered.
 
-In this case, not only is the HTML resource not personalized, but script will observe restrictions that would not ordinarily apply until navigation actually occurs. A page designed with this in mind could tolerate this at prerender time, and "upgrade" itself on navigation by accessing storage or fetching from the network.
+  Today, we have the underspecified `<link rel="prerender">` element to provide these indications. Although something like this (probably with a different name, to avoid compatibility issues?) might be a good starting point, there are other points in the design space worth considering. For example, allowing the referring site to provide more blanket permission to prerender lets the user agent use its own heuristics.
 
-Since existing web pages are unlikely to behave well with these restrictions today, and it is impractical for user agents to distinguish such pages, we propose a lightweight way for a page to declare that it is prepared for this and will, if necessary, upgrade itself when it gains access to unpartitioned storage and other privileges.
+* [**Prerendered content opt-in**](./opt-in.md), which allows pages to opt in to being prerendered by other cross-origin pages.
 
-There has been previous discussion along these lines in [w3c/resource-hints#82](https://github.com/w3c/resource-hints/issues/82#issuecomment-536492276). (It also proposes a new `prenavigate` hint; defining triggers for these loading modes is not yet part of this proposal.)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,168 +1,156 @@
...
+Each of these pieces is connected in various ways. However, we think they're decoupled enough that we will start by developing them as separate explainers and spec documents, cross-linking to each other as appropriate. It's also possible to implement only a subset of these, if a user agent is only interested in certain [scenarios](#example-scenarios).
 
-A server could defer non-idempotent behavior such as impression counting,  or it could reject prerender requests altogether without needing to do the full work to generate a response (since it need only inspect the request headers).
+## Same-origin vs. cross-origin prerendering
 
-> TODO: Should we also/instead send `Sec-Purpose`? Should `Sec-Fetch-Mode` differ?
+When prerendering same-origin content, many fewer constraints are necessary. Because there is no privacy concern, we can use normal fetching modes, and thus do not need an opt-in from the prerendered page. And the prerendering browsing context becomes simpler. Thus, the majority of work for specifying same-origin prerendering is in the prendering triggers, the restrictions on disruptive APIs, and the transition to a normal top-level browsing context.
```
mfalken, line 40: typo: prendering

README.md
```diff
@@ -1,168 +1,156 @@
...
+function afterPrerendering() {
+  // grab user data from cookies/IndexedDB
+  // update the UI
+}
 
-`Sec-Required-Document-Policy` would not replace the need for a [request header](#request-header), since this is settable by content that is not genuinely prerendering, e.g. via `<iframe policy>`.
+if (!document.loadingMode || document.loadingMode.type === 'default') {
+  afterPrerendering();
+} else {
+  document.addEventListener('loadingmodeechange', () => {
```
mfalken, line 130: typo(?): modeechange

README.md
```diff
@@ -1,168 +1,156 @@
...
 
-## Risks
+This simplicity benefits web developers as well, as they don't need to do the upgrade-from-uncredentialed dance which is necessary in cross-origin cases.
 
-__Credentialed requests__ allow fetching the right resource for the user, but the web is increasingly moving toward a model whereby an origin cannot trigger credentialed requests to another user except as part of a committed action (i.e., a navigation). Forward-looking prerendering should prefer to use uncredentialed requests to reduce the risk of identifying the user when doing so is undesirable, and should also deny access to unpartitioned storage under those circumstances.
+The tradeoff is that we now require opt-in from the referring page. The user agent cannot just heuristically prefetch or prerender any same-origin links that it sees; doing so would have bad consequences for links like `<a href="/logout">`.
```
mfalken, line 44: I don't quite understand why opt-in is required for same-origin. If there's no opt-in, could it make sense to do automatic prerendering for a same-origin link and just use the same restrictions as cross-origin prerendering?

I guess we assume if the site bothered to put opt-in on the content response, it would have bothered to put opt-in on the referring page if it were worthwhile, so there's no need to try automatic prerendering?

README.md
```diff
@@ -1,168 +1,156 @@
...
+or the HTML `<meta>` element
 
-This strongly indicates an opt-in in the HTML document, such as a `<meta>` tag, attribute of the `<html>` element, or other affordance entirely within the HTML response body.
+```html
+<meta http-equiv="Supports-Loading-Mode"
+      content="uncredentialed-prerender">
+```
 
-### Document policy
+As in the last example, based on its heuristics or informed by prerender triggers, the browser can attempt to prerender these linked-to news articles. Since they are cross-origin, however, the process is more restricted. The initial fetch, as well as any subresource fetches, are performed without credentials. If the response that comes back does not have the opt-in, then the result is discarded.
```
mfalken, line 117: Is opt-in needed if there were no credentials when the prerender started?

browsing-context.md
```diff
@@ -0,0 +1,28 @@
...
+
+Script can observe and react to changes in the loading mode, if applicable. For example, it can observe a change from `uncredentialed-prerender` to `default` when a user navigates to the prerendered document. This can be used to defer personalization and logic which are not necessary for prerendering.
+
+```js
+function afterPrerendering() { /* ... */ }
+
+if (!document.loadingMode || document.loadingMode.type === 'default') {
+    afterPrerendering();
+} else {
+    document.addEventListener('loadingmodeechange', () => {
```
mfalken, line 21: typo(?): modeechange

mfalken (commented):
This is great, just had some non-blocking comments as I read it.

README.md
domenic: I think you're right; we could do same-origin prerendering with only destination-side opt-in. (Would we also omit credentials? Hmm.)

Since this section is supposed to be comprehensive, I think I should mention that, even if we end up deciding to build something that sticks to the simpler rule I propose here. (I.e., sticks to triggering-page opt-in for same-origin, destination-page opt-in for cross-origin.)

README.md
domenic: This is still unclear to me. There's talk of "adaptive prerendering" which I have a hard time understanding, and could use some help incorporating into this document.

In particular, a model that tries to prerender without opt-in runs into the following issues:

- Does that break impression tracking?
- What happens if the page tries to write to storage?
  - If it fails, then we're likely to break pages. So an opt-in would be better.
  - If it goes to partitioned storage, then we have to figure out how to join the partitioned storage and the unpartitioned storage after transition to a real top-level browsing context. Which seems basically impossible to do automatically. So an opt-in, plus custom page logic to handle the transition, makes more sense.

README.md
jeremyroman: Destination side opt-in only is a problem because you don't know if the GET will have side effects, which is especially acute in the credentialed (same-origin) case.

domenic: Why would a side-effecting page opt-in?

jeremyroman: You cannot tell if a destination page has opted in until after you have already fetched it.

README.md
```diff
@@ -1,168 +1,156 @@
...
+  Today, we have the underspecified `<link rel="prerender">` element to provide these indications. Although something like this (probably with a different name, to avoid compatibility issues?) might be a good starting point, there are other points in the design space worth considering. For example, allowing the referring site to provide more blanket permission to prerender lets the user agent use its own heuristics.
 
-Since existing web pages are unlikely to behave well with these restrictions today, and it is impractical for user agents to distinguish such pages, we propose a lightweight way for a page to declare that it is prepared for this and will, if necessary, upgrade itself when it gains access to unpartitioned storage and other privileges.
+* [**Prerendered content opt-in**](./opt-in.md), which allows pages to opt in to being prerendered by other cross-origin pages.
```
jeremyroman, line 20: It does confuse the issue here a bit, but FWIW the reason the opt-in is described a little funny is to allow room for fenced frames to have a separate loading mode with different restrictions if they need that in the future.

README.md
```diff
@@ -1,168 +1,156 @@
-# Alternate Loading Modes
+# Prerendering, revamped
```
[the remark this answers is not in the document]
jeremyroman: One option is to use a more general term like "speculation" instead, that captures dns-prefetch, preconnect, prefetch, and prerender. Obviously some pieces are prerender-specific. But obviously the more general a term we use the less clear it will be at first glance what we mean.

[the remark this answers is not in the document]
jeremyroman: I also think the ladder of speculation is interesting. If there are too many links that would be okay to prerender and we're not sure, I think we should be free to prefetch the 3 most likely options instead, for instance.

fetch.md
```diff
@@ -0,0 +1,11 @@
+# Prerendering fetching modes
+
+Placeholder. TODO.
+
+Some text from Jeremy's doc:
+
+First, consider the __fetch__ of the resource. User agents would ideally prefetch the content in a way that does not identify the user. For example, the user agent could:
```
jeremyroman, line 7: I'm not sure about separating this so much from the discussion of what the consequences of this are. opt-in.md still discusses it a little, but in a way that might be too cursory to be useful to understand without clicking through.

README.md
jeremyroman: Probably my typo. So hard to spot them when allthewordsruntogether.

README.md
```diff
@@ -1,168 +1,156 @@
...
+  "allow": [
+    {
+      "action": "prerender",
+      "url_patterns": ["/**"]
+    }
+  ],
+  "disallow": [
+    {
+      "action": "prerender",
+      "urls": ["/logout"]
```
jeremyroman, line 72: Possibly a signal that we need to iterate more on semantics, but the current processing model I have described in the doc this is drawn from would interpret this as prohibiting prerendering but not prefetching (allow if action <= prerender; disallow if action >= prerender, thus action==prefetch clears both).

Changing "prerender" to "prefetch" in the disallow rule fixes this.

README.md
```diff
@@ -1,168 +1,156 @@
...
+```json
+{
+  "action": "prerender",
+  "selectors": [".high-likelihood-prerender"],
+  "likelihood": "high"
+}
+```
 
-It might be possible for the user agent to augment author declarations with a list of origins or documents known to behave well, as a browser feature. This would depend on identifying such sites in a fairly reliable way and providing a mechanism for users to reload if they observe brokenness. However, such a mechanism would have limitations and would necessarily not behave the same in all browsers. The web platform should provide a way to more predictably get the desired behavior.
+and then decorate certain `<a>` elements with `class="high-likelihood-prerender"`.
```
jeremyroman, line 96: Noting, of course, that it would be totally legitimate to split off selector support into a followup, if we're worried about it. (My doc describes it because I do think it might be useful, and because it demonstrates extensibility.)

README.md
jeremyroman: I'm also unsure how we want to reason about adaptive prerendering long-term, and what we can/cannot do without modification.

@kinu

README.md
```diff
@@ -1,168 +1,156 @@
...
-`Sec-Required-Document-Policy` would not replace the need for a [request header](#request-header), since this is settable by content that is not genuinely prerendering, e.g. via `<iframe policy>`.
+if (!document.loadingMode || document.loadingMode.type === 'default') {
+  afterPrerendering();
+} else {
+  document.addEventListener('loadingmodeechange', () => {
+    if (document.loadingMode.type === 'default') {
+      afterPrerendering();
+    }
+  });
+}
```
jeremyroman, line 135: Noting, of course, that I would prefer to make something like privacycg/storage-access#55 a default choice for cases like this that really just care about whether they can see their storage yet. Pending ongoing discussions about how storage access transitions will work in other contexts, also on that repo.

The detailed loadingmodechange API can hopefully be used mainly by stuff that actually cares about why they can't touch storage and other restrictions that apply in a prerendered browsing context. Very bikesheddable.

opt-in.md
```diff
@@ -0,0 +1,142 @@
...
+
+Similarly, [prerendering browsing contexts](./browsing-context.md) allow HTML parsing, subresource fetching, and script execution, but such actions are restricted to avoid identifying the user or causing user-visible annoyance.
+
+Pages designed with these restrictions in mind can "upgrade" themseles when they load, by personalizing based on data in unpartitioned storage and by fetching personalized content from the server. But existing web pages are unlikely to behave well with these restrictions today. (And, it is impractical for user agents to distinguish such pages.)
+
+As such, we propose a lightweight way for a page to declare that it is prepared for such prerendering, and will, if necessary, upgrade itself when it gains access to unpartitioned storage and other privileges.
+
+There has been previous discussion along these lines in [w3c/resource-hints#82](https://github.com/w3c/resource-hints/issues/82#issuecomment-536492276). (It also proposes a new `prenavigate` hint; defining triggers for these loading modes is not yet part of this proposal.)
+
```
jeremyroman, line 12: I'm blindly assuming everything below this is unmodified since I don't see an easy way to diff it...

README.md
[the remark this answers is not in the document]
domenic: My take was that we should focus this around prerendering, and say that we might capture some value for prefetching etc. along the way. But it's easier to paint a coherent picture as to why these all belong as part of the same effort, if that effort is prerendering.

opt-in.md
domenic: Oh interesting; locally it counted as a rename. Anyway, your assumption is correct, apart from the new Acknowledgments section,

README.md
domenic: So we could do same-origin destination-side only opt-in as long as we applied the same restrictions that we do for cross-origin, i.e. no credentials/storage access/etc. Right? I'll push a commit mentioning that possibility, but saying that we're not currently pursuing it.

README.md
domenic: Yeah, in general I'm not sure about these "scenarios" sections. They're currently more detailed than the sub-documents they link to. Maybe when we have the full explainer, we can change this to reference a "potential feature", linking to an explanation of it as being an extension point.

fetch.md
domenic: The idea is that this will eventually grow to a more detailed explanation of the changes and how they flow through the system. As such, these seem like the kernel of such a detailed description, and I moved them here.

My intention for opt-in.md is that you're supposed to arrive at that document taking for granted that an opt-in is needed, and if you want to find out why, then you'll have to click through to fetch.md or maybe README.md. (Probably some of https://github.com/WICG/portals#privacy-threat-model-and-restrictions needs to make its way here...)

README.md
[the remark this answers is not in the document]
domenic: I've pushed a commit that expands the "Prefetching" section to explain this a bit more.

README.md
domenic: Hmm, what do you mean? In particular, the opt-in description doesn't seem strange to me.

README.md
kinu: I'm not sure if mentioning same-origin destination-side opt-in case is important or necessary at this point, it may just confuse readers?  One thing that's clear is that we'll start with referrer side API (either in implicit or explicit)

kinu: Ah I saw the new text below. I'm good with the current framing.

README.md
```diff
@@ -1,168 +1,177 @@
...
 
-## Risks
+## Prefetching
 
-__Credentialed requests__ allow fetching the right resource for the user, but the web is increasingly moving toward a model whereby an origin cannot trigger credentialed requests to another user except as part of a committed action (i.e., a navigation). Forward-looking prerendering should prefer to use uncredentialed requests to reduce the risk of identifying the user when doing so is undesirable, and should also deny access to unpartitioned storage under those circumstances.
+Although these explainers focus largely on prerendering, we expect some of the work they produce to be useful for _prefetching_ as well. Prefetching currently exists in [`<link rel="prefetch">`](https://w3c.github.io/resource-hints/#dfn-prefetch), but as with prerendering, it is underspecified, and its current implementations have potential privacy issues which will require some work to address.
```
kinu, line 61: Large missing part in prefetch is for cross-origin navigation (that is what I'm hoping that this one can just address/subsume), while same-origin subresource prefetches are relatively in a okay state (though there're still various underspecified parts there too, yes)

README.md
kinu: "action" can probably have multiple actions?  I feel that'd be less confusing, I'm a bit concerned about introducing some precedence / ordering between the modes.

kinu (commented):
This is looking great. I'm still reading but pushing some comments

README.md
```diff
@@ -1,168 +1,177 @@
...
+{
+  "allow": [
+    {
+      "action": "prerender",
+      "url_patterns": ["/**"]
+    }
+  ],
+  "disallow": [
+    {
+      "action": "prefetch",
```
kinu, line 82: Can "action" have multiple actions? (Related to the resolved comment thread, but I'm a bit concerned about introducing ordering or inclusion relationships between multiple actions)

README.md
kinu: Note, in a world where everyone opts-in we don't need this.

My reasoning is that it shouldn't have a tough conflict as far as it can go without requiring cookies (while the site can still opt-out and drop the request). For actions like prefetching this is even more plausible.

No storage writes should happen, they will be deferred as much as possible, or if sync access happens the prerender can bail out.

Exception is the 1p cookie, they will be written back and merged, and this is possible (and we have code) because we're talking about no-storage cases.

kinu: Adaptive one is an exploration of what we can do *without* conflicting with what we want to be doing.  Therefore there shouldn't be many that need to be incorporated explicitly for adaptive cases, while, one exception is the first main resource response.  If UA has no cookies it doesn't feel that the necessity to have response opt-in is a bit slim.

README.md
jeremyroman: Right, but I'd prefer to err on the side of not sending a GET unless we have *some* sort of indication that it's not going to have side effects. Being uncredentialed lowers the bar, but I'm not sure I'm ready to lower the bar to zero.

fetch.md
```diff
@@ -0,0 +1,11 @@
...
+
+* send a request without credentials (e.g., no `Cookie` or `Authorization` request header)
+* establish the connection from a different client IP address (e.g., using a proxy server or virtual private network, if available)
+* use a previously fetched response, including one previously fetched by a third party if it can be authenticated
```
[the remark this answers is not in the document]
jeremyroman: This is a handwave about caching, including SXG caching.

README.md
jeremyroman: Mainly it's a little abstract because it wants to allow the possibility of other loading modes with restrictions other than those for prerendering, in the future.

README.md
jeremyroman: It could (though alternatively you could just repeat the whole rule).

I would really like to keep some form of ordering because I really think we should be able to prefetch but not prerender if we have network bandwidth available but don't want to waste CPU/memory, or if we're uncertain between various possible outgoing links. It's definitely possible to make this less footgun-y.

kinu: Sure, ordering is a hard problem in general.  In my ideal mental model prerendering can be decomposed into { connect, fetch-main, parse-html, fetch-sub, run-javascripts, render ... } or something, but it may be too detailed.

jeremyroman: Yeah, @domenic and I are discussing some of the options here in the internal doc (which will be probably published to this repo shortly).

README.md
kinu: Sorry, I had to struggle to come up with a possible alternative text (and the reasonable behavior).  I think I'm also okay to drop this eventually, but not yet fully comfortable saying that "we require opt-ins for all responses, otherwise UA does nothing".

What do we think if we tweak this like following:

'... are performed without credentials.  This means that the target site may get uncredentialed requests even when there are credentials stored in UA. If the response that comes back does not have the opt-in in such cases, the result is discarded.'

kinu:
Added one more suggestion / possible alternative text for the opt-in requirement, but otherwise this LGTM. Thanks!

README.md
```diff
@@ -1,168 +1,177 @@
...
+```html
+<meta http-equiv="Supports-Loading-Mode"
+      content="uncredentialed-prerender">
+```
 
-The largest drawback here is one of adoption. In order to make as much content as possible available for uncredentialed prerendering, we would like to make it as easy as possible for authors to mark eligible content. We have heard from developers that many of them find it much easier to deploy changes that only affect content than changes which also require server behavior changes, even relatively straightforward ones. For example, these may be managed by different teams or not be possible at all. One key example here is that GitHub Pages doesn't allow users to set response headers.
+As in the last example, based on its heuristics or informed by prerender triggers, the browser can attempt to prerender these linked-to news articles. Since they are cross-origin, however, the process is more restricted. The initial fetch, as well as any subresource fetches, are performed without credentials. If the response that comes back does not have the opt-in, then the result is discarded.
```
mfalken, line 128: the "response that comes back" is talking about the initial fetch only, is that right? Should that be clarified?

README.md
mfalken: kinu's suggestion works for me.

browsing-context.md
```diff
@@ -0,0 +1,28 @@
...
+    afterPrerendering();
+} else {
+    document.addEventListener('loadingmodechange', () => {
+        if (document.loadingMode.type === 'default')
+            afterPrerendering();
+    });
+}
+```
+
+Script can also observe this by using APIs particular to the behavior they are interested in. For instance, the [`document.hasStorageAccess()`][storage-access] can be used in supporting browsers to observe whether unpartitioned storage is available.
```
mfalken, line 28: The storage-access link seems broken in the rendered output.

mfalken (approved):
Really lucid explanations, thanks for writing this up.

README.md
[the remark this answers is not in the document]
domenic: That'd be up to the UA.

README.md
```diff
@@ -1,168 +1,177 @@
...
 
-A server could defer non-idempotent behavior such as impression counting,  or it could reject prerender requests altogether without needing to do the full work to generate a response (since it need only inspect the request headers).
+_Aside: it's probably safe to also allow same-origin prerendering with only a destination-side opt-in, as long as all of the same restrictions are applied (e.g., no credentials or storage access). But, this complicates the model a good deal, for both implementers and web developers. For now, we're concentrating on the model described above._
 
-> TODO: Should we also/instead send `Sec-Purpose`? Should `Sec-Fetch-Mode` differ?
+Finally, we'll note that browser-initiated prerenders fall somewhere in between these cases. In particular, the user typing `https://example.com/` in the URL bar, even before they press <kbd>Enter</kbd>, might serve as a reasonable prerender trigger, and perhaps even the prerendering could be done with credentials. The need to prevent user annoyance is still present, so the prerendering browsing context concept is important. But, what if the user types `https://example.com/logout`? Our thinking is still evolving in this area.
```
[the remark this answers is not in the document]
domenic: Changing to "browser UI-initiated prerender"; thanks!

README.md
domenic: Incorporated; thanks!

browsing-context.md
```diff
@@ -0,0 +1,28 @@
+# Prerendering browsing contexts
+
+TODO placeholder. Grab stuff from portals.
+
+* apply mitigations as above to subresource and scripted fetches
+* deny scripted access to unpartitioned storage, such as cookies and IndexedDB
+* deny permission to invoke `window.alert`, autoplay audio, and other APIs inappropriate at this time
+
+JS API probably belongs here (maybe it should use page visibility API instead):
```
jeremyroman, line 9: Yeah, possibly. I think the utility of a dedicated API basically depends on how many different kinds of modes there are that authors need to distinguish.

browsing-context.md
```diff
@@ -25,4 +25,4 @@ if (!document.loadingMode || document.loadingMode.type === 'default') {
 }
 ```
 
-Script can also observe this by using APIs particular to the behavior they are interested in. For instance, the [`document.hasStorageAccess()`][storage-access] can be used in supporting browsers to observe whether unpartitioned storage is available.
+Script can also observe this by using APIs particular to the behavior they are interested in. For instance, the [`document.hasStorageAccess()`][https://github.com/privacycg/storage-access] API can be used in supporting browsers to observe whether unpartitioned storage is available.
```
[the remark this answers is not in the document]
domenic: Oops, thanks, fixed.

domenic:
I don't have write access, so @jeremyroman, would you mind merging?

Issue #2, opened 2020-09-03
JavaScript API general discussion


domenic:
What is `document.loadingMode`? All we see is its `type` property, right?

Should we use a promise instead of the sync getter + change event pattern?

jeremyroman:
This is a bit of an attribute of the generality. I think that the simpler use cases are probably best served by something like privacycg/storage-access#55, which is simply a promise.

My thinking here was to expressly expose what the current loading mode is, which matters more if this expands. For example, if we end up needing "network-isolated" and "fenced" as separate modes, you'd want to be able to distinguish them. And some of these might make sense to parameterize (for example, we might wish to expose the notion of being proxied in some way), so exposing a dictionary or interface rather than an enum makes this more extensible.

If we're going to have an underlying primitive of "you are in this loading mode [therefore permissions policy is restricted in these ways, therefore you shouldn't necessarily record an impression, therefore ...]" in the spec, it seems natural to me to expose that to the platform too.

domenic:
Summarizing some discussion from https://github.com/jeremyroman/alternate-loading-modes/pull/5#discussion_r505846955 :

We have a few paths we could go with a JS API for prerendering and other loading modes.

- Super-minimal; no dedicated API. Maybe add a "prerender" mode to visibilityState, since per @bokan that seems web-compatible. (And it already has a change event.) But mostly people rely on purpose-specific APIs like permissions and storage access.

- Minimal but extensible. That's what the current proposal, slightly modified by #5, is trying to do. It has a LoadingMode interface, which so far only has EventTarget / onchange and the type property, but could grow if we want it to. The main delta vs. super-minimal here is that it allows distinguishing things like prerender vs. uncredentialed-prerender directly, instead of relying on other APIs like hasStorageAccess().

- Maximal, i.e. expose a bunch of useful data. (But, we don't know exactly what that is yet.)

PR #5, opened 2020-10-15
First draft of prerendering browsing context

domenic:
A lot of this is extracted and adapted from portals, but there are some differences.

I plan to do a second PR tomorrow which guts the portals README down to the portals-specific pieces, which will probably provide some more clarity and result in another round of edits. So feel free to hold off on reviewing this.

Some of the "rebasing" of portals on top of this is kind of interesting, e.g. if we envision prerendering browsing contexts as being offscreen, then it's much more obvious how the rendering APIs should behave. (As such, I deleted some of the text that portals has, since it's so obvious; in portals, I'll update to mention something like "like all prerendering browsing contexts, Intersection Observer doesn't work here. That's true even though the `<portal>` shows a preview.) Similarly, for portals we said we wouldn't respect `prefetch-src`, but I think if we're rebasing on top of this concept, then we should start respecting it.

Things which have been floating in my mind but I haven't written yet because I'm not sure they'll be important, or I'm not sure they need to be mentioned in the explainer as opposed to the spec:

- We're going to need some sort of "original URL" for these browsing contexts. E.g. if you prerender `https://a.example/` but that page redirects to `https://b.example/`, then the browser needs to know that the prerendering browsing context currently containing `https://b.example/` can be used to fulfill navigations to `https://a.example/`.

- Distinguishing between cross-origin and same-origin prerendering browsing contexts is nontrivial when redirects get involved. My initial instinct is that the "original URL" should be used for this, so if you're on `https://a.example/` and you prerender `https://b.example/`, that counts as a cross-origin prerendering, even if the response from `https://b.example/` redirects (either client or server-side) to `https://a.example/`. But maybe letting it change over time isn't so bad. I'll get into it more when we start speccing, I think...

- If we prerender `https://a.example/`, but the user Ctrl+clicks on `<a href="https://a.example/">Click me</a>` to open it in a new tab three times, then only the first of those new-tab loads will get to reuse the prerendering browsing context and be instant. Interesting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ function afterPrerendering() {
 if (!document.loadingMode || document.loadingMode.type === 'default') {
   afterPrerendering();
 } else {
-  document.addEventListener('loadingmodechange', () => {
+  document.loadingMode.addEventListener('change', () => {
     if (document.loadingMode.type === 'default') {
       afterPrerendering();
     }
--- a/browsing-context.md
+++ b/browsing-context.md
@@ -1,28 +1,251 @@
 # Prerendering browsing contexts
 
-TODO placeholder. Grab stuff from portals.
+We envision modernized prerendering to work by loading content into a **prerendering browsing context**. A prerendering browsing context can be thought of as a tab that is not yet shown to the user, and which the user has not yet affirmatively indicated an intention to visit. As such, it has additional restrictions placed on it to ensure the user's privacy and prevent disruptions.
 
-* apply mitigations as above to subresource and scripted fetches
-* deny scripted access to unpartitioned storage, such as cookies and IndexedDB
-* deny permission to invoke `window.alert`, autoplay audio, and other APIs inappropriate at this time
+Prerendering browsing contexts can be _activated_, which causes them to transition to being full top-level browsing contexts (i.e. tabs). From a user experience perspective, activation acts like an instantaneous navigation, since unlike normal navigation it does not require a network round-trip, creation of a `Document`, or running of the web-developer-provided initialization JavaScript. All of that has already been done in the prerendering browsing context. Activation might replace an existing top-level browsing context, for example if the user clicks a normal link whose target has been prerendered. Or it might just cause the a new top-level browsing context to exist, for example if the user clicks on a `target="_blank"` link. Activation lifts the restrictions on the prerendered content, as by that point a user-visible navigation has occurred.
 
-JS API probably belongs here (maybe it should use page visibility API instead):
+In general, activation of a prerendering browsing context is done by the user agent, when it notices a navigation that could use the prerendered contents. However, some forms of prerendering, such as [portals](https://github.com/WICG/portals/blob/master/README.md), can provide explicit entry points for navigation.
+
+`Document`s rendered within a prerendering browsing context have the ability to react to activation, which they can use to upgrade themselves once free of the restrictions. For example, they could start using permission-requiring APIs, or get access to unpartitioned storage.
+
+_Note: a browsing context is the right primitive here, as opposed to a `Window` or `Document`, as we need these restrictions to apply even across navigations. For example, if you prerender `https://a.example/` which contains `<meta http-equiv="refresh" content="0; URL=https://a.example/home">` then we need to continue applying these restrictions while loading the `/home` page._
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+## Table of contents
+
+- [Example](#example)
+- [Restrictions](#restrictions)
+  - [Privacy-based restrictions](#privacy-based-restrictions)
+    - [Storage access blocking](#storage-access-blocking)
+    - [Communications channels that are blocked](#communications-channels-that-are-blocked)
+    - [Communications channels that match navigation](#communications-channels-that-match-navigation)
+  - [Restrictions on the basis of being non-user-visible](#restrictions-on-the-basis-of-being-non-user-visible)
+  - [Restrictions on loaded content](#restrictions-on-loaded-content)
+- [JavaScript API](#javascript-api)
+  - [Current proposal](#current-proposal)
+  - [Adjacent APIs](#adjacent-apis)
+- [Session history](#session-history)
+- [Rendering-related behavior](#rendering-related-behavior)
+- [CSP integration](#csp-integration)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Example
+
+Consider `https://a.example/`, which contains the following HTML:
+
+```html
+<link rel="prerender2" href="https://b.example/">
+
+<a href="https://b.example/">Click me!</a>
+```
+
+_The `"prerender2"` rel here is illustrative only. See the [triggers document](./triggers.md) for more serious discussion of potential APIs for triggering prerendering._
+
+Upon loading `https://a.example/`, the browser notices the request to prerender `https://b.example/`. It does so by creating a prerendering browsing context, which it navigates to `https://b.example/`. This navigation takes place using [special fetch modes](./fetch.md), which ensure that `https://b.example/` has [opted in](./opt-in.md) to being prerendered, and ensures that the request for `https://b.example/` and any of its subresources is performed without any credentials that might identify the user.
+
+Within this prerendering browsing context, assuming the opt-in check passes, loading of `https://b.example/` proceeds mostly as normal. This includes any expensive web-developer-provided JavaScript necessary to initialize the web app found there. It could even include server- or client-side redirects to other pages, perhaps even other domains.
+
+However, if `https://b.example/` is one of those sites that requests notification permissions on first load, such a permission prompt will be denied, as if the user had declined. (TODO or should it hang, as if the user refused to respond?) Similarly, if `https://b.example/` performs an `alert()` call, the call will instantly return, without the user seeing anything. Another key difference is that `https://b.example/` will not have any storage access, including to cookies. Thus, the content it initially renders will be a logged-out view of the web app, or perhaps a specially-tailed "prerendering" view which leaves things like logged-in state indeterminate.
+
+Now, the user clicks on the "Click me!" link. At this point the user agent notices that it has a prerendering browsing context originally created for `https://b.example/`, so it activates it, replacing the one displaying `https://a.example/`. The user observes their browser navigating to `https://b.example/`, e.g., via changes in the URL bar contents and the back/forward UI. And since `https://b.example/` was already loaded in the prerendering browsing context, this navigation occurs seamlessly and instantly, providing a great user experience.
+
+Upon activation, `https://b.example/` gets notified via [the API](#javascript-api). At this point, it now has access to storage and cookies, so it can upgrade itself to a logged-in view if appropriate:
+
+```js
+document.loadingMode.addEventListener('change', () => {
+    if (document.loadingMode.type === 'default') {
+        document.getElementById('user').textContent = localStorage.getItem('current-user');
+    }
+});
+```
+
+This completes the journey to a fully-rendered view of `https://b.example/`, in a user-visible top-level browsing context.
+
+## Restrictions
+
+### Privacy-based restrictions
+
+Prerendering is intended to comply with the [W3C Target Privacy Threat Model](https://w3cping.github.io/privacy-threat-model/). This section discusses the aspects of that threat model that are particularly relevant to the browsing context part of the story, and how the design satisfies them.
+
+A prerendering browsing context can contain either a same-site or cross-site resource. Same-site prerendered content don't present any privacy risks, but cross-site resources risk enabling [cross-site recognition](https://w3cping.github.io/privacy-threat-model/#model-cross-site-recognition) by creating a messaging channel across otherwise-partitioned domains. For simplicity, when a cross-site channel needs to be blocked, we also block it for same-site cross-origin content. In some cases we event block it for same-origin content.
+
+Because prerendered browsing contexts can be activated, they (eventually) live in the first-party [storage shelf](https://storage.spec.whatwg.org/#storage-shelf) of their origin. This means that the usual plan of [storage partitioning](https://github.com/privacycg/storage-partitioning) does not suffice for prerendering browsing contexts as it does for nested browsing contexts (i.e. iframes). Instead, we take the following measures to restrict cross-origin prerendered content:
+
+- Prevent communication with the referring document, to the same extent we prevent it with a cross-site link opened in a new tab.
+- Block all storage access while content is prerendered.
+
+If we allowed communication, then the prerendered content could be given the user ID from the host site. Then, after activation gives the prerendered page access to first-party storage, it would join that user ID with information from its own first-party storage to perform cross-site tracking.
+
+If we allowed access to unpartitioned storage, then side channels available pre-activation (e.g., server-side timing correlation) could potentially be used to join two separate user identifiers, one from the referring site and one from the prerendered site's unpartitioned storage.
+
+The below subsections explore the implementation of these restrictions in more detail.
+
+#### Storage access blocking
+
+Prerendered pages that are cross-origin to their referring site will have no access to storage, similar to how an opaque-origin `<iframe>` behaves. (See this [discussion on the spec mechanism](https://github.com/whatwg/storage/issues/18#issuecomment-615336554).)
+
+We could attempt to address the threat by providing partitioned or ephemeral storage access, but then it is unclear how to transition to _unpartitioned_ storage upon activation. It would likely require some kind of web-developer-written merging logic. Completely blocking storage access is thus deemed simpler; prerendered pages should not be doing anything which requires persistent storage before activation.
+
+This means that most existing content will appear "broken" when prerendered by a cross-origin referrer. This necessitates an explicit opt-in to allow cross-origin content to be prerendered, [discussed elsewhere](./opt-in.md). Such content might optionally "upgrade" itself to a credentialed view upon activation, as shown in the [example](#example) above.
+
+For a more concrete example, consider `https://aggregator.example/` which wants to prerender this GitHub repository. To make this work, GitHub would need to add the opt-in to allow the page to be prerendered. Additionally, GitHub should add code to adapt their UI to show the logged-in view upon activation, by removing the "Join GitHub today" banner, and retrieving the user's credentials from storage and using them to replace the signed-out header with the signed-in header. Without such adapter code, activating the prerendering browsing context would show the user a logged-out view of GitHub in the top-level tab that the prerendering browsing context has been activated into. This would be a bad and confusing user experience, since the user is logged in to GitHub in all of their other top-level tabs.
+
+#### Communications channels that are blocked
+
+- Prerendering browsing contexts have no reference to the `Window`, or other objects, of their referrer. Thus, they cannot communicate using [`postMessage()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) or other APIs.
+- `BroadcastChannel` is disabled within prerendering browsing contexts.
+- Fetches within cross-origin prerendering browsing contexts, including the initial request for the page, do not use credentials. Credentialed fetches could be used for cross-site recognition, for example by:
+  - Using the sequence of loads. The referring page could encode a user ID into the order in which a sequence of URLs are prerendered. To prevent the target from correlating this ID with its own user ID without a navigation, a document loaded into a cross-origin prerendering browsing context is fetched without credentials and doesn't have access to storage, as described above.
+  - The host creates a prerendering browsing context, and the prerendered site decides between a 204 and a real response based on the user's ID. Or the prerendered site delays the response by an amount of time that depends on the user's ID. Because the prerendering load is done without credentials, the prerendered site can't get its user ID in order to make this sort of decision.
+- Sizing side channels: prerendering browsing contexts always perform layout based on the initial size of their referring browsing context, as its most likely that upon activation, they'll end up with that same size. However, further resizes to the referring browsing context are not used to update the size of the prerendering browsing context, as this could be used to communicate a user ID. For simplicity, this is the case for same-origin portals too.
+
+#### Communications channels that match navigation
+
+As mentioned above, we prevent communications to the same extent we prevent it with a cross-site link opened in a new tab. In particular:
+
+- The prerendered content's own URL and the referring URL are available to prerendered content to the same extent they're available to normal navigations. Solutions to link decoration will apply to both.
+
+Note that since a non-activated prerendering browsing context has no storage access, it cannot join any information stored in the URL with any of the prerendered site's data. So it's only activation, which gives full first-party storage access, which creates a navigation-equivalent communications channel. This equivalence makes sense, as activating a prerendering browsing context is much like clicking a link.
+
+### Restrictions on the basis of being non-user-visible
+
+Apart from the privacy-related restrictions to communications and storage, while prerendered, pages are additionally restricted in various ways due to the fact that the user has not yet expressed any intent to interact. All of these restrictions apply regardless of the same- or cross-origin status of the prerendered content.
+
+- Any features that are controlled by the [Permissions API](https://w3c.github.io/permissions/) ([list](https://w3c.github.io/permissions/#permission-registry)) will be automatically denied without prompting. TODO or deferred until activation?
+
+- Any features controlled by [Permissions Policy](https://w3c.github.io/webappsec-permissions-policy/) ([list](https://github.com/w3c/webappsec-permissions-policy/blob/master/features.md)) will be disabled, unless their default allowlist is `*`. There is no ability for the referring page to delegate these permissions. (In particular, there is no counterpart to `<iframe>`'s `allow=""` attribute.)
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -139,7 +139,7 @@ function afterPrerendering() {
 if (!document.loadingMode || document.loadingMode.type === 'default') {
   afterPrerendering();
 } else {
-  document.addEventListener('loadingmodechange', () => {
+  document.loadingMode.addEventListener('change', () => {
```
jeremyroman, line 142: This has the knock-on effect of requiring this to be an interface so that it can derive EventTarget, rather than being a dictionary, and also implies that its set of properties must remain fixed since the properties of a platform object don't change and we cannot swap it out for a different one because it has listeners.

That's why semantically I thought of it as changes to a property of the document, as opposed to an active thing in and of itself (analogous to orientationchange, languagechange, readystatechange, paymentmethodchange, etc).

browsing-context.md
```diff
@@ -1,28 +1,251 @@
...
 
-JS API probably belongs here (maybe it should use page visibility API instead):
+In general, activation of a prerendering browsing context is done by the user agent, when it notices a navigation that could use the prerendered contents. However, some forms of prerendering, such as [portals](https://github.com/WICG/portals/blob/master/README.md), can provide explicit entry points for navigation.
+
+`Document`s rendered within a prerendering browsing context have the ability to react to activation, which they can use to upgrade themselves once free of the restrictions. For example, they could start using permission-requiring APIs, or get access to unpartitioned storage.
```
jeremyroman, line 9: nit: the formatting of `Document`s  is weird, and since we're actually referring to the document rather than the WebIDL object that implements `Document`, I actually think Documents is preferred here, no?

browsing-context.md
```diff
@@ -1,28 +1,251 @@
...
+
+This completes the journey to a fully-rendered view of `https://b.example/`, in a user-visible top-level browsing context.
+
+## Restrictions
+
+### Privacy-based restrictions
+
+Prerendering is intended to comply with the [W3C Target Privacy Threat Model](https://w3cping.github.io/privacy-threat-model/). This section discusses the aspects of that threat model that are particularly relevant to the browsing context part of the story, and how the design satisfies them.
+
+A prerendering browsing context can contain either a same-site or cross-site resource. Same-site prerendered content don't present any privacy risks, but cross-site resources risk enabling [cross-site recognition](https://w3cping.github.io/privacy-threat-model/#model-cross-site-recognition) by creating a messaging channel across otherwise-partitioned domains. For simplicity, when a cross-site channel needs to be blocked, we also block it for same-site cross-origin content. In some cases we event block it for same-origin content.
```
jeremyroman, line 72: sp: s/event/even/

browsing-context.md
```diff
@@ -1,28 +1,251 @@
...
+  - The host creates a prerendering browsing context, and the prerendered site decides between a 204 and a real response based on the user's ID. Or the prerendered site delays the response by an amount of time that depends on the user's ID. Because the prerendering load is done without credentials, the prerendered site can't get its user ID in order to make this sort of decision.
+- Sizing side channels: prerendering browsing contexts always perform layout based on the initial size of their referring browsing context, as its most likely that upon activation, they'll end up with that same size. However, further resizes to the referring browsing context are not used to update the size of the prerendering browsing context, as this could be used to communicate a user ID. For simplicity, this is the case for same-origin portals too.
```
jeremyroman, line 102: Remove or revise last sentence to include prerendering if we want it to cover that as well. At the moment the mention of portals seems kinda out of left field in context.

browsing-context.md
```diff
@@ -1,28 +1,251 @@
...
+_Note: ideally Permissions Policy would become a superset of the Permissions API and iframe sandboxing. Then we could use that infrastructure as the single place to impose and lift these restrictions. Currently that is not the case, so the spec will be more messy._
+
+### Restrictions on loaded content
+
+To simplify implementation, specification, and the web-developer facing consequences, prerendering browsing contexts cannot host non-HTTP(S) `Document`s. In particular, they cannot host:
+
+- `javascript:` URLs
+- `data:` URLs
+- `blob:` URLs
+- `about:` URLs, including `about:blank` and `about:srcdoc`
```
jeremyroman, line 135: Hopefully nobody minds not being able to prerender `about:blank`. 😄

browsing-context.md
```diff
@@ -1,28 +1,251 @@
...
+
+In other cases, like `javascript:` URLs or `about:blank`, the problem is that those URLs generally inherit properties from their creator, and we don't want to allow this cross-`Document` influence for prerendered content. Overall, restricting to HTTP(S) URLs ensures that prerendered content always has a well-defined origin, that is not contingent on the referring page.
+
+The removal of the script-visible `about:blank` in prerendering browsing contexts also greatly simplifies them; its existence in other browsing contexts causes `Window`s and `Document`s to lose their normally one-to-one relationship.
+
+If a prerendering browsing context navigates itself to a non-HTTP(S) URL, e.g. via `window.location = "data:text/plain,foo"`, then the prerendering browsing context will be immediately discarded, and no longer be used by the user agent for anything.
```
jeremyroman, line 143: I guess it would still be fine for the user agent to cache the HTTP responses that lead there and use that on navigation. "for anything" seems a little strong.

browsing-context.md
```diff
@@ -1,28 +1,251 @@
...
+
+Goals:
+
+- Allow prerendered content to know that it's prerendered.
+- Allow prerendered content to know when it stops being prerendered.
+- Play nicely with existing or related APIs such as [the storage access API](https://github.com/privacycg/storage-access/) or the [page visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API).
+
+Nice-to-haves:
+
+- Be general enough to accomodate other "alternate loading modes", such as prefetching or [fenced frames](https://github.com/shivanigithub/fenced-frame/)
```
jeremyroman, line 157: Prefetching doesn't require a browsing context at all, though, so I'm not sure it has a mode.

browsing-context.md
```diff
@@ -1,28 +1,251 @@
...
+
+## Rendering-related behavior
+
+Prerendered content needs to strike a delicate balance, of doing enough rendering to be useful, but not actually displaying any pixels on the user's screen. As such, we want developers to avoid performing expensive work which is not beneficial while being prerendered. And ideally, doing this should require minimal additional coding by the developer of the page being prerendered.
+
+Generally speaking, our plan is to treat content as if it were in a "background tab": it will still perform layout, using (for [privacy and simplicity reasons](#communications-channels-that-are-blocked)) the creation-time size of the referring page as the viewport. Rendering APIs which communicate visibility information, such as [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) or the [`loading` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading), will indicate that no content is in the viewport. (TODO is this right, or should the initial screenful be counted as in the viewport? How does it compare to background tabs?)
```
jeremyroman, line 243: @bokand would be most up to date on this.

browsing-context.md
```diff
@@ -1,28 +1,251 @@
...
+
+## CSP integration
+
+A prerendered `Document` can apply CSP to itself as normal. Being in a prerendering browsing context vs. a normal top-level browsing context does not change any of the impacts of CSP. Note that since prerendered documents are [always loaded from HTTP(S) URLs](#restrictions-on-loaded-content), there is no need to worry about complex CSP inheritance semantics.
+
+Prerendered content will be affected by [`prefetch-src`](https://w3c.github.io/webappsec-csp/#directive-prefetch-src) on the referring page, which provides a way of preventing prefetching in addition to the [triggers](./triggers.md).
+
+The [`navigate-to`](https://w3c.github.io/webappsec-csp/#directive-navigate-to) directive prevents navigations, which means that if prerendered content is prevented from being navigated to via this mechanism, then the corresponding prerendering browsing context will never be activated. This mostly falls out automatically from the CSP spec preventing navigations, but any prerendering APIs that explicitly expose the activation operation (such as [portals](https://github.com/WICG/portals/blob/master/README.md)) will need to account for it in their specification.
```
jeremyroman, line 251: Interesting question whether the navigate-to directive should apply to the original or final URL. I think it's actually probably the former, since ideally it shouldn't detect any redirects that occurred (because it wouldn't in an ordinary nav).

jeremyroman (commented):
@kinu and @mfalken almost certainly have thoughts here too

README.md
domenic: It can't be a dictionary though, since it's returned from a Web IDL attribute. If it were a dictionary we'd need `document.loadingMode()` or similar.

Is there a situation in which we'd expect the properties to change? Stated another way, do you have a vision for this API that goes beyond what's outlined in the API section here?

browsing-context.md
domenic: I don't think there's a real distinction between "document" vs. "`Document` object"; the HTML spec generally doesn't make one. But perhaps for informal prose it's nicer.

browsing-context.md
domenic: Darn, I thought I got them all in my copy-pasting.

browsing-context.md
domenic: I still think this is accurate... the browsing context isn't used.

browsing-context.md
domenic: Good point. But the HTTP header/opt-in does apply for prefetching, and I guess what I'm getting at is that this is trying to stick with similar terminology... I'm unsure how to phrase that, hmm. Will think on it, and in the meantime delete the mention of prefetching.

browsing-context.md
domenic: I agree. That's kind of what I was discussing in the OP, about "original URL". Maybe I can write in something.

domenic: I guess that makes this less automatic; interesting.

README.md
jeremyroman: Hmm, I didn't realize IDL attributes can't return dictionaries. Basically I'm wondering if we'll want to parameterize these in the future and tell the content things like `{type: 'isolated-prerender', allowedOrigins: [...]}` or `{type: 'uncredentialed-prerender', viaProxy: '...'}`. I don't have concrete use cases for those at the moment, but if it's not too far out of the way it would be nice to allow for it in the future.

I suppose we do always have the option of later changing to a union of some kind but that's still potentially breaking.

I might be overengineering here. If so, maybe this should just be an enum and a `loadingmodechange` event. If not...hmm, what else could we do? If we do go for a platform object (like an `EventTarget`), it seems mildly weird to me to call it `loadingMode` (which sounds like inert data to me).

mfalken:
I haven't looked at the PR yet, but just responding to the conversation for now.

> * Distinguishing between cross-origin and same-origin prerendering browsing contexts is nontrivial when redirects get involved. My initial instinct is that the "original URL" should be used for this, so if you're on `https://a.example/` and you prerender `https://b.example/`, that counts as a cross-origin prerendering, even if the response from `https://b.example/` redirects (either client or server-side) to `https://a.example/`. But maybe letting it change over time isn't so bad. I'll get into it more when we start speccing, I think...

Considering A→B→A chains to be cross-origin is what the Fetch spec started to do in https://github.com/whatwg/fetch/issues/737. See also a blink-dev Intent to Ship for doing that for CSS at https://groups.google.com/a/chromium.org/d/msg/blink-dev/7OSy00oxVpk/siufiQVBBwAJ (described as "Change <span>#</span>2").

> * If we prerender `https://a.example/`, but the user Ctrl+clicks on `<a href="https://a.example/">Click me</a>` to open it in a new tab three times, then only the first of those new-tab loads will get to reuse the prerendering browsing context and be instant. Interesting.

For what it's worth, in Chrome I think we won't initially support Ctrl+click etc for prerendering as it requires some deeper implementation work than the new "Multi-Page Architecture" provides.  The prerendered contents will only be used for the tab that triggered it.

browsing-context.md
```diff
@@ -1,28 +1,290 @@
...
+  - The host creates a prerendering browsing context, and the prerendered site decides between a 204 and a real response based on the user's ID. Or the prerendered site delays the response by an amount of time that depends on the user's ID. Because the prerendering load is done without credentials, the prerendered site can't get its user ID in order to make this sort of decision.
+- Sizing side channels: prerendering browsing contexts always perform layout based on the initial size of their referring browsing context, as its most likely that upon activation, they'll end up with that same size. However, further resizes to the referring browsing context are not used to update the size of the prerendering browsing context, as this could be used to communicate a user ID. For simplicity, we apply this sizing model to same-origin prerendered content as well.
```
mfalken, line 102: I think these make sense but they seem a bit far-fetched compared to the referring site just using link decoration if it wants to pass user ID state over. Should we mention something about that?

browsing-context.md
```diff
@@ -1,28 +1,290 @@
...
+
+#### Communications channels that match navigation
```
mfalken, line 104: (I got as far as here, will read the rest when I can but sending what I have now)

mfalken (commented):
Thanks for cc'ing me. I added some comments.

README.md
domenic: I feel like we have three paths for this API:

- Super-minimal; no dedicated API. Maybe add a `"prerender"` mode to `visibilityState`, since per @bokan that seems web-compatible. (And it already has a change event.) But mostly people rely on purpose-specific APIs like permissions and storage access.

- Minimal but extensible. That's what my proposal here is trying to do: it has a `LoadingMode` interface, which so far only has `EventTarget` / `onchange` and the `type` property, but could grow if we want it to.

- Maximal, i.e. try to figure out the whole API including parameterization before shipping.

Maybe super-minimal is best?

browsing-context.md
```diff
@@ -1,28 +1,290 @@
...
+
+Because prerendered browsing contexts can be activated, they (eventually) live in the first-party [storage shelf](https://storage.spec.whatwg.org/#storage-shelf) of their origin. This means that the usual plan of [storage partitioning](https://github.com/privacycg/storage-partitioning) does not suffice for prerendering browsing contexts as it does for nested browsing contexts (i.e. iframes). Instead, we take the following measures to restrict cross-origin prerendered content:
+
+- Prevent communication with the referring document, to the same extent we prevent it with a cross-site link opened in a new tab.
+- Block all storage access while content is prerendered.
+
+If we allowed communication, then the prerendered content could be given the user ID from the host site. Then, after activation gives the prerendered page access to first-party storage, it would join that user ID with information from its own first-party storage to perform cross-site tracking.
```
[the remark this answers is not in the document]
domenic: Without activation, there's no user info to exfiltrate to the server, since you have no storage access.

browsing-context.md
```diff
@@ -1,28 +1,290 @@
...
+
+If we allowed access to unpartitioned storage, then side channels available pre-activation (e.g., server-side timing correlation) could potentially be used to join two separate user identifiers, one from the referring site and one from the prerendered site's unpartitioned storage.
```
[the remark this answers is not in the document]
domenic: > I got a little lost here as it's talking about unpartitioned storage while it a few paragraphs back said partitioning storage is not sufficient.

Maybe it would be clearer as "(unpartitioned) storage" instead? The idea is to explain why you get no storage access at all.

> But the two user IDs can only be joined if "referrer.test" and "prerendered.test" have access to the same storage, no?

Or if they have access to any communications channels, including side channels. So for example, if both `referrer.test` and `prerendered.test` know the time that `prerendered.test` loads, they can both communicate back to a collaborating server like so:

- At time t = 1603137935244, a user with `referrer.test` user ID "DomenicReferrer" was on `referrer.test`, and loaded `prerendered.test`.
- At time t = 1603137935244, a user with `prerendered.test` user ID "DomenicPrerendered" loaded `prerendered.test`.

The server then realizes that DomenicReferrer and DomenicPrerendered are the same person.

The mitigation is to cut off `prerendered.test`'s storage access pre-activation, so that it doesn't know the user ID "DomenicPrerendered" at prerendering time. It only gets to know that user ID after a user-visible full-screen activation.

README.md
kinu: I like starting with super-minimal, but getting back to this one once we started to feel something like this may be needed could probably work?

browsing-context.md
```diff
@@ -1,28 +1,290 @@
...
 
-* apply mitigations as above to subresource and scripted fetches
-* deny scripted access to unpartitioned storage, such as cookies and IndexedDB
-* deny permission to invoke `window.alert`, autoplay audio, and other APIs inappropriate at this time
+Prerendering browsing contexts can be _activated_, which causes them to transition to being full top-level browsing contexts (i.e. tabs). From a user experience perspective, activation acts like an instantaneous navigation, since unlike normal navigation it does not require a network round-trip, creation of a `Document`, or running of the web-developer-provided initialization JavaScript. All of that has already been done in the prerendering browsing context. Activation might replace an existing top-level browsing context, for example if the user clicks a normal link whose target has been prerendered. Or it might just cause the a new top-level browsing context to exist, for example if the user clicks on a `target="_blank"` link. Activation lifts the restrictions on the prerendered content, as by that point a user-visible navigation has occurred.
```
kinu, line 5: Some nit:
'All of that has already been done' --> Sites may choose not to run everything (or UA may freeze), maybe use a bit milder word (or say 'All or majority of ...')?

'cause the a new top-level' -> dup'ed 'the' and 'a'

kinu, line 5: Links opened with target="_blank" by default have opener relationship, then the prerendered page may not be able to be used?

jeremyroman: For what it's worth, this doesn't seem to be the case anymore:
https://groups.google.com/a/chromium.org/d/topic/blink-dev/FFX6VkvladY/discussion

kinu: Oh good to learn, I was behind! Thanks for sharing the intent link.

browsing-context.md
```diff
@@ -1,28 +1,290 @@
 # Prerendering browsing contexts
 
-TODO placeholder. Grab stuff from portals.
+We envision modernized prerendering to work by loading content into a **prerendering browsing context**. A prerendering browsing context can be thought of as a tab that is not yet shown to the user, and which the user has not yet affirmatively indicated an intention to visit. As such, it has additional restrictions placed on it to ensure the user's privacy and prevent disruptions.
```
kjmcnee, line 3: nit: Implied by the next paragraph, but maybe explicitly refer to a prerendering browsing context as a top level browsing context.

browsing-context.md
kjmcnee, line 5: nit: "the a" -> "the"

kjmcnee, line 5: nit: "cause the new top-level browsing context to exist" -> "cause the prerendered context to be shown in a new tab/window" ?

README.md
domenic: I'll make it clear this is a more on-going discussion. We can centralize that discussion in #2.

browsing-context.md
```diff
@@ -1,28 +1,295 @@
...
+
+If we allowed access to (unpartitioned) storage, then side channels available pre-activation (e.g., server-side timing correlation) could potentially be used to join two separate user identifiers, one from the referring site and one from the prerendered site's unpartitioned storage.
+
+The below subsections explore the implementation of these restrictions in more detail.
+
+#### Storage access blocking
+
+Prerendered pages that are cross-origin to their referring site will have no access to storage, similar to how an opaque-origin `<iframe>` behaves. (See this [discussion on the spec mechanism](https://github.com/whatwg/storage/issues/18#issuecomment-615336554).)
+
+We could attempt to address the threat by providing partitioned or ephemeral storage access, but then it is unclear how to transition to _unpartitioned_ storage upon activation. It would likely require some kind of web-developer-written merging logic. Completely blocking storage access is thus deemed simpler; prerendered pages should not be doing anything which requires persistent storage before activation.
```
kinu, line 92: Could we mention the possibility to block/defer granting access until activation at least for the ones that require no synchronous access as well?  At least while we're experimenting its viability I want the spec to keep mentioning it in some way.  If it can be a viable option opt-in could potentially express the behavior too?

browsing-context.md
```diff
@@ -1,28 +1,295 @@
...
+
+Note that since a non-activated prerendering browsing context has no storage access, it cannot join any information stored in the URL with any of the prerendered site's data. So it's only activation, which gives full first-party storage access, which creates a navigation-equivalent communications channel. This equivalence makes sense, as activating a prerendering browsing context is much like clicking a link.
+
+### Restrictions on the basis of being non-user-visible
+
+Apart from the privacy-related restrictions to communications and storage, while prerendered, pages are additionally restricted in various ways due to the fact that the user has not yet expressed any intent to interact. All of these restrictions apply regardless of the same- or cross-origin status of the prerendered content.
+
+- Any features that are controlled by the [Permissions API](https://w3c.github.io/permissions/) ([list](https://w3c.github.io/permissions/#permission-registry)) will be automatically denied without prompting. TODO or deferred until activation?
```
kinu, line 119: Reg: the TODO: yes, preferred that this also mentions deferring as a potential option.

browsing-context.md
```diff
@@ -1,28 +1,295 @@
...
+
+Goals:
+
+- Allow prerendered content to know that it's prerendered.
+- Allow prerendered content to know when it stops being prerendered.
+- Play nicely with existing or related APIs such as [the storage access API](https://github.com/privacycg/storage-access/) or the [page visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API).
+
+Nice-to-haves:
+
+- Be general enough to accomodate other "alternate loading modes", such as [fenced frames](https://github.com/shivanigithub/fenced-frame/)
```
kinu, line 160: Is life-cycle API or lifecycle state relevant here?

kinu: Or do we rather want to make this a new API like loadingMode?  The mode currently mostly focuses on prerendering state, while the naming may imply that this can express other state related to loading, is there a possible extension that we wan to think about in a future?  E.g. what about loaded from prefetch cache, or bfcache?

browsing-context.md
```diff
@@ -1,28 +1,295 @@
...
+
+Within this prerendering browsing context, assuming the opt-in check passes, loading of `https://b.example/` proceeds mostly as normal. This includes any expensive web-developer-provided JavaScript necessary to initialize the web app found there. It could even include server- or client-side redirects to other pages, perhaps even other domains.
+
+However, if `https://b.example/` is one of those sites that requests notification permissions on first load, such a permission prompt will be denied, as if the user had declined. (TODO or should it defer prompting the user until activation, so that the relevant promise doesn't settle until then?) Similarly, if `https://b.example/` performs an `alert()` call, the call will instantly return, without the user seeing anything. Another key difference is that `https://b.example/` will not have any storage access, including to cookies. Thus, the content it initially renders will be a logged-out view of the web app, or perhaps a specially-tailored "prerendering" view which leaves things like logged-in state indeterminate.
```
kinu, line 53: "will not have any (unpartitioned) storage access"

to be consistent with the statement below?

browsing-context.md
kinu, line 92: One more thoughts... for cookies I think it can be plausible that  a page can start with partitioned then can migrate to unpartitioned (where the cookie changes can be merged).  We have code that does this for NSP, because resource requests can also touch cookies.  It may be something we need more detailed convergence  /cc @mfalken 

(If we start to talk about all other storages can be more complicated though)

kinu:
It looks like there are a few points where we may not be fully converged, so I tired to comment on them. We'll want to discuss them sooner rather than later, but this PR doesn't need to be blocked on them.  I'd still like the text to keep some leeways for the exploration when possible.  Thanks!

browsing-context.md
domenic: Very interesting. We haven't discussed freezing in this explainer at all. Perhaps it's a good idea to add a separate section. I suspect we'd just use the existing page lifecycle API though, and not integrate with the loadingMode one.

Whether we want a new API (vs. using existing APIs like visibilityState), and what the scope of that API is, are still open questions. I've tried to link to #2 to discuss that.

browsing-context.md
domenic: I think it's more accurate in this section to say that you won't have any storage access, since that's the case. Below we're discussing the subtleties about unpartitioned vs. partitioned, but here we're discussing the whole landscape.

browsing-context.md
domenic: Added in https://github.com/jeremyroman/alternate-loading-modes/issues/7, and I'll add a note here linking there.

jeremyroman (approved):
lgtm generally

browsing-context.md
```diff
@@ -1,28 +1,311 @@
...
+
+- Any feature which requires [user activation](https://html.spec.whatwg.org/multipage/interaction.html#tracking-user-activation) will not be available, since user activation is not possible in prerendering browsing contexts.
+
+After activation, these restrictions are lifted: the content is treated like a normal top-level browsing context, and is able to use all these features in the normal way. The [API](#javascript-api) can be used to request permissions upon activation, if necessary. (Although doing so gives a user experience equivalent to requesting permissions on load, and thus is rarely the best design.)
+
+_Note: ideally Permissions Policy would become a superset of the Permissions API and iframe sandboxing. Then we could use that infrastructure as the single place to impose and lift these restrictions. Currently that is not the case, so the spec will be more messy._
+
+### Restrictions on loaded content
+
+To simplify implementation, specification, and the web-developer facing consequences, prerendering browsing contexts cannot host non-HTTP(S) `Document`s. In particular, they cannot host:
```
mfalken, line 139: This is talking about the top-level Document, right? The prerendering page can have about:blank iframes, I'd assume.

browsing-context.md
mfalken: Sounds good to have the note.

mfalken (approved):
lgtm, thanks for writing this.

kinu (approved):

browsing-context.md
domenic: Yep! I'll try to make it clearer.

PR #9, opened 2020-11-03
First draft of a spec

domenic:
This ports over portions of https://wicg.github.io/portals/, attempting to generalize along the way. The novel thing it introduces is storing all prerendering browsing contexts in a map, and patching the navigation algorithm to allow activation in place of some navigations. Most of what it's doing is setting up the scaffolding for further spec work on the restrictions and fetching changes, by giving the browsing context a "loading mode".

DO NOT MERGE until we figure out our gh-pages vs. master branch situation; this currently includes the built output, which makes me sad.

/cc @jakearchibald @kjmcnee as the new generalized "activation" has a clear spot for session history handling, and I'd be interested in whether it looks like it will line up with what we're hoping for. Right now it just has

> TODO prepend the existing session history of |predecessorBC| into |successorBC|? Or, probably better, use the new "browsing session" concept to bridge them? Be sure to respect |historyHandling|.

General note: we'll probably need a variety of special hooks where portals can do its thing, e.g. predecessor adoption and firing portalactivate and such during activation, or maybe some cleanup on failed navigations as noted inline. We can add those when I work on rebasing portals on top of this; for now I'm keeping it relatively self-contained and minimal.


domenic:
Spec preview: https://htmlpreview.github.io/?https://raw.githubusercontent.com/jeremyroman/alternate-loading-modes/38a5f466e7873b60f5da718357c0de86ebc71c8c/index.html

index.bs
```diff
@@ -1,17 +1,218 @@
...
+
+  1. [=Navigate=] |bc| to |startingURL| with the [=source browsing context=] set to |referrerDoc|'s [=Document/browsing context=].
+
+  1. Return |bc|.
+</div>
+
+<div algorithm>
+  To <dfn for="prerendering browsing context">activate</dfn> a [=prerendering browsing context=] |successorBC| in place of a [=top-level browsing context=] |predecessorBC| given a [=history handling behavior=] |historyHandling|:
+
+  1. Assert: |historyHandling| is either "`default`" or "`replace`".
```
kjmcnee, line 83: A historyHandling of "entry update" also makes sense for portals due to adoption, but presumably not for ordinary prerendering.

domenic: Hmm interesting, yeah, I think adopted browsing contexts are going to have a number of differences from other prerendering browsing contexts... they might need a separate loading mode or something like it. E.g. they get to keep storage access, IIRC.

index.bs
```diff
@@ -1,17 +1,218 @@
...
+    1. If |successorBC| is not [=prerendering browsing context/empty=], then:
+
+      1. [=prerendering browsing context/Activate=] |successorBC| in place of |browsingContext| given |historyHandling|.
+
+      1. Return.
+</div>
+
+<h3 id="always-replacement">Maintaining a trivial session history</h3>
+
+Patch the [=navigate=] algorithm to ensure the [=session history=] of a [=prerendering browsing context=] stays trivial as follows:
```
kjmcnee, line 141: I think we also need to patch
https://html.spec.whatwg.org/multipage/history.html#url-and-history-update-steps
as that does not appear to go through the navigation steps.

domenic: Great catch!!

index.bs
```diff
@@ -1,17 +1,222 @@
...
+Every [=browsing context=] has a <dfn for="browsing context">loading mode</dfn>, which is one of the following:
+
+: "`default`"
+:: No special considerations are applied to content loaded in this browsing context
+: "`prerender`"
+:: This browsing context is displaying prerendered content
+: "`uncredentialed-prerender`"
+:: This browsing context is displaying prerendered content, and furthermore that content cannot make credentialed fetches
+
+By default, a [=browsing context=]'s [=browsing context/loading mode=] is "`default`". A browsing context whose [=browsing context/loading mode=] is either "`prerender`" or "`uncredentialed-prerender`" is known as a <dfn>prerendering browsing context</dfn>.
```
jeremyroman, line 50: While I'm not super interested in prerendering navigations inside an iframe, is there a reason not to allow iframes to load in a prerendered page? If not, shouldn't those nested browsing contexts have a non-default loading mode?

index.bs
```diff
@@ -1,17 +1,222 @@
...
+<div algorithm="create a prerendering browsing context">
+  To <dfn export>create a prerendering browsing context</dfn> given a [=URL=] |startingURL| and a {{Document}} |referrerDoc|:
+
+  1. Assert: |startingURL|'s [=url/scheme=] is a [=HTTP(S) scheme=].
+
+  1. If |referrerDoc|'s [=Document/prerendering browsing contexts map=][|startingURL|] [=map/exists=], then return it.
+
+  1. Let |bc| be the result of [=creating a new top-level browsing context=].
+
+  1. Set |bc|'s [=browsing context/loading mode=] to "`uncredentialed-prerender`".
```
jeremyroman, line 69: nit: Any particular reason to express it as being set and then possibly changed on the next line, as opposed to saying it's "`prerender`" if they are same-origin, and "`uncredentialed-prerender`" otherwise?

index.bs
```diff
@@ -1,17 +1,222 @@
...
+
+  1. [=Prompt to unload=] the [=active document=] of |predecessorBC|. If the user [=refused to allow the document to be unloaded=], then return.
+
+  1. [=Document/Abort=] the [=active document=] of |predecessorBC|.
+
+  <!-- End copied section. -->
+
+  1. TODO prepend the existing session history of |predecessorBC| into |successorBC|? Or, probably better, use the new "browsing session" concept to bridge them? Be sure to respect |historyHandling|.
+
+  1. [=In parallel=]:
```
jeremyroman, line 101: I assume this section would get subsumed if we refactored navigation to allow BC swapping that preserves session history?

index.bs
```diff
@@ -1,17 +1,222 @@
...
+<div algorithm="navigate activate patch">
+  In [=navigate=], append the following steps after the fragment navigation handling (currently step 6):
+
+  1. If all of the following are true:
+
+      * |browsingContext| is a [=top-level browsing context=]
+      * |historyHandling| is "`default`" or "`replace`"
+      * <var ignore>navigationType</var> is "`other`"
+      * |resource| is a [=request=] whose [=request/method=] is \``GET`\`
+      * |browsingContext|'s [=active document=]'s [=Document/prerendering browsing contexts map=][|resource|'s [=request/URL=]] [=map/exists=]
```
jeremyroman, line 126: Are there any other request fields that need to form part of this key? e.g. referrer policy (I think not)?

index.bs
```diff
@@ -1,17 +1,222 @@
...
+</div>
+
+<h3 id="always-replacement">Maintaining a trivial session history</h3>
+
+<div algorithm="navigate historyHandling patch">
+  Patch the [=navigate=] algorithm to ensure the [=session history=] of a [=prerendering browsing context=] stays trivial by prepending the following step before all others:
+
+  1. If <var ignore>browsingContext</var> is a [=prerendering browsing context=], then:
+
+    1. Assert: |historyHandling| is not "`entry update`", since prerendering browsing contexts have trivial session histories and thus will never end up [=traverse the history|traversing=] back to an entry with null [=session history entry/document=].
```
jeremyroman, line 146: Nit: mixing a normative assertion with a non-normative explanation seems a little odd; is this usual?

jeremyroman (approved):

index.bs
domenic: Good point. (Although I think the plan was to disallow loading of pages without an opt-in.)

I think I'll keep these three values for now as it's clear how they're needed, and add a red box saying that it needs to expand to handle iframes once we do some more work on the fetch parts of things.

index.bs
domenic: I think the plan is to refactor navigation to introduce the "browsing session" concept, where session history moves to. So browsing contexts won't really change. So I think that impacts the line above, but not these steps.

index.bs
domenic: I at first thought not, but now that you bring up referrer policy, I think that does need to be a key. In particular, consider

```
<link rel="prerender2" referrerpolicy="unsafe-url" href="https://example.com">

<a referrerpolicy="no-referrer" href="https://example.com/">
```

I don't think it's a good idea to use the prerendered version from the `<link>` to fulfill any navigations from that `<a>`...

index.bs
domenic: Yeah, I try to explain the reasoning behind all non-obvious assertions in this sort of way.

domenic:
To deal with the branch situation, I won't end up merging this PR directly, but instead creating a new "main" branch and then overwriting gh-pages with the built output.

domenic:
OK, I merged it to main, with GitHub actions auto-deploying it to gh-pages:

- https://jeremyroman.github.io/alternate-loading-modes/
- https://github.com/jeremyroman/alternate-loading-modes/tree/main
- https://github.com/jeremyroman/alternate-loading-modes/tree/gh-pages

What remains is for Jeremy to update the default branch from gh-pages to main.

PR #11, opened 2020-11-24
Initial draft of prerendering-state.md

bokand:
Adds a document describing a proposal for `document.prerendering` and
updates some of the text in browsing-context.md to match current state
of thinking.

This is something along the lines of a "minimal" proposal in #2.

### DIFF
--- a/browsing-context.md
+++ b/browsing-context.md
@@ -203,7 +203,7 @@ In the future, `document.loadingMode` might have additional properties; for exam
 For exposing the semantic notion of prerendering (i.e. the user didn't initiate the load, the page is not interactive), we propose `document.prerendering`:
 
 ```js
-if (document.prerendering) {
+if (!document.prerendering) {
     afterPrerendering();
 } else {
     document.onprerenderingchange = afterPrerendering;
--- a/browsing-context.md
+++ b/browsing-context.md
@@ -167,7 +167,7 @@ Goals:
 
 Nice-to-haves:
 
-- Be general enough to accomodate other "alternate loading modes", such as [fenced frames](https://github.com/shivanigithub/fenced-frame/)
+- Be general enough to accommodate other "alternate loading modes", such as [fenced frames](https://github.com/shivanigithub/fenced-frame/)
 
 ### Current proposal
 
@@ -177,6 +177,8 @@ The current proposed API is a `document.loadingMode` object with:
 - An event, `"change"`, which fires when `type` changes.
 
 Envisioned usage is as follows:
+
+TODO: This example would likely want to use `document.prerendering` - do we have a example that demonstrates the need for additional granularity?
 
 ```js
 function afterPrerendering() {
@@ -197,6 +199,18 @@ if (!document.loadingMode || document.loadingMode.type === 'default') {
 ```
 
 In the future, `document.loadingMode` might have additional properties; for example, it might expose the notion that the page was loaded via some proxy, as mentioned in the [fetch integration](./fetch.md).
+
+For exposing the semantic notion of prerendering (i.e. the user didn't initiate the load, the page is not interactive), we propose `document.prerendering`:
+
+```js
+if (document.prerendering) {
+    afterPrerendering();
+} else {
+    document.onprerenderingchange = afterPrerendering;
+}
+```
+
+See [prerendering-state](prerendering-state.md) for more details.
 
 ### Adjacent APIs
 
@@ -224,7 +238,7 @@ if (geoPermission.state === "denied") {
 
 _Note: the above code only makes sense if we decide that permissions are denied in prerendering browsing contexts, instead of having them hang until activation. [That plan](#restrictions-on-the-basis-of-being-non-user-visible) is still tentative._
 
-Finally, there's the case of the [page visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API). Our current plan is to treat prerendering browsing contexts as hidden, until activation, in which case code that only wants to run upon the user viewing the page could be done as follows:
+Finally, there's the case of the [page visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API). Our current plan is to treat (non-portal) prerendering browsing contexts as hidden, until activation, in which case code that only wants to run upon the user viewing the page could be done as follows:
 
 ```js
 if (document.visibilityState === "visible") {
@@ -240,8 +254,6 @@ document.addEventListener("visibilitychange", () => {
     }
 });
 ```
-
-_An alternative is to reintroduce the `"prerender"` visibility state, which was briefly specified, but never implemented. However, we worry that this would break code like the above, which assumes that there are only two visibility states in existence._
 
 ## Page lifecycle and freezing
 
@@ -286,7 +298,7 @@ Another interesting situation to consider is what happens if the user right-clic
 
 Prerendered content needs to strike a delicate balance, of doing enough rendering to be useful, but not actually displaying any pixels on the user's screen. As such, we want developers to avoid performing expensive work which is not beneficial while being prerendered. And ideally, doing this should require minimal additional coding by the developer of the page being prerendered.
 
-Generally speaking, our plan is to treat content as if it were in a "background tab": it will still perform layout, using (for [privacy and simplicity reasons](#communications-channels-that-are-blocked)) the creation-time size of the referring page as the viewport. Rendering APIs which communicate visibility information, such as [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) or the [`loading` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading), will indicate that no content is in the viewport. (TODO is this right, or should the initial screenful be counted as in the viewport? How does it compare to background tabs?)
+Generally speaking, our plan is to treat content as if it were in a "background tab": it will still perform layout, using (for [privacy and simplicity reasons](#communications-channels-that-are-blocked)) the creation-time size of the referring page as the viewport. Rendering APIs which communicate visibility information, such as [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) or the [`loading` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading), will indicate visibility based on the creation-time viewport.
 
 ## CSP integration
 
--- a/browsing-context.md
+++ b/browsing-context.md
@@ -178,7 +178,7 @@ The current proposed API is a `document.loadingMode` object with:
 
 Envisioned usage is as follows:
 
-TODO: This example would likely want to use `document.prerendering` - do we have a example that demonstrates the need for additional granularity?
+TODO: This example would likely want to use `document.prerendering` - even with `document.prerendering`, additional loading-mode information may still be useful for specialized cases where fetching may need to depend on the exact loading mode, can we find any examples/use cases? `document.prerendering` will likely cover the most common use cases without getting into the level of granularity of fetching restrictions.
 
 ```js
 function afterPrerendering() {
@@ -209,6 +209,8 @@ if (document.prerendering) {
     document.onprerenderingchange = afterPrerendering;
 }
 ```
+
+Note: it's possible that `document.prerendering` may be  just syntactic sugar for some subset of `document.loadingMode.type`, e.g. `document.loadingMode.type != 'default'` or `document.loadingMode.type == 'prerender' || document.loadingMode.type == 'uncredentialed-prerender'`.
 
 See [prerendering-state](prerendering-state.md) for more details.
 
--- a/prerendering-state.md
+++ b/prerendering-state.md
@@ -92,7 +92,18 @@ a user may not have initiated the page load in the first place.
 | Portal         | true           | 'visible'         |
 | Prerender      | true           | 'hidden'          |
 
-<<<<<<< HEAD
+### Relationship to BFCache
+
+Some user agents provide a "Back-Forward" cache which keeps recent documents from the session history alive, enabling
+instant history based navigations. While in the cache, these documents are typically put into a frozen state where
+script cannot run and APIs and features with side-effects are blocked (e.g. camera, BroadcastMessage, etc.).
+
+On the surface, this frozen state shares some resemblance with prerendering restrictions, which also aim to limit
+visible side-effects and may, in some circumstances, completely freeze a page. It's likely that some of these mechanisms
+will be shared between prerendering browsing contexts and BFCache implementations; however, we don't anticipate
+BFCache'd documents to be put into a prerendering browsing context. As such, `document.prerendering` would not reflect a
+BFCache'd state. The [Page Lifecycle APIs](https://wicg.github.io/page-lifecycle) remain relevant for this case.
+
 ## Open Questions
 
 ### `visibilityState` as a termination signal
--- a/prerendering-state.md
+++ b/prerendering-state.md
@@ -6,7 +6,7 @@ We propose introducing a boolean property `document.prerendering` and associated
 prerendering browsing contexts from regular ones.
 
 _Note: This is based on the more general discussion in this [background
-doc](https://docs.google.com/document/d/1Xzw0k8DgltI2ohapuDKmjRZLv7NVrRFGusW8IBtiCT0/edit?usp=sharing)_
+doc](https://docs.google.com/document/d/1Xzw0k8DgltI2ohapuDKmjRZLv7NVrRFGusW8IBtiCT0/edit?usp=sharing)._
 
 ## Background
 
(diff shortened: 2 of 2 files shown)


prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+    // Defer loading large images
+}
+```
+
+An analysis [found several
+examples](https://docs.google.com/document/d/1Xzw0k8DgltI2ohapuDKmjRZLv7NVrRFGusW8IBtiCT0/edit#heading=h.rkorueof7xev)
+of such behavior.
+
+(re-)Introducing a `prerender` state could break such pages. In particular, portals (or some other, future
+prerendering-while-visible context) would show static or missing content. A hidden prerender may not have immediate
```
ianvollick, line 63: The term "preview" might be a more intuitive way of describing prerendering-while-visible.

bokand: I agree, but I think it makes sense for this be a signal that we're in a prerendering-browsing-context. If that's the case, I think it'd make sense to match the names of the two things;  "prerendering-browsing-context" has the same issue. Would it make sense to rename them both? @jeremyroman @domenic for their thoughts.

ianvollick: Ah, I was only suggesting we might use the term preview in the prose describing what the new prerender state did, not changing the name of anything web-facing.

bokand: Ah - true, but that does beg the question whether naming the web-facing APIs 'prerender' evokes the right meaning if we need to clarify in prose

browsing-context.md
```diff
@@ -178,6 +178,8 @@ The current proposed API is a `document.loadingMode` object with:
 
 Envisioned usage is as follows:
 
+TODO: This example would likely want to use `document.prerendering` - do we have a example that demonstrates the need for additional granularity?
```
domenic, line 181: I can work on this TODO after this is merged.

In particular, I'm now feeling that we should reframe this section to say something like:

- Use all these individual APIs!
- But, maybe it'd be a good idea to expose the underlying primitive that the user agent is using, i.e. the loading mode? If so, here's our proposal. We're not yet aware of any use cases though, so consider this on-hold.

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
+# Prerendering State API
+
+## Summary
+
+We propose introducing a boolean property `document.prerendering` and associated change event to distinguish
+prerendering browsing contexts from regular ones.
+
+_Note: This is based on the more general discussion in this [background
+doc](https://docs.google.com/document/d/1Xzw0k8DgltI2ohapuDKmjRZLv7NVrRFGusW8IBtiCT0/edit?usp=sharing)_
```
domenic, line 9:
```suggestion
doc](https://docs.google.com/document/d/1Xzw0k8DgltI2ohapuDKmjRZLv7NVrRFGusW8IBtiCT0/edit?usp=sharing)._
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Background
+
+Pages need to know when they're being rendered inside a [prerendering browsing context](browsing-context.md).
+Some examples of divergent behavior:
+
+* Avoid fetching large video resources
+* Avoid measuring ad impressions
+* Avoid running javascript animations
```
domenic, line 18:
```suggestion
* Avoid running JavaScript animations
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+* Avoid measuring user interactions for analytics
+* Avoid starting server-coordinated behavior (e.g. ticket purchase countdown).
+
+A previous iteration of prerendering used [page-visibility](https://www.w3.org/TR/page-visibility-2) to denote a
```
domenic, line 22:
```suggestion
A previous iteration of prerendering used [page-visibility](https://w3c.github.io/page-visibility/) to denote a
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+prerendering context. However, this was unimplemented and [removed from the
+specification](https://github.com/w3c/page-visibility/issues/42).
+
+With the re-introduction of prererendering, we need to bring back some form of this API.
+
+## Why not visibilityState=='prerender'?
+
+The most straightforward way to do this would be to bring back the previously specified `visibilityState ==
```
domenic, line 30:
```suggestion
The most straightforward way to do this would be to bring back the previously specified `document.visibilityState ==
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+'prerender'`.  Doing so has a number of drawbacks:
+
+### Not mutually-exclusive
+
+In actuality, some of the above use cases are interested in visibility rather than prerendering state.  For example, an
+ordinary (i.e. non-prerendering) page that's hidden would likely want to "avoid running javascript animations" and
```
domenic, line 36:
```suggestion
ordinary (i.e. non-prerendering) page that's hidden would likely want to "avoid running JavaScript animations" and
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+"avoid fetching large video resources". Thus, pages would have to consider both the `hidden` state as well as a
+re-introduced `prerender` to get the desired behavior.
+
+However, the [portals proposal](https://github.com/WICG/portals/) means that `prerender` isn't mutually exclusive
+with `visible` either. In that case, we do want to "run javascript animations" and "fetch large resources" since it'll
```
domenic, line 41:
```suggestion
with `visible` either. In that case, we do want to "run JavaScript animations" and "fetch large resources" since it'll
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+be visible to the user, despite being in a prerendering browsing context (and all the restrictions that come with that).
+
+### Web-compatibility
+
+Related to the above point, a common pattern on the web is to assume only two visibility states. For example, code such
+as:
+
+```
```
domenic, line 49:
```suggestion
```js
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+
+(re-)Introducing a `prerender` state could break such pages. In particular, portals (or some other, future
+prerendering-while-visible context) would show static or missing content. A hidden prerender may not have immediate
+user-visible effects, but may use more resources than necessary.
+
+## document.prerendering
+
+We propose adding a new state variable:
+
+```
```
domenic, line 70:
```suggestion
```webidl
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+
+`document.prerendering` returns true if the document is currently hosted in a [prerendering browsing
+context](browsing-context.md). The `prerenderingchange` event is fired whenever this value changes.
+
+A state separate from visibility ensures existing visibility-based states are correctly accounted for. It also forces
+authors to consider their page's behavior in light of prerendering restrictions and non-interactivity and the fact that
+a user may not have initiated the page load in the first place.
+
+### Relationship to visibilityState
```
domenic, line 84:
```suggestion
### Relationship to `visibilityState`
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+`prerendering` and `visibilityState` are entirely independent and all combinations of states are possible:
+
+| State          | `prerendering` | `visibilityState` |
+| -------------- | -------------- | ----------------- |
+| Foreground Tab | false          | 'visible'         |
+| Background Tab | false          | 'hidden'          |
+| Portal         | true           | 'visible'         |
+| Prerender      | true           | 'hidden'          |
+
+#### VisibilityState as a termination signal
```
domenic, line 95:
```suggestion
#### `visibilityState` as a termination signal
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+
+An analysis [found several
+examples](https://docs.google.com/document/d/1Xzw0k8DgltI2ohapuDKmjRZLv7NVrRFGusW8IBtiCT0/edit#heading=h.rkorueof7xev)
+of such behavior.
+
+(re-)Introducing a `prerender` state could break such pages. In particular, portals (or some other, future
+prerendering-while-visible context) would show static or missing content. A hidden prerender may not have immediate
+user-visible effects, but may use more resources than necessary.
+
+## document.prerendering
```
domenic, line 66:
```suggestion
## `document.prerendering`
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+* Avoid measuring user interactions for analytics
+* Avoid starting server-coordinated behavior (e.g. ticket purchase countdown).
+
+A previous iteration of prerendering used [page-visibility](https://www.w3.org/TR/page-visibility-2) to denote a
+prerendering context. However, this was unimplemented and [removed from the
+specification](https://github.com/w3c/page-visibility/issues/42).
+
+With the re-introduction of prererendering, we need to bring back some form of this API.
+
+## Why not visibilityState=='prerender'?
```
domenic, line 28:
```suggestion
## Why not `visibilityState=='prerender'`?
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+| Foreground Tab | false          | 'visible'         |
+| Background Tab | false          | 'hidden'          |
+| Portal         | true           | 'visible'         |
+| Prerender      | true           | 'hidden'          |
+
+#### VisibilityState as a termination signal
+
+Authors have been [encouraged](https://www.igvita.com/2015/11/20/dont-lose-user-and-app-state-use-page-visibility/) to
+switch from `unload` and `beforeunload` events to `visibilityState=='hidden'` as a signal that their page may be
+terminated. See also [PageVisibility#59](https://github.com/w3c/page-visibility/issues/59) for related discussion.
```
domenic, line 99:
```suggestion
terminated. See also [w3c/PageVisibility#59](https://github.com/w3c/page-visibility/issues/59) for related discussion.
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+
+Entering a `prerendering` state, via being put into a portal, is likely a point where such a termination signal should
```
domenic, line 101: maybe "being put into a portal (via portal predecessor adoption)"?

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+be fired; it is analogous to navigating to a new page. However, this means pages would now also have to listen for a
+switch to `prerendering == true`.
+
+This [background
+doc](https://docs.google.com/document/d/1Xzw0k8DgltI2ohapuDKmjRZLv7NVrRFGusW8IBtiCT0/edit#heading=h.acmnp6zdmcik) deals
+extensively with this issue. An ideal solution would introduce a new event for this use case, rather than implicitly
+tying it to any particular state.
```
domenic, line 108: Maybe note that you think that can be done as a followup, and in the meantime pages won't handle the case of being put into a portal perfectly, and that's OK?

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Relationship to loading-modes
```
domenic, line 110: Let's delete this section and roll it into the TODO you inserted above, which I commented on.

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+examples?).
+
+However, we believe `document.prerendering` will cover the most common use cases without getting into the level of
+granularity of fetching restrictions.
+
+Note: it's possible that `document.prerendering` may be  just syntactic sugar for some subset of
+`document.loadingMode.type`, e.g. `document.loadingMode.type != 'default'` or `document.loadingMode.type ==
+'prerender' || document.loadingMode.type == 'uncredentialed-prerender'`.
+
+### Relationship to other states
```
domenic, line 124: I wonder if you want an "open questions" h2-level section, which contains "visibilityState as a termination signal", "relationship to other states", and "Naming bikeshed"?

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+would think about it. Additionally, a portal may enter the prerendering state from a regular context; this feels
+unintuitive to call "prerendering".
+
+We should also consider future use cases. For example, if
+[fenced-frames](https://github.com/shivanigithub/fenced-frame/) were to use this special browsing context, does it make
+sense to call it "prerendering"?
+
+The term "interactive" seems appealing as it captures a major difference between prerendering/portals and regular
+browsing contexts. However, it would be unintuitive to explain how a background tab is considered "interactive". In
+particular, if we want this to solve [PageVisibility#59](https://github.com/w3c/page-visibility/issues/59) we'd have the
```
domenic, line 148:
```suggestion
particular, if we want this to solve [w3c/PageVisibility#59](https://github.com/w3c/page-visibility/issues/59) we'd have the
```

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+As noted in the [background
+doc](https://docs.google.com/document/d/1Xzw0k8DgltI2ohapuDKmjRZLv7NVrRFGusW8IBtiCT0/edit#heading=h.14z99pd6akf0), there
+are other states where it may make sense to report `prerendering` state. For example, a page loading into a mobile
+background tab maybe terminated, without notice, before the user ever sees it. Similarly, the live-preview in a mobile app
+or tab switcher is a non-interactive preview, reminiscent of a portal.
+
+Should these states also report `prerendering==true`? Should they use the prerendering browsing context concept? The
+answers here aren't clear but this is worth considering.
+
+### Naming bikeshed
```
domenic, line 135: Good summary of the issues!

domenic (approved):
Generally looks good! I added nits as inline suggestions; only a few substantive comments.

prerendering-state.md
```diff
@@ -0,0 +1,150 @@
...
+
+We should also consider future use cases. For example, if
+[fenced-frames](https://github.com/shivanigithub/fenced-frame/) were to use this special browsing context, does it make
+sense to call it "prerendering"?
+
+The term "interactive" seems appealing as it captures a major difference between prerendering/portals and regular
+browsing contexts. However, it would be unintuitive to explain how a background tab is considered "interactive". In
+particular, if we want this to solve [PageVisibility#59](https://github.com/w3c/page-visibility/issues/59) we'd have the
+bizarre case that entering the app switcher would make a page non-interactive, then backgrounding it from there would make
+it interactive.
```
jeremyroman, line 150: It might be worth mentioning how it does (or doesn't, IIUC) interact with bfcache.

browsing-context.md
jeremyroman: Totally open to hiding loading mode from the JS API if we think the use cases that matter are covered by narrower APIs, and adding it later on if we actually need it (and are better informed about what it should look like).

prerendering-state.md
bokand: Done

prerendering-state.md
bokand: Done

prerendering-state.md
bokand: Good idea, done.

prerendering-state.md
bokand: Done

browsing-context.md
bokand: I've been thinking of `document.prerendering` as being based on whether or not the document is being rendered in a prerendering browsing context. Assuming that's true, it's also an underlying primitive.

The loading-mode seems related but maybe orthogonal? Thought experiment: would it ever make sense to have an existing loading mode (other than `default`), or perhaps a new one in the future, that was used in a regular browsing context? Or using the 'default' loading mode in a prerendering browsing context (would we use that for same-origin prerender/portal)? This doesn't sound crazy to me so I think maybe `document.prerendering` isn't necessarily just sugar for `loadingMode != 'default'`

prerendering-state.md
bokand: I've been thinking of `document.prerendering` as "is in prerendering browsing context". Does BFCache have any expected interaction with the concept of a prerendering browsing context introduced in this repo? I expect not (and in that case I'll add that here) but I'm not sure if we're thinking of reusing some of the rendering/permissions restrictions.

prerendering-state.md
bokand: Thanks :)

prerendering-state.md
jeremyroman: I think likely not. Though it's similar in some regards to prerendering, script can't observe that much because it gets stopped for the duration where that would apply. But I think it's non-obvious enough to note.

browsing-context.md
```diff
@@ -198,6 +200,20 @@ if (!document.loadingMode || document.loadingMode.type === 'default') {
 
 In the future, `document.loadingMode` might have additional properties; for example, it might expose the notion that the page was loaded via some proxy, as mentioned in the [fetch integration](./fetch.md).
 
+For exposing the semantic notion of prerendering (i.e. the user didn't initiate the load, the page is not interactive), we propose `document.prerendering`:
+
+```js
+if (document.prerendering) {
```
mfalken, line 206: if (!document.prerendering) ?

prerendering-state.md
```diff
@@ -0,0 +1,140 @@
...
+`prerendering` and `visibilityState` are entirely independent and all combinations of states are possible:
+
+| State          | `prerendering` | `visibilityState` |
+| -------------- | -------------- | ----------------- |
+| Foreground Tab | false          | 'visible'         |
+| Background Tab | false          | 'hidden'          |
+| Portal         | true           | 'visible'         |
+| Prerender      | true           | 'hidden'          |
+
+<<<<<<< HEAD
```
mfalken, line 95: remove?

prerendering-state.md
```diff
@@ -0,0 +1,140 @@
...
+"avoid fetching large video resources". Thus, pages would have to consider both the `hidden` state as well as a
+re-introduced `prerender` to get the desired behavior.
+
+However, the [portals proposal](https://github.com/WICG/portals/) means that `prerender` isn't mutually exclusive
+with `visible` either. In that case, we do want to "run JavaScript animations" and "fetch large resources" since it'll
+be visible to the user, despite being in a prerendering browsing context (and all the restrictions that come with that).
+
+### Web-compatibility
+
+Related to the above point, a common pattern on the web is to assume only two visibility states. For example, code such
```
mfalken, line 46: Hm, this is rather unfortunate since `prerender` was only removed from the spec in 2019 and WebKit just removed it from trunk a few weeks ago in https://bugs.webkit.org/show_bug.cgi?id=215851. It's unfortunate that visibilityState is not a boolean, but we don't want to add new types because sites are treating it as one. But the argument above about prerendering not being mutually exclusive with visibility and hidden makes sense.

prerendering-state.md
mfalken: +1 could be good to mention this. My understanding also is that bfcached pages do not run script so can't observe document.prerendering.

mfalken (approved):
lgtm for what it's worth, with some comments.

jeremyroman (approved):
lgtm whenever others are happy

prerendering-state.md
bokand: This is still outstanding but unresolved. I filed #12 to keep this in mind but I think we can punt on this for now and leave it as "prerendering" since it's much clearer to outsiders what we're talking about. As other use cases shape up we can revisit it in the future.

bokand:
Thanks everyone!

I don't have write access so could someone hit "merge" for me?

PR #16, opened 2021-01-11
Update our thinking on pre-activation restrictions

domenic:
This follows much of the discussion in https://docs.google.com/document/d/1zY15k_wFTik2EoxBf3_RT7YjYpFMDaeNspy15n0rtww/edit?usp=sharing, porting that to the explainer.

The specification is updated for the single so-far-specified restriction, of downloads, to show how we might specify the others.

Potential reviewers: @jeremyroman, @jakearchibald, @domfarolino

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ We envision prerendering having several related pieces:
 
 * [**Prerendering browsing contexts**](./browsing-context.md), which are special browsing contexts that are not displayed to the user, and within which content is constrained to not perform disruptive or side-effecting operations.
 
-  In all prerendering browsing contexts, side-effecting or disruptive APIs, such as those that could play media, require a permission prompt, or otherwise display UI, will be disabled in API-specific ways. In those tagged as being used for cross-origin prerendering, storage access will not be available, and all fetches will need to use the prerendering fetching modes.
+  In all prerendering browsing contexts, disruptive APIs, such as those that could play media, require a permission prompt, or otherwise display UI, will be disabled in API-specific ways. In those tagged as being used for cross-origin prerendering, storage access will not be available, and all fetches will need to use the prerendering fetching modes.
 
   Crucially, prerendering browsing contexts have the ability to transition to becoming normal top-level browsing contexts, so that all of the prerendered content is reused and immediately displayed to the user.
 
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ We envision prerendering having several related pieces:
 
 * [**Prerendering browsing contexts**](./browsing-context.md), which are special browsing contexts that are not displayed to the user, and within which content is constrained to not perform disruptive or side-effecting operations.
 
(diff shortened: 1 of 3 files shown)


index.bs
```diff
@@ -59,6 +59,8 @@ Every {{Document}} has a <dfn for="Document">prerendering browsing contexts map<
 
 <p class="issue">Should this map be scoped to the [=user agent=] instead? Or, allowed to be copied between documents?
 
+Every {{Document}} has a <dfn for="Document">post-activation steps list</dfn>, which is a [=list=] where each [=list/item=] is a series of algorithm steps.
```
jeremyroman, line 62: I suspect at this granularity we'd want to use a more verbose term than "activation", since the term "active" is already used for other purposes (e.g., `document.activeElement`).

jeremyroman (approved):
lgtm

domfarolino (approved):
Spec lgtm

README.md
```diff
@@ -29,7 +29,7 @@ We envision prerendering having several related pieces:
 
 * [**Prerendering browsing contexts**](./browsing-context.md), which are special browsing contexts that are not displayed to the user, and within which content is constrained to not perform disruptive or side-effecting operations.
 
-  In all prerendering browsing contexts, side-effecting or disruptive APIs, such as those that could play media, require a permission prompt, or otherwise display UI, will automatically error or no-op. In those tagged as being used for cross-origin prerendering, storage access will not be available (or, perhaps, is partitioned?), and all fetches will need to use the prerendering fetching modes.
+  In all prerendering browsing contexts, side-effecting or disruptive APIs, such as those that could play media, require a permission prompt, or otherwise display UI, will be disabled in API-specific ways. In those tagged as being used for cross-origin prerendering, storage access will not be available, and all fetches will need to use the prerendering fetching modes.
```
jakearchibald, line 32: In this context does side-effecting mean it's observable from non-prerendered documents?

index.bs
```diff
@@ -108,6 +110,10 @@ Every {{Document}} has a <dfn for="Document">prerendering browsing contexts map<
     1. [=Queue a global task=] on the [=networking task source=] given |successorBC|'s [=browsing context/active window=] to perform the following steps
 
       1. Set |successorBC|'s [=browsing context/loading mode=] to "`default`".
+
+      1. [=list/For each=] |steps| in |successorBC|'s [=active document=]'s [=Document/post-activation steps list=], run |steps|.
```
jakearchibald, line 114: Do we need to handle exceptions in any particular way here?

jakearchibald (approved):
Couple of comments but LGTM

index.bs
domfarolino: Hmm, right now the only usage of this construct in this spec cannot throw any kind of an exception, but that's an interesting question in general. I kind of hope that any explicit exception handling could be done in the steps that appear in this list, and that we wouldn't actually have to handle anything ourselves, but I guess that might be too naive.

README.md
domenic: I guess it meant "user-affecting"? Let's just delete it; it's a fraught term.

index.bs
domenic: I'll add an assert that the specs don't throw an exception; at least as I'm currently imagining these, we'd never put exception-throwing steps in here.

PR #19, opened 2021-01-12
Initial document.prerendering spec text

bokand:
Adds some initial spec text for the `document.prerendering` API and its associated event.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -131,18 +131,13 @@ Amend the {{Document}} interface:
   };
 </pre>
 
-<dl dfn-type=attribute dfn-for=Document>
-  <dt>
-    <dfn>prerendering</dfn>
-    <dd>Is true if [=Document=] has a non-null [=Document/Browsing Context=] and its loading mode is "`prerender`" or "`uncredentialed-prerender`", false otherwise.</dd>
-    <p class="note">This attribute lets pages know when they're being presented in a non-interactive "prerendering-like" context. In the future, this would include a visible document in a `<portal>` element, both when loaded into it or via predecessor adoption.</p>
-  </dt>
 
-  <dt>
-    <dfn>onprerenderingchange</dfn>
-    <dd>This event handler, of event handler type `prerenderingchange`, MUST be fired whenever the value of the {{Document/prerendering}} attribute changes.</dd>
-  </dt>
-</dl>
+<p>The {{Document}} {{Document/prerendering}} attribute getter must return true if [=this=] has a non-null [=Document/Browsing Context=] whose [=browsing context/loading mode=] is "`prerender`" or "`uncredentialed-prerender`". Otherwise, returns false.</p>
+
+<p class="note">This attribute lets pages know when they're being presented in a non-interactive "prerendering-like" context. In the future, this would include a visible document in a `<portal>` element, both when loaded into it or via predecessor adoption.</p>
+
+<p>The {{Document}} {{Document/onprerenderingchange}} attribute is an [=event handler IDL attribute=] corresponding to the `prerenderingchange` [=event handler event type=].</p>
+
 
 <h2 id="navigation">Navigation and session history</h2>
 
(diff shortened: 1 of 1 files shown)


bokand:
One question I had...if my read of the [Document creation steps](https://html.spec.whatwg.org/multipage/browsing-the-web.html#initialise-the-document-object) and related concepts in HTML is correct, a Document won't have an associated BrowsingContext until the document is added to the BC's session history. In the steps, that happens a short while after the Document is created.

Technically, in that period of time, `document.prerendering` would return false as per my spec text - then, if it's added to a prerendering browsing context, this will change to true and so we should fire the `prerenderingchange` event. I expect none of this would be visible to script and it probably doesn't make sense to fire the event in this case (or maybe it does? curious if there's an argument for firing the event at creation) so currently we only fire the event during activation. Is the existing text enough or do we need to change the definition of the `prerenderingchange` event to specifically avoid this case?

index.bs
```diff
@@ -108,8 +108,32 @@ Every {{Document}} has a <dfn for="Document">prerendering browsing contexts map<
...
+<dl dfn-type=attribute dfn-for=Document>
+  <dt>
+    <dfn>prerendering</dfn>
+    <dd>Is true if [=Document=] has a non-null [=Document/Browsing Context=] and its loading mode is "`prerender`" or "`uncredentialed-prerender`", false otherwise.</dd>
+    <p class="note">This attribute lets pages know when they're being presented in a non-interactive "prerendering-like" context. In the future, this would include a visible document in a `<portal>` element, both when loaded into it or via predecessor adoption.</p>
+  </dt>
+
+  <dt>
+    <dfn>onprerenderingchange</dfn>
+    <dd>This event handler, of event handler type `prerenderingchange`, MUST be fired whenever the value of the {{Document/prerendering}} attribute changes.</dd>
```
domenic, line 133: This is not an appropriate place to put MUST requirements, especially one that overlaps (contradicts?) the above event firing. Instead this should just be defined like other IDL event handlers, e.g. https://html.spec.whatwg.org/#audiotracklist-and-videotracklist-objects:event-handler-idl-attributes:

> The `onprerenderingchange` attribute is an [=event handler IDL attribute=] corresponding to the `prerenderingchange` [=event handler event type=].

domenic (commented):
Right, so we should never fire events using "action at a distance" like "when something changes"; we should find the places that change, and fire the event from there. You've found at least one of them; probably that's the only one?

index.bs
bokand: Done.

index.bs
```diff
@@ -108,8 +108,32 @@ Every {{Document}} has a <dfn for="Document">prerendering browsing contexts map<
...
+Amend the {{Document}} interface:
+
+<pre class='idl'>
+  partial interface Document {
+      readonly attribute boolean prerendering;
+      attribute EventHandler onprerenderingchange;
+  };
+</pre>
+
+<dl dfn-type=attribute dfn-for=Document>
```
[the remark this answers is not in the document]
bokand: Done

bokand:
> You've found at least one of them; probably that's the only one?

I think so, until/if we add portal predecessor adoption.

domenic (approved):
Looks great. I pushed some editorial tweaks for more Bikeshed feature usage.

I rebased this on top of my PR that delays downloads until post-activation, and in particular I fired the event _before_ the "post-prerendering activation steps". Currently, since those steps only run downloads, so this order isn't observable (well, maybe with server-side timing corellation??).

Do you have thoughts on the ordering for the future, when it will contain things like "show a permissions prompt" or "fulfill pending async storage gets"?

bokand:
Hmm...I think firing `prerenderingchange` before running the activation steps is probably best - if those steps have any kind of side effects like running immediate script, it should probably look like we're out of prerendering and in an active browsing context.

bokand:
...and editorial tweaks look good, thanks!

PR #21, opened 2021-01-13
Remove referrer policy key from prerendering browsing contexts map

### LINKED ISSUE #18: What exactly are we trying to achieve with keying off of referrer policy?
While reviewing some internal threads, and @domenic [Prerendering activation matching](https://docs.google.com/document/d/1E82ZqXibjCvpPdOgMmOfPIatPMEBVAiJ2KQm6r1_E94/edit) doc (+ the discussion on it), I was trying to think of what exactly keying off of the referrer policy buys us.

I initially thought that it was to capture the fact that requests sent with different referrer policies from the same document will be sent with different `Referer`s, and thus could yield different responses from the server [1], so we shouldn't activate the same prerendered browsing context for both of these requests. In other words, we're using the referrer policy as a rough proxy for what the `Referer` would be, which then slightly intersects with `Vary: Referer` behavior. Furthermore, I thought it might buy us consistency with the value of document.referrer on the prerendered document. All of this assumes that two requests with the same referrer policy coming from the same document will be sent with the same `Referer` header though.

I then realized that it's pretty easy to send two requests from the same document that ultimately have the same referrer policy, but different `Referer`s if you use redirects. For example:
 1. A.com prerenders B.com {referrer policy: origin, `Referer: A.com`}
 1. A.com navigates to D.com/ToB {referrer policy: no-referrer, /no referrer/}
    - D.com redirects to B.com, attempting to activate the prerender
    - The redirect sets the referrer policy to `origin` to match the prerender
    - The request for B.com looks like {referrer policy: origin, /no referrer/}, and matches the prerender given our referrer policy criteria

This results in a successful prerender activation. Both the prerender and activation requests have the same referrer policy, but entirely different `Referer`s (in fact one was sent without this header altogether). This might not be a big deal, especially since it's an edge case, but I felt it was worth pointing out.

I guess the alternative would be to key off of the computed `Referer`, but that's basically the same as if ever prerender response had `Vary: Referer` (and we honored it in our activation logic), which seems pretty unlikely.

Keying off of the referrer policy just felt a little strange given the case I mentioned, and the fact that it seems to be acting as a pseudo-proxy for the `Referer` header I guess. I'm fine with it, but I just want to make sure I understand the intent fully.

[1]: This is the argument we made when trying to partition HTML's "list of available images" cache on the referrer policy as well, for consistency with other properties if nothing else

domfarolino:
This PR closes #18 by removing the referrer policy key and matching logic from the prerendering browsing context map.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4,6 +4,7 @@ Shortname: prerendering-revamped
 Status: DREAM
 Repository: jeremyroman/alternate-loading-modes
 Editor: Domenic Denicola, Google https://www.google.com/, d@domenic.me
+Editor: Dominic Farolino, Google https://www.google.com/, domfarolino@gmail.com
 Abstract: This document contains a collection of specification patches for well-specified prerendering.
 Markup Shorthands: css no, markdown yes
 Assume Explicit For: yes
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -55,26 +56,26 @@ By default, a [=browsing context=]'s [=browsing context/loading mode=] is "`defa
 
 A [=prerendering browsing context=] is <dfn for="prerendering browsing context">empty</dfn> if the only entry in its [=session history=] is the initial `about:blank` {{Document}}.
 
-Every {{Document}} has a <dfn for="Document">prerendering browsing contexts map</dfn>, which is an [=ordered map=] of ([=URL=], [=referrer policy=]) [=tuples=] to [=prerendering browsing contexts=]. This is used to fulfill [=navigate|navigations=] to a given URL by instead [=prerendering browsing context/activating=] the corresponding prerendering browsing context.
+Every {{Document}} has a <dfn for="Document">prerendering browsing contexts map</dfn>, which is an [=ordered map=] of [=URL=]s to [=prerendering browsing contexts=]. This is used to fulfill [=navigate|navigations=] to a given URL by instead [=prerendering browsing context/activating=] the corresponding prerendering browsing context.
```
domenic, line 59:
```suggestion
Every {{Document}} has a <dfn for="Document">prerendering browsing contexts map</dfn>, which is an [=ordered map=] of [=URLs=] to [=prerendering browsing contexts=]. This is used to fulfill [=navigate|navigations=] to a given URL by instead [=prerendering browsing context/activating=] the corresponding prerendering browsing context.
```

index.bs
```diff
@@ -55,26 +56,26 @@ By default, a [=browsing context=]'s [=browsing context/loading mode=] is "`defa
...
 
-<p class="issue">Should this map be scoped to the [=user agent=] instead? Or, allowed to be copied between documents?
+<p class="note">This spec scopes the [=Document/prerendering browsing contexts map=] to a {{Document}} for simplicity. This is described in more in the <a href="https://docs.google.com/document/d/1E82ZqXibjCvpPdOgMmOfPIatPMEBVAiJ2KQm6r1_E94/edit#heading=h.fvk61aveq6y9">Prerendering activation matching</a> document but is subject to change in the future.
```
domenic, line 61:
```suggestion
<p class="note">This spec scopes the [=Document/prerendering browsing contexts map=] to a {{Document}} for simplicity. This is discussed in more detail in the <a href="https://docs.google.com/document/d/1E82ZqXibjCvpPdOgMmOfPIatPMEBVAiJ2KQm6r1_E94/edit#heading=h.fvk61aveq6y9">Prerendering activation matching</a> document but is subject to change in the future.
```

domenic (approved):
LGTM with nits

domenic:
Hmm, although I just say https://github.com/jeremyroman/alternate-loading-modes/issues/18#issuecomment-759703856, maybe we should resolve that first...

domfarolino:
Per more discussion, I think we're going to stick with this for now so I'll close this.

PR #22, opened 2021-01-14
Add a simple prerender processing model

domfarolino:
This PR adds a simple "starter" `<link rel=prerender>` processing model section that:
 - Defines the new `rel=prerender` link type that establishes an [#external-resource-link](https://html.spec.whatwg.org/multipage/links.html#external-resource-link)
 - Overrides the [#default-fetch-and-process-the-linked-resource] algorithm for this link type to create a prerendering browsing context
 - Adds reasonable (at the time of writing this) "appropriate times" to invoke the processing model algorithm (these appropriate times are somewhat inspired from other link types in the HTML Standard)
 - Documents the fact that soon we're going to define a "sufficiently strict referrer policy set" which will be used to restrict prerenders to those whose `referrerpolicy` attr is in the set

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12,6 +12,9 @@ Complain About: accidental-2119 yes, missing-example-ids yes
 Indent: 2
 Boilerplate: omit conformance
 </pre>
+<pre class="link-defaults">
+spec:html; type:element; text:link
+</pre>
 <pre class="anchors">
 spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
@@ -30,14 +33,9 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: abort; for: Document; url: abort-a-document
       text: create and initialize a Document object; url: initialise-the-document-object
       text: history handling behavior; url: history-handling-behavior
-      text: link; url: the-link-element
       text: prompt to unload; url: prompt-to-unload-a-document
       text: refused to allow the document to be unloaded; url: refused-to-allow-the-document-to-be-unloaded
       text: traverse the history; url: traverse-the-history
-    urlPrefix: semantics.html
-      for: link
-        text: href; url: attr-link-href
-        text: referrerpolicy; url: attr-link-referrerpolicy
     urlPrefix: urls-and-fetching.html
       text: parse a URL; url: parse-a-url
       text: resulting URL record
@@ -45,7 +43,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
 
 <h2 id="prerendering-bcs">Prerendering browsing contexts</h2>
 
-<em>The following section would be added as a new sub-section of [[HTML]]'s <a href="https://html.spec.whatwg.org/#windows">Browsing contexts</a> section.</em>
+<em>The following section would be added as a new sub-section of [[HTML]]'s <a href="https://html.spec.whatwg.org/multipage/browsers.html#windows">Browsing contexts</a> section.</em>
 
 Every [=browsing context=] has a <dfn for="browsing context">loading mode</dfn>, which is one of the following:
 
@@ -86,8 +84,6 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps 
   1. Let |request| be a new [=request=] whose [=request/URL=] is |startingURL| and [=request/referrer policy=] is |referrerPolicy|.
 
   1. [=Navigate=] |bc| to |request| with the [=source browsing context=] set to |referrerDoc|'s [=Document/browsing context=].
-
-  1. Return |bc|.
 </div>
 
 <div algorithm>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -29,9 +30,17 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
       text: create and initialize a Document object; url: initialise-the-document-object
       text: history handling behavior; url: history-handling-behavior
+      text: link; url: the-link-element
       text: prompt to unload; url: prompt-to-unload-a-document
       text: refused to allow the document to be unloaded; url: refused-to-allow-the-document-to-be-unloaded
       text: traverse the history; url: traverse-the-history
+    urlPrefix: semantics.html
+      for: link
+        text: href; url: attr-link-href
+        text: referrerpolicy; url: attr-link-referrerpolicy
```
domfarolino, line 40: I _think_ this is the best way to achieve what I want, but let me know if you think there's a better way

index.bs
```diff
@@ -140,6 +149,25 @@ The <dfn attribute for="Document">prerendering</dfn> getter steps are to return
...
+<h3 id="link-rel-prerender">Link type "<dfn data-lt="rel-prerender"><code>prerender</code></dfn>"</h3>
+
+The <a lt="rel-prerender"><code>prerender</code></a> keyword may be used with <a spec=HTML><code>link</code></a> elements. This keyword creates an <a spec=HTML>external resource link</a>. This keyword is <a spec=HTML>body-ok</a>.
+
+To <a lt="process the linked resource" spec=HTML>process this type of linked resource</a> given a <a spec=HTML><code>link</code></a> element <var ignore>el</var>, run these steps:
+
+<div algorithm="prerender-processing-model">
+  1. If |el|'s [=link/href=] attribute's value is the empty string, then return.
+  1. Let |referrer policy| be  the current state of |el|'s [=link/referrerpolicy=] attribute.
+  1. [=parse a URL|Parse=] the [=URL=] given by |el|'s [=link/href=] attribute, relative to |el|'s [=Node/node document=]. If that fails, then return. Otherwise, let |url| be the [=resulting URL record=].
```
domfarolino, line 165: As a slight personal preference I've decided to keep the linking text "parse a URL" and just override it with "Parse" here, but I'm open to just changing the link text to Parse if it is preferred.

index.bs
domenic: You should just be able to do `<{link/href}>` and `<{link/referrerpolicy}>`, without these. Same for `<{link}>`, without the above.

index.bs
```diff
@@ -140,6 +149,25 @@ The <dfn attribute for="Document">prerendering</dfn> getter steps are to return
 The <dfn attribute for="Document">onprerenderingchange</dfn> attribute is an [=event handler IDL attribute=] corresponding to the <dfn event for="Document">prerenderingchange</dfn> [=event handler event type=].
 
 
+<h2 id="prerender-processing-model">Prerender processing model</h2>
```
domenic, line 152: Let's delete this heading, and just move the sentence "Add a new..." into the "Link type prerender" section? Compare to https://jeremyroman.github.io/alternate-loading-modes/#prerendering-bcs

index.bs
```diff
@@ -140,6 +149,25 @@ The <dfn attribute for="Document">prerendering</dfn> getter steps are to return
...
+
+Add a new link type to the "Link types" section of the HTML Standard, replacing the prerender documentation in the Resource Hints standard.
+
+<h3 id="link-rel-prerender">Link type "<dfn data-lt="rel-prerender"><code>prerender</code></dfn>"</h3>
```
domenic, line 156:
```suggestion
<h3 id="link-rel-prerender">Link type "<dfn attr-value for="link/rel"><code>prerender</code></dfn>"</h3>
```

then reference it as, I believe, `<{link/rel/prerender}>` elsewhere?

index.bs
```diff
@@ -140,6 +149,25 @@ The <dfn attribute for="Document">prerendering</dfn> getter steps are to return
...
+
+The <a lt="rel-prerender"><code>prerender</code></a> keyword may be used with <a spec=HTML><code>link</code></a> elements. This keyword creates an <a spec=HTML>external resource link</a>. This keyword is <a spec=HTML>body-ok</a>.
+
+To <a lt="process the linked resource" spec=HTML>process this type of linked resource</a> given a <a spec=HTML><code>link</code></a> element <var ignore>el</var>, run these steps:
```
domenic, line 160: Put this inside the `<div algorithm>`, and then you can use `|el|`.

index.bs
```diff
@@ -140,6 +149,25 @@ The <dfn attribute for="Document">prerendering</dfn> getter steps are to return
...
+
```
domenic, line 161: We should add appropriate times to fetch and process the linked resource. Probably they are UA-defined? Or should we do some of the usual stuff (when it becomes BC-connected, or its relevant attributes change) as something like when the UA should *consider* doing a fetch-and-process?

index.bs
domenic, line 160: I think we want to override all of fetch-and-process, right? Not just process?

index.bs
domenic, line 165: So, modulepreload does this wrong. It should be "[Parse the URL] given by...". Stating that it's a `[=URL=]` is incorrect since _el_'s `href` attribute is a string, not a URL record. "Parse the URL" would be better named as "Parse the URL string", really...

index.bs
```diff
@@ -140,6 +149,25 @@ The <dfn attribute for="Document">prerendering</dfn> getter steps are to return
...
+<div algorithm="prerender-processing-model">
+  1. If |el|'s [=link/href=] attribute's value is the empty string, then return.
+  1. Let |referrer policy| be  the current state of |el|'s [=link/referrerpolicy=] attribute.
+  1. [=parse a URL|Parse=] the [=URL=] given by |el|'s [=link/href=] attribute, relative to |el|'s [=Node/node document=]. If that fails, then return. Otherwise, let |url| be the [=resulting URL record=].
+  1. [=Create a prerendering browsing context=] with |url|, |referrer policy|, and |el|'s [=Node/node document=].
```
domenic, line 166: Maybe, since it turns out we don't need the return value, update that algorithm to remove the "return _bc_" step?

domenic (commented):
Exciting times! Mostly editorial comments, a number around Bikeshed linking.

index.bs
```diff
@@ -140,6 +149,25 @@ The <dfn attribute for="Document">prerendering</dfn> getter steps are to return
...
+</div>
```
domenic, line 167: Maybe add a note that we intentionally don't fire any load or error events, since comparing to other link types, I might expect to see such steps.

index.bs
domfarolino: Yep, I messed that up nice catch.

index.bs
domfarolino: Yikes, I actually got that from https://html.spec.whatwg.org/multipage/semantics.html#default-fetch-and-process-the-linked-resource but yeah that makes sense. I'll fix this and those two instances in HTML.

index.bs
domfarolino: Nice catch. Done.

index.bs
domfarolino: Good call, I also added a sentence mentioning that UAs should not delay the load event of the link's node document.

index.bs
domfarolino: Yeah, I think the usual stuff is best for now, especially since our implementation isn't really UA-defined... at least I think it's taking the existence of the link to be a directive to perform all the fetching and prerendering etc.

domfarolino: The wording I used for this matches https://html.spec.whatwg.org/multipage/links.html#link-type-modulepreload which seems pretty standard, and I think matches preload's though it is worded slightly differently

index.bs
```diff
@@ -140,6 +145,38 @@ The <dfn attribute for="Document">prerendering</dfn> getter steps are to return
 The <dfn attribute for="Document">onprerenderingchange</dfn> attribute is an [=event handler IDL attribute=] corresponding to the <dfn event for="Document">prerenderingchange</dfn> [=event handler event type=].
 
 
+<h2 id="link-rel-prerender">Link type "<dfn attr-value for="link/rel"><code>prerender</code></dfn>"</h2>
+
+<em>The following section would be added as a new sub-section of [[HTML]]'s <a href="https://html.spec.whatwg.org/multipage/links.html#linkTypes">Link types</a> section. This should replace the prerender documentation in the Resource Hints standard</em>
```
domenic, line 150:
```suggestion
<em>The following section would be added as a new sub-section of [[HTML]]'s <a href="https://html.spec.whatwg.org/multipage/links.html#linkTypes">Link types</a> section. This should replace the prerender documentation in [[RESOURCE-HINTS]].</em>
```

(I'm just guessing that [[RESOURCE-HINTS]] is the right name...)

index.bs
```diff
@@ -140,6 +145,38 @@ The <dfn attribute for="Document">prerendering</dfn> getter steps are to return
...
+
+The <{link/rel/prerender}> keyword may be used with <{link}> elements. This keyword creates an <a spec=HTML>external resource link</a>. This keyword is <a spec=HTML>body-ok</a>.
+
+The appropriate times to <a spec=HTML>fetch and process the linked resource</a> for such a link are:
```
domenic, line 154: I think some of these (I only checked "browsing-context connected") are exported, so you can use `[=browsing-context connected=]` instead of `<a spec="HTML">browsing context connected</a>`. The `spec=HTML` is a hack to get access to non-exported terms without having to put them in the links block at the top.

index.bs
```diff
@@ -140,6 +145,38 @@ The <dfn attribute for="Document">prerendering</dfn> getter steps are to return
...
+
+<ul>
+  <li>When the <a spec=HTML>external resource link</a> is created on a <{link}> element that is already <a spec=HTML>browsing-context connected</a>.</li>
+  <li>When the <a spec=HTML>external resource link</a>'s <{link}> element <a spec=HTML>becomes browsing-context connected</a>.</li>
+  <li>When the <{link/href}> attribute of the <{link}> element of an <a spec=HTML>external resource link</a> that is already <a spec=HTML>browsing-context connected</a> is changed.</li>
```
domenic, line 159: Should also react to referrerpolicy changes, I think?

index.bs
```diff
@@ -140,6 +145,38 @@ The <dfn attribute for="Document">prerendering</dfn> getter steps are to return
...
+  1. [=Parse a URL=] given |el|'s <{link/href}> attribute, relative to |el|'s [=Node/node document=]. If that fails, then return. Otherwise, let |url| be the [=resulting URL record=].
+
+  1. Let |referrer policy| be  the current state of |el|'s <{link/referrerpolicy}> attribute.
+
+  1. [=Create a prerendering browsing context=] with |url|, |referrer policy|, and |el|'s [=Node/node document=].
+</div>
+
+A user agent must not <a spec=HTML>delay the load event</a> of the <{link}> element's <a for=Node>node document</a> for this link type.
+
+<p class="note">Note that this type of link does not fire {{HTMLElement/error}} or {{HTMLElement/load}} events, unlike many other links that create an <a spec=HTML>external resource links</a>.
```
domenic, line 176:
```suggestion
<p class="note">Note that this type of link does not fire {{HTMLElement/error}} or {{HTMLElement/load}} events, unlike many other link types that create <a spec=HTML>external resource links</a>.
```

domenic (approved):
Editorial nits + referrerpolicy checking

index.bs
domfarolino: Nice, Done.

index.bs
domfarolino: Hmm, normally I'd say no for links in general, but I guess since the processing model will potentially fail to prerender links if the referrer policy is not strict enough it seems OK? I'll go ahead and add it.

PR #24, opened 2021-01-21
First draft of a spec for speculation rules for prefetch

jeremyroman:
This is a draft of a minimal version of the speculation rules proposal, which covers only how to trigger prefetching with an optional "requires anonymous client IP" flag for a fixed list of URLs.

I think the bulk of the complexity is in the stuff that's independent of trigger, i.e. "what does it mean to prefetch".

@domenic @domfarolino @mfalken does this seem reasonable?

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/pull/24.html" title="Last updated on Jan 26, 2021, 7:44 PM UTC (954fba6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/24/9aebf57...954fba6.html" title="Last updated on Jan 26, 2021, 7:44 PM UTC (954fba6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -200,6 +200,8 @@ Periodically, for any [=document=] |document|, the user agent may [=queue a glob
 <div algorithm="consider speculation">
   To <dfn>consider speculation</dfn> for a [=document=] |document|:
 
+  1. If |document| is not [=Document/fully active=], then return.
+     <p class="issue">It's likely that we should also handle prerendered and back-forward cached documents.
   1. For each |ruleSet| of |document|'s [=document/list of speculation rule sets=]:
     1. [=list/For each=] |rule| of |ruleSet|'s [=speculation rule set/prefetch rules=]:
       1. Let |requiresAnonymousClientIP| be true if |rule|'s [=speculation rule/requirements=] [=set/contains=] "`anonymous-client-ip`", and false otherwise.
--- a/index.bs
+++ b/index.bs
@@ -205,7 +205,7 @@ Periodically, the user agent may [=queue a task=] to the <a data-link-type="inte
       1. Let |requiresAnonymousClientIP| be true if |rule|'s [=speculation rule/requirements=] [=set/contains=] "`anonymous-client-ip`", and false otherwise.
       1. [=list/For each=] |url| of |rule|'s [=speculation rule/URLs=]:
         1. The user agent may prefetch |url| given |requiresAnonymousClientIP|.
-           <p class="issue">TODO: expand this to actually elaborate on how prefetch works, once initiated, and to incorporate the |requiresAnonymousClientIP| flag.
+           <p class="issue">TODO: expand this to actually elaborate on how prefetch works, once initiated, and to incorporate the |requiresAnonymousClientIP| flag. We may wish to include language about when the UA should deduplicate requests.
 </div>
 
 <p class="issue">
--- a/index.bs
+++ b/index.bs
@@ -156,7 +156,7 @@ Inside the [=prepare a script=] algorithm we make the following changes:
 <div algorithm="parse speculation rules">
   To <dfn>parse speculation rules</dfn> given a [=string=] |input| and a [=URL=] |baseURL|, perform the following steps. They return a [=speculation rule set=] or null.
 
-  1. Let |parsed| be the result of [=Parse JSON into Infra values|parsing JSON into Infra values=] given |input|.
+  1. Let |parsed| be the result of [=parsing a JSON string to an Infra value=] given |input|.
   1. If |parsed| is not a [=map=], then return null.
   1. Let |result| be an empty [=speculation rule set=].
   1. If |parsed|["`prefetch`"] [=map/exists=] and is a [=list=], then [=list/for each=] |prefetchRule| of |parsed|["`prefetch`"]:
@@ -187,12 +187,12 @@ Inside the [=prepare a script=] algorithm we make the following changes:
   1. Return a [=speculation rule=] with [=speculation rule/URLs=] |urls| and [=speculation rule/requirements=] |requirements|.
 </div>
 
-<h3 id="speculation-rules-processing">Processing Model</h3>
+<h3 id="speculation-rules-processing">Processing model</h3>
 
 A [=document=] has a <dfn for=document export>list of speculation rule sets</dfn>, which is an initially empty [=list=].
 
 <!-- TODO(domfarolino): Get rid of the `data-link-type="interface"` once we fix the dfn in HTML. -->
-Periodically, the user agent may [=queue a task=] to the <a data-link-type="interface">DOM manipulation task source</a> to [=consider speculation=] for any [=document=].
+Periodically, for any [=document=] |document|, the user agent may [=queue a global task=] on the <a data-link-type="interface">DOM manipulation task source</a> with |document|'s [=relevant global object=] to [=consider speculation=] for |document|.
 
 <p class="note">
   The user agent will likely do this after the insertion of new speculation rules, or when resources are idle and available.
--- a/index.bs
+++ b/index.bs
@@ -154,7 +154,7 @@ Inside the [=prepare a script=] algorithm we make the following changes:
   This reduces the risk of unintended activity by user agents which are unaware of most recently added directives which might limit the scope of a rule.
 
 <div algorithm="parse speculation rules">
-  To <dfn export>parse speculation rules</dfn> given a [=string=] |input| and a [=URL=] |baseURL|, perform the following steps. They return a [=speculation rule set=] or null.
+  To <dfn>parse speculation rules</dfn> given a [=string=] |input| and a [=URL=] |baseURL|, perform the following steps. They return a [=speculation rule set=] or null.
 
   1. Let |parsed| be the result of [=Parse JSON into Infra values|parsing JSON into Infra values=] given |input|.
   1. If |parsed| is not a [=map=], then return null.
@@ -168,7 +168,7 @@ Inside the [=prepare a script=] algorithm we make the following changes:
 </div>
 
 <div algorithm="parse a speculation rule">
-  To <dfn export>parse a speculation rule</dfn> given a [=map=] |input| and a [=URL=] |baseURL|, perform the following steps. They return a [=speculation rule=] or null.
+  To <dfn>parse a speculation rule</dfn> given a [=map=] |input| and a [=URL=] |baseURL|, perform the following steps. They return a [=speculation rule=] or null.
 
   1. If |input| has any [=map/key=] other than "`source`", "`urls`", and "`requires`", then return null.
   1. If |input|["`source`"] does not [=map/exist=] or is not the [=string=] "`list`", then return null.
@@ -198,7 +198,7 @@ Periodically, the user agent may [=queue a task=] to the <a data-link-type="inte
   The user agent will likely do this after the insertion of new speculation rules, or when resources are idle and available.
 
 <div algorithm="consider speculation">
-  To <dfn export>consider speculation</dfn> for a [=document=] |document|:
+  To <dfn>consider speculation</dfn> for a [=document=] |document|:
 
   1. For each |ruleSet| of |document|'s [=document/list of speculation rule sets=]:
     1. [=list/For each=] |rule| of |ruleSet|'s [=speculation rule set/prefetch rules=]:
--- a/index.bs
+++ b/index.bs
@@ -204,7 +204,7 @@ Periodically, the user agent may [=queue a task=] to the <a data-link-type="inte
     1. [=list/For each=] |rule| of |ruleSet|'s [=speculation rule set/prefetch rules=]:
       1. Let |requiresAnonymousClientIP| be true if |rule|'s [=speculation rule/requirements=] [=set/contains=] "`anonymous-client-ip`", and false otherwise.
       1. [=list/For each=] |url| of |rule|'s [=speculation rule/URLs=]:
-        1. The user agent MAY prefetch |url| given |requiresAnonymousClientIP|.
+        1. The user agent may prefetch |url| given |requiresAnonymousClientIP|.
            <p class="issue">TODO: expand this to actually elaborate on how prefetch works, once initiated, and to incorporate the |requiresAnonymousClientIP| flag.
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -100,7 +100,7 @@ A <dfn>speculation rule set</dfn> is a [=struct=] with the following [=struct/it
 
 Add "`speculationrules`" to the list of valid values for <a spec=html>the script's type</a>.
 
-Rename [=the script's script=] to <dfn>the script's result</dfn>, which can be a [=speculation rule set=] in addition to any other valid type.
+Rename [=the script's script=] to <dfn>the script's result</dfn>, which can be either a script or a [=speculation rule set=].
 
 Inside the [=prepare a script=] algorithm:
 
@@ -113,15 +113,23 @@ Inside the [=prepare a script=] algorithm:
     <dd>
       1. Let |result| be the result of [=parse speculation rules|parsing speculation rules=] given source text and base URL.
 
-      1. If |result| is not null, set [=the script's result=] to |result|.
+      1. Set [=the script's result=] to |result|.
 
       1. <a spec=html>The script is ready</a>.
     </dd>
   </dl>
 
-Inside the <a spec=html>execute a script block</a> algorithm:
+* Insert the following case to the switch in the subsequent step beginning "Then, follow the first of the following options...." after the cases which apply only to "`classic`" and "`module`" scripts:
+  <dl>
+    <dt>If <a spec=html>the script's type</a> is "`speculationrules`"</dt>
+    <dd>
+      1. When <a spec=html>the script is ready</a>, run the following steps:
 
-<p class="issue">Adjust this consistent with the resolution of <a href="https://github.com/WICG/import-maps/issues/243">WICG/import-maps#243</a>.</p>
+        1. If [=the script's result=] is not null, [=list/append=] it to the element's [=node document=]'s [=document/list of speculation rule sets=].
+    </dd>
+  </dl>
+
+Inside the <a spec=html>execute a script block</a> algorithm:
 
 * Add the following case to the switch on <a spec=html>the script's type</a>:
   <dl>
@@ -133,10 +141,9 @@ Inside the <a spec=html>execute a script block</a> algorithm:
 
 <h3 id="speculation-rules-parsing">Parsing</h3>
 
-<div class="note">
+<p class="note">
   The general principle here is to allow the existence of directives which are not understood, but not to accept into the rule set a rule which the user agent does not fully understand.
   This reduces the risk of unintended activity by user agents which are unaware of most recently added directives which might limit the scope of a rule.
-</div>
 
 <div algorithm="parse speculation rules">
   To <dfn export>parse speculation rules</dfn> given a [=string=] |input| and a [=URL=] |baseURL|, perform the following steps. They return a [=speculation rule set=] or null.
@@ -174,26 +181,26 @@ Inside the <a spec=html>execute a script block</a> algorithm:
 
 <h3 id="speculation-rules-processing">Processing Model</h3>
 
+A [=document=] has a <dfn for=document export>list of speculation rule sets</dfn>, which is an initially empty [=list=].
+
 Periodically, the user agent may [=queue a task=] to the [=DOM manipulation task source=] to [=consider speculation=] for any [=document=].
 
-<div class="note">
+<p class="note">
   The user agent will likely do this after the insertion of new speculation rules, or when resources are idle and available.
-</div>
 
 <div algorithm="consider speculation">
   To <dfn export>consider speculation</dfn> for a [=document=] |document|:
 
-  1. For each <{script}> element [=connected=] to |document| where <a spec=html>the script's type</a> is "`speculationrules`" and [=the script's result=] is a [=speculation rule set=]:
-    1. [=list/For each=] |rule| of [=the script's result=]:
+  1. For each |ruleSet| of |document|'s [=document/list of speculation rule sets=]:
+    1. [=list/For each=] |rule| of |ruleSet|'s [=speculation rule set/prefetch rules=]:
       1. Let |requiresAnonymousClientIP| be true if |rule|'s [=speculation rule/requirements=] [=set/contains=] "`anonymous-client-ip`", and false otherwise.
-      1. [=list/For each=] |url| of |rule|'s [=speculation rule/URLs=]
+      1. [=list/For each=] |url| of |rule|'s [=speculation rule/URLs=]:
         1. The user agent MAY prefetch |url| given |requiresAnonymousClientIP|.
-           <div class="issue">TODO: expand this to actually elaborate on how prefetch works, once initiated, and to incorporate the |requiresAnonymousClientIP| flag.</div>
+           <p class="issue">TODO: expand this to actually elaborate on how prefetch works, once initiated, and to incorporate the |requiresAnonymousClientIP| flag.
 </div>
 
-<div class="issue">
+<p class="issue">
   We should also notice removals and consider cancelling speculated actions.
-</div>
 
 <h2 id="prerendering-bcs">Prerendering browsing contexts</h2>
 
--- a/index.bs
+++ b/index.bs
@@ -128,7 +128,7 @@ Inside the [=prepare a script=] algorithm we make the following changes:
   <dl>
     <dt>"`speculationrules`"</dt>
     <dd>
-      1. Let |result| be the result of [=parse speculation rules|parsing speculation rules=] given source text and base URL.
+      1. Let |result| be the result of [=parsing speculation rules=] given source text and base URL.
 
       1. Set [=the script's result=] to |result|.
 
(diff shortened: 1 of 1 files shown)


jeremyroman:
Not sure why Bikeshed is no longer happy with "DOM manipulation task source". It worked on the old version of Bikeshed before I updated...

index.bs
```diff
@@ -74,6 +75,125 @@ A user agent must not <a spec=HTML>delay the load event</a> of the <{link}> elem
...
+
+      1. If |result| is not null, set [=the script's result=] to |result|.
+
+      1. <a spec=html>The script is ready</a>.
+    </dd>
+  </dl>
+
+Inside the <a spec=html>execute a script block</a> algorithm:
+
+<p class="issue">Adjust this consistent with the resolution of <a href="https://github.com/WICG/import-maps/issues/243">WICG/import-maps#243</a>.</p>
```
domfarolino, line 124: As per https://github.com/WICG/import-maps/issues/243#issuecomment-764452366, I think we can remove this issue but I believe we'll need to add some readiness handling like https://wicg.github.io/import-maps/#ref-for-concept-script-type%E2%91%A2.

If we don't "handle" this script type in [prepare a script](https://html.spec.whatwg.org/multipage/scripting.html#prepare-a-script) step 28, then the "Otherwise" block will trigger, thus executing the script immediately even if other scripts are running, which seems particularly bad here. So we should have some sort of "type=importmap" handling here. Exactly what that handling will do ties into a comment I have below, so please continue reading there.

index.bs
```diff
@@ -74,6 +75,125 @@ A user agent must not <a spec=HTML>delay the load event</a> of the <{link}> elem
...
+  <dl>
+    <dt>"`speculationrules`"</dt>
+    <dd>
+      1. [=Assert=]: Never reached.
+    </dd>
+  </dl>
+
+<h3 id="speculation-rules-parsing">Parsing</h3>
+
+<div class="note">
```
domfarolino, line 136: I think it's more common to have `<p class="note">`, or at least it is more consistent with the notes in this spec. Also feel free to omit the end tag like other notes here.

index.bs
```diff
@@ -74,6 +75,125 @@ A user agent must not <a spec=HTML>delay the load event</a> of the <{link}> elem
...
+The only valid string for [=speculation rule/requirements=] to contain is "`anonymous-client-ip`".
+
+A <dfn>speculation rule set</dfn> is a [=struct=] with the following [=struct/items=]:
+* <dfn for="speculation rule set">prefetch rules</dfn>, a [=list=] of [=speculation rules=]
+
+<h3 id="speculation-rules-script">The <{script}> element</h3>
+
+<em>Note</em>: This section contains modifications to the corresponding section of [[HTML]].
+
+<p class="issue">We should consider whether we also want to make this execute even if scripting is disabled.</p>
```
domfarolino, line 95: nit: Feel free to omit the end tag as is consistent with the other notes in this spec

index.bs
```diff
@@ -74,6 +75,125 @@ A user agent must not <a spec=HTML>delay the load event</a> of the <{link}> elem
...
+
+  1. For each <{script}> element [=connected=] to |document| where <a spec=html>the script's type</a> is "`speculationrules`" and [=the script's result=] is a [=speculation rule set=]:
+    1. [=list/For each=] |rule| of [=the script's result=]:
+      1. Let |requiresAnonymousClientIP| be true if |rule|'s [=speculation rule/requirements=] [=set/contains=] "`anonymous-client-ip`", and false otherwise.
+      1. [=list/For each=] |url| of |rule|'s [=speculation rule/URLs=]
+        1. The user agent MAY prefetch |url| given |requiresAnonymousClientIP|.
+           <div class="issue">TODO: expand this to actually elaborate on how prefetch works, once initiated, and to incorporate the |requiresAnonymousClientIP| flag.</div>
+</div>
+
+<div class="issue">
```
domfarolino, line 194: I think it's more common to have `<p class="issue">`, or at least it is more consistent with the notes in this spec. Also feel free to omit the end tag like other notes/issues here.

index.bs
```diff
@@ -74,6 +75,125 @@ A user agent must not <a spec=HTML>delay the load event</a> of the <{link}> elem
...
+Periodically, the user agent may [=queue a task=] to the [=DOM manipulation task source=] to [=consider speculation=] for any [=document=].
+
+<div class="note">
+  The user agent will likely do this after the insertion of new speculation rules, or when resources are idle and available.
+</div>
+
+<div algorithm="consider speculation">
+  To <dfn export>consider speculation</dfn> for a [=document=] |document|:
+
+  1. For each <{script}> element [=connected=] to |document| where <a spec=html>the script's type</a> is "`speculationrules`" and [=the script's result=] is a [=speculation rule set=]:
```
domfarolino, line 186: My high-level feedback for this algorithm is a continuation of my comment above. I think we should model the way other scripts handle their "readiness", by marking the script as ready and adding the "result" to some per-document list of "available speculation rules" or something. That way we can just access that list of results here and use them instead of having to iterate through all connected scripts whose type = [...] and is result is non-null etc etc. Does that make sense? If that's unclear at all I'm happy to add what I'm thinking directly to this PR or help clarify what I'm getting at.

Also has per my one comment above, I think this will be necessary to avoid the "Otherwise" block from immediately executing the speculation rules scripts, assuming I'm reading that right.

domfarolino (changes requested):
Some high-level feedback for now. I haven't looked at the parsing section yet, but mostly checked out the script processing model bits. I like the direction it's heading in!

domfarolino:
> Not sure why Bikeshed is no longer happy with "DOM manipulation task source". It worked on the old version of Bikeshed before I updated...

Yeah this is really weird. Bikeshed is interpreting this definition as an "interface" for some reason. I was able to get it to compile by writing `{{DOM manipulation task source}}` instead of `[= ... =]`, which seems wrong. I did a little digging around and found the discrepancy below, but I'm not sure if it is related since it looks like the files haven't been updated super recently:
 - https://github.com/tabatkins/bikeshed-data/blob/c69021a3a5779a923c422e190d3c26ef2df2b87c/data/anchors/anchors-do.data#L166
 - https://github.com/tabatkins/bikeshed-data/blob/eca3174e065b85b778b77da9b23076d88e1b41ad/data/anchors/anchors-ne.data#L889

@tabatkins do you know what's going on?

index.bs
```diff
@@ -74,6 +75,125 @@ A user agent must not <a spec=HTML>delay the load event</a> of the <{link}> elem
...
+
+<p class="issue">We should consider whether we also want to make this execute even if scripting is disabled.</p>
+
+<p class="issue">We should also incorporate the case where a {{HTMLScriptElement/src}} attribute is set.</p>
+
+<p class="issue">We could fire {{HTMLElement/error}} and {{HTMLElement/load}} events if we wanted to.</p>
+
+Add "`speculationrules`" to the list of valid values for <a spec=html>the script's type</a>.
+
+Rename [=the script's script=] to <dfn>the script's result</dfn>, which can be a [=speculation rule set=] in addition to any other valid type.
```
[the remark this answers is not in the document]
jeremyroman: Sure. Yeah, I was trying to make the merge with import maps slightly more obvious, but it's not like it's hard to merge two unions. :)

index.bs
```diff
@@ -74,6 +75,125 @@ A user agent must not <a spec=HTML>delay the load event</a> of the <{link}> elem
...
+Inside the [=prepare a script=] algorithm:
+
+* Insert the following step as the second-last sub-step under "Determine the script's type as follows:":
+  * If the script block's type string is an [=ASCII case-insensitive=] match for the string "`speculationrules`", <a spec=html>the script's type</a> is "`speculationrules`".
+
+* Insert the following case in the switch on <a spec=html>the script's type</a> within the step which begins "If the element does not have a {{HTMLScriptElement/src}} content attribute..."
+  <dl>
+    <dt>"`speculationrules`"</dt>
+    <dd>
+      1. Let |result| be the result of [=parse speculation rules|parsing speculation rules=] given source text and base URL.
```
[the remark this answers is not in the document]
jeremyroman: That's exactly what I gave. I assume you mean `[=parsing speculation rules=]`, but it doesn't generate a link and produces

```
LINE ~114: No 'dfn' refs found for 'parsing speculation rules'.
```

index.bs
```diff
@@ -74,6 +75,125 @@ A user agent must not <a spec=HTML>delay the load event</a> of the <{link}> elem
...
+  The user agent will likely do this after the insertion of new speculation rules, or when resources are idle and available.
+</div>
+
+<div algorithm="consider speculation">
+  To <dfn export>consider speculation</dfn> for a [=document=] |document|:
+
+  1. For each <{script}> element [=connected=] to |document| where <a spec=html>the script's type</a> is "`speculationrules`" and [=the script's result=] is a [=speculation rule set=]:
+    1. [=list/For each=] |rule| of [=the script's result=]:
+      1. Let |requiresAnonymousClientIP| be true if |rule|'s [=speculation rule/requirements=] [=set/contains=] "`anonymous-client-ip`", and false otherwise.
+      1. [=list/For each=] |url| of |rule|'s [=speculation rule/URLs=]
```
mfalken, line 189: nit: missing colon at the end

mfalken (commented):
I took a look and didn't have any substantial comment, looks good and Dominic is already giving good comments. Thanks for adding me, I'll keep watching but you don't need to block on me.

index.bs
jeremyroman: done

index.bs
jeremyroman: done

index.bs
jeremyroman: done

index.bs
jeremyroman: Sure, I was on the fence about whether to have an explicit list or not. I think I would like the ability to remove them which will require hooking the removal of script elements, but that's not hard to do. (And I don't really have to do it in this PR.)

index.bs
jeremyroman: done

jeremyroman (commented):
Apparently I used GitHub wrong and sent a bunch of individual comments. That aside, I think I've applied the requested changes. `[=node document=]` is now unrecognized too for some reason...

tabatkins:
> Yeah this is really weird. Bikeshed is interpreting this definition as an "interface" for some reason.

Ah, this is because HTML is one of the specs that Shepherd applies heuristics to, and one of those (usually quite reliable and necessary) heuristics is that definitions whose ID starts with `dom-` are IDL-related; if Shepherd can't find anything they should be part of, it's assumed to be an interface.

This can be fixed by manually specifying the type of the definition with a `data-dfn-type=dfn` attribute in HTML.

> but I'm not sure if it is related since it looks like the files haven't been updated super recently:

Yeah the updater fell over just before I went on holiday last year; I'll be figuring out what's up shortly. It's not related to the problem here, but I will need to fix this before any HTML changes get picked up.

index.bs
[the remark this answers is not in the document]
domfarolino: That's very weird, it works for me. Can you try again and also try with make remote? (Also `git pull` since I just added a commit myself)

domfarolino:
> That aside, I think I've applied the requested changes. [=node document=] is now unrecognized too for some reason...

I've fixed this in a small commit.

Thanks Tab! For now I've just specified `<a data-dfn-type=interface>...` so that it'll pick it up properly without the interface font etc. I'll fix this in HTML and when Bikeshed picks the changes up, I'll circle back here and update this.

index.bs
domfarolino: Cool, it seems like we have the "removal" part tracked by another TODO. I'll go ahead and resolve this. I will say it could be interesting to consider what happens if we remove a speculationrules script from one Document and move it to another. Normal scripts that do that don't execute (they have to execute on the document they were prepared on), but I'm not sure if we want to consider doing the same thing here or not. Import maps doesn't seem to care.

index.bs
```diff
@@ -68,12 +71,145 @@ The appropriate times to <a spec=HTML>fetch and process the linked resource</a>
...
+
+<h3 id="speculation-rules-processing">Processing Model</h3>
+
+A [=document=] has a <dfn for=document export>list of speculation rule sets</dfn>, which is an initially empty [=list=].
+
+<!-- TODO(domfarolino): Get rid of the `data-link-type="interface"` once we fix the dfn in HTML. -->
+Periodically, the user agent may [=queue a task=] to the <a data-link-type="interface">DOM manipulation task source</a> to [=consider speculation=] for any [=document=].
+
+<p class="note">
+  The user agent will likely do this after the insertion of new speculation rules, or when resources are idle and available.
```
domfarolino, line 198: I'm starting to wonder what the flow of this should look like. Should we directly invoke "consider speculation" from the steps you added where we mark a `speculationrules` script "ready"? It just feels weird that we mention below that UAs "likely" do this after the insertion of new speculation rules, when in fact we can actually make this happen if we want.

index.bs
```diff
@@ -68,12 +71,145 @@ The appropriate times to <a spec=HTML>fetch and process the linked resource</a>
...
+
+<div algorithm="consider speculation">
+  To <dfn export>consider speculation</dfn> for a [=document=] |document|:
+
+  1. For each |ruleSet| of |document|'s [=document/list of speculation rule sets=]:
+    1. [=list/For each=] |rule| of |ruleSet|'s [=speculation rule set/prefetch rules=]:
+      1. Let |requiresAnonymousClientIP| be true if |rule|'s [=speculation rule/requirements=] [=set/contains=] "`anonymous-client-ip`", and false otherwise.
+      1. [=list/For each=] |url| of |rule|'s [=speculation rule/URLs=]:
+        1. The user agent MAY prefetch |url| given |requiresAnonymousClientIP|.
+           <p class="issue">TODO: expand this to actually elaborate on how prefetch works, once initiated, and to incorporate the |requiresAnonymousClientIP| flag.
```
domfarolino, line 208: We should probably remove the speculation rule set from this list after we process, right? Else it seems like we might re-process the same speculation rule set N times for as many speculation rules scripts as a dev provides.

index.bs
```diff
@@ -68,12 +71,145 @@ The appropriate times to <a spec=HTML>fetch and process the linked resource</a>
...
+
+
+<h3 id="speculation-rules-parsing">Parsing</h3>
+
+<p class="note">
+  The general principle here is to allow the existence of directives which are not understood, but not to accept into the rule set a rule which the user agent does not fully understand.
+  This reduces the risk of unintended activity by user agents which are unaware of most recently added directives which might limit the scope of a rule.
+
+<div algorithm="parse speculation rules">
+  To <dfn export>parse speculation rules</dfn> given a [=string=] |input| and a [=URL=] |baseURL|, perform the following steps. They return a [=speculation rule set=] or null.
```
domfarolino, line 157: These exports aren't necessary, however I notice that they are also sprinkled throughout the import maps spec, so I'll defer to @domenic as for whether or not we should have these.

index.bs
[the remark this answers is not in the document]
jeremyroman: pip upgrading bikeshed fixed it

index.bs
jeremyroman: Quite honestly I've been doing it for consistency with there, but I can certainly strip them. I have no idea whether we should be adding them by some criteria or only when/if some other spec wants to link o them.

index.bs
jeremyroman: Yeah, so my thinking here is that the rules determine what the UA could start, but I don't want to compel UAs to start prefetching immediately -- especially if in a future we allow rules that say things like "any link on this page". Instead I'm trying to express that when it's a good time (page is loaded, CPU is idle, network is fast, the user is hovering over a corresponding link right now, whatever) the UA may choose to do one or more permitted prefetches.

So I want to say that the UA won't do this in the middle of a task (so that you can make multiple changes without being surprised), but otherwise isn't confined to acting immediately. I'm reluctant to spec exactly what those heuristics are.

This sort relates to your other comments. In my view, rules should remain "in effect" and it should be up to the UA to not do silly things like repeatedly fetch the same URL. In fact maybe there are circumstances where it might want to -- if you walk away from your computer but then interact in a way that makes it likely you're going to navigate (like hovering a link), the first prefetch might be stale but a fresh one would be fruitful.

One of the concerns about prerender in particular is that it's hard to hold correctly without taking too much of a performance hit -- this lets the UA take a little more of the burden of deciding when it's fruitful to pre*.

index.bs
domenic: I've mostly added them in cases where otherwise Bikeshed would yell at me because nothing else in the spec references them. (Because, eventually in a non-monkeypatch spec, something _would_ reference them.) Otherwise I would not export them.

domenic:
@jeremyroman can you install and add the PR preview GitHub app? https://github.com/tobie/pr-preview/#install

index.bs
```diff
@@ -68,12 +71,145 @@ The appropriate times to <a spec=HTML>fetch and process the linked resource</a>
...
+    1. If |parsedURL| is failure, then [=iteration/continue=].
+    1. If |parsedURL|'s [=url/scheme=] is not an [=HTTP(S) scheme=], then [=iteration/continue=].
+    1. [=list/Append=] |parsedURL| to |urls|.
+  1. Let |requirements| be an empty [=ordered set=].
+  1. If |input|["`requires`"] [=map/exists=], but is not a [=list=], then return null.
+  1. [=list/For each=] |requirement| of |input|["`requires`"]:
+    1. If |requirement| is not the [=string=] "`anonymous-client-ip`", then return null.
+    1. [=set/Append=] |requirement| to |requirements|.
+  1. Return a [=speculation rule=] with [=speculation rule/URLs=] |urls| and [=speculation rule/requirements=] |requirements|.
+</div>
```
domenic, line 188: Should UAs give warnings when they don't understand a speculation rule? Or would that be too noisy? It might be helpful to mention one way or the other.

index.bs
domenic: Relatedly, should the spec deduplicate multiple prefetches for the same URL? For the same (url, requires anonymous client IP) pair? We could always let UAs do that deduplication via the "MAY", but it seems nicer to nail it down...

index.bs
```diff
@@ -68,12 +71,145 @@ The appropriate times to <a spec=HTML>fetch and process the linked resource</a>
...
+  The user agent will likely do this after the insertion of new speculation rules, or when resources are idle and available.
+
+<div algorithm="consider speculation">
+  To <dfn export>consider speculation</dfn> for a [=document=] |document|:
+
+  1. For each |ruleSet| of |document|'s [=document/list of speculation rule sets=]:
+    1. [=list/For each=] |rule| of |ruleSet|'s [=speculation rule set/prefetch rules=]:
+      1. Let |requiresAnonymousClientIP| be true if |rule|'s [=speculation rule/requirements=] [=set/contains=] "`anonymous-client-ip`", and false otherwise.
+      1. [=list/For each=] |url| of |rule|'s [=speculation rule/URLs=]:
+        1. The user agent MAY prefetch |url| given |requiresAnonymousClientIP|.
```
domenic, line 207:
```suggestion
        1. The user agent may prefetch |url| given |requiresAnonymousClientIP|.
```

https://infra.spec.whatwg.org/#conformance

domenic (commented):
Looking pretty solid...

tabatkins:
> I'll fix this in HTML and when Bikeshed picks the changes up, I'll circle back here and update this.

Cool, I've fixed Bikeshed so it's picking up changes again, so feel free to update this whenever HTML gets fixed.

index.bs
jeremyroman: My suspicion is that surfacing this in a dev tools panel dealing with prefetch, prerender, etc. should surface this information (along with information about the rules that were accepted), but that logging it as a console warning is likely to be too noisy. But we don't have direct experience with this yet, and I could be wrong.

Because I'm not really sure, I leaned away from prescribing what developer tools should do. Do you have a strong opinion?

index.bs
jeremyroman: We certainly could say something like. "The user agent should not prefetch |url| if it has already done so for this document", but it kinda leads into asking how the UA should determine that, which seems like it doesn't make as much sense until the data structures that define prefetching and prefetch matching exist. At that point, maybe we can use "should" language with reference to whatever maps are used for that.

index.bs
domenic: I don't have a strong opinion. For import maps we believed pretty strongly that there were somewhat-sharp edges (like `"foo.mjs"` vs. `"./foo.mjs"`) and we wanted to warn when people ran into them. That is, we thought it was pretty clear that certain errors were developer errors that should be fixed as soon as possible. If you don't think we have that kind of clarity here then omitting seems good for now.

index.bs
domenic: Yeah, OK, deferring this until we have an idea of how prefetch works makes sense. Maybe just update the TODO to note this so that we don't forget.

jeremyroman: Okay, done.

jeremyroman:
> @jeremyroman can you install and add the PR preview GitHub app? https://github.com/tobie/pr-preview/#install

Installed.

index.bs
```diff
@@ -68,12 +71,145 @@ The appropriate times to <a spec=HTML>fetch and process the linked resource</a>
...
+    1. [=list/Append=] |parsedURL| to |urls|.
+  1. Let |requirements| be an empty [=ordered set=].
+  1. If |input|["`requires`"] [=map/exists=], but is not a [=list=], then return null.
+  1. [=list/For each=] |requirement| of |input|["`requires`"]:
+    1. If |requirement| is not the [=string=] "`anonymous-client-ip`", then return null.
+    1. [=set/Append=] |requirement| to |requirements|.
+  1. Return a [=speculation rule=] with [=speculation rule/URLs=] |urls| and [=speculation rule/requirements=] |requirements|.
+</div>
+
+<h3 id="speculation-rules-processing">Processing Model</h3>
```
domenic, line 190:
```suggestion
<h3 id="speculation-rules-processing">Processing model</h3>
```

index.bs
```diff
@@ -68,12 +71,145 @@ The appropriate times to <a spec=HTML>fetch and process the linked resource</a>
...
+<h3 id="speculation-rules-parsing">Parsing</h3>
+
+<p class="note">
+  The general principle here is to allow the existence of directives which are not understood, but not to accept into the rule set a rule which the user agent does not fully understand.
+  This reduces the risk of unintended activity by user agents which are unaware of most recently added directives which might limit the scope of a rule.
+
+<div algorithm="parse speculation rules">
+  To <dfn>parse speculation rules</dfn> given a [=string=] |input| and a [=URL=] |baseURL|, perform the following steps. They return a [=speculation rule set=] or null.
+
+  1. Let |parsed| be the result of [=Parse JSON into Infra values|parsing JSON into Infra values=] given |input|.
```
domenic, line 159:
```suggestion
  1. Let |parsed| be the result of [=parsing a JSON string to an Infra value=] given |input|.
```

index.bs
```diff
@@ -68,12 +71,145 @@ The appropriate times to <a spec=HTML>fetch and process the linked resource</a>
...
+    1. [=set/Append=] |requirement| to |requirements|.
+  1. Return a [=speculation rule=] with [=speculation rule/URLs=] |urls| and [=speculation rule/requirements=] |requirements|.
+</div>
+
+<h3 id="speculation-rules-processing">Processing Model</h3>
+
+A [=document=] has a <dfn for=document export>list of speculation rule sets</dfn>, which is an initially empty [=list=].
+
+<!-- TODO(domfarolino): Get rid of the `data-link-type="interface"` once we fix the dfn in HTML. -->
+Periodically, the user agent may [=queue a task=] to the <a data-link-type="interface">DOM manipulation task source</a> to [=consider speculation=] for any [=document=].
```
domenic, line 195:
```suggestion
Periodically, for any [=document=] |document|, the user agent may [=queue a global task=] on the <a data-link-type="interface">DOM manipulation task source</a> with |document|'s [=relevant global object=] to [=consider speculation=] for |document|.
```

Maybe we should require that _document_ is fully active?

domenic (approved):
LGTM with nits

index.bs
jeremyroman: Added a check within the task (because this could technically change in between, and nothing stops the UA from lifting that check for efficiency). I think prerendered pages (and possibly also bfcache?) still qualify as fully active, so I've left an inline issue to refine that.

index.bs
domfarolino: All of that makes sense, however we've gotten a good share of pushback when it comes to spec'ing things that have magical heuristics before. This might not be as bad since it is less visible on the referrer page etc, and solely an optimization, but we had a lot of push back when it came to automatic lazyload heuristics for example.

domfarolino (approved):
LGTM however I just want to really call out that it is possible we're opening ourselves to a lot of pushback in the future when it comes to the automagical heuristics.

Also circling back to https://github.com/jeremyroman/alternate-loading-modes/pull/24#discussion_r562135293, are there any exports here that fit Domenic's "otherwise" condition, and can therefore be removed @jeremyroman ? This is just an editorial-esque nit so no big deal really...

index.bs
jeremyroman: I removed them.

jeremyroman:
> LGTM however I just want to really call out that it is possible we're opening ourselves to a lot of pushback in the future when it comes to the automagical heuristics.

Yeah, I think it kinda arises from the tension between this being exclusively a platform feature and exclusively a browser feature. At a minimum I don't think we yet know exactly what heuristics would be most useful. It's not impossible it would turn out that relatively simple specable heuristics work in the future and we can tune it, but it seems like leaving it as implementation-defined is appropriate right now.

PR #25, opened 2021-01-22
Add initial fetching restrictions explainer design

domfarolino:
Latest version can be viewed at https://gist.github.com/domfarolino/8f9633db06089b9f4a55e46619a6b5ac

This PR describes the fetching restrictions that will be imposed on what I'm calling "speculative navigation requests". In this PR, this includes:
 - HTTP Cache Partitioning + two alternatives (below)
    - Alternative 1: Using some out-of-band ephemeral cache for requests
    - Alternative 2: Bypassing the HTTP cache altogether
 - Fetching with no credentials
 - Stripping referrer information

### DIFF
--- a/fetch.md
+++ b/fetch.md
@@ -50,7 +50,7 @@ need to utilize this cache key:
  - Subresources under top-level speculative resources
  - Nested browsing contexts and their subresources
 
-### Cache partitioning after activation
+### Cache Partitioning After Activation
 
 At the point of activation, the various speculative HTTP cache partitions may contain resources that are now
 safe to expose to the "typical" cache partitions that non-speculative pages use, since the user has clearly
@@ -67,7 +67,7 @@ that will work spec-wise.
 
 ### Alternatives Considered
 
-#### Using an out-of-band ephemeral cache
+#### Using an Out-of-Band Ephemeral Cache
 
 One alternative to using a speculative HTTP cache partition in the way described above is to use some
 new of out-of-band ephemeral cache to hold these resources, and potentially discard them after activation.
@@ -88,7 +88,7 @@ usage of this brand new cache, particularly because:
  1. Implementation-defined constraints around fetching speculative resources could lead to these resources
     non-deterministically not being fetched, making it difficult to test the cache
 
-##### Ephepermal cache usage activation
+##### Ephepermal Cache After Activation
 
 All requests originating from a [prerendering browsing context][4] after activation would stop targeting the
 ephemeral resource cache, and use their "typical" HTTP cache partition.
@@ -97,7 +97,7 @@ Like the main proposal, we could potentially provide an optional cache merging s
 in the ephemeral cache could be transferred to the HTTP cache for a more-permanent lifetime. This is not without
 its own difficulties and complexities as previously described.
 
-#### Not storing resources in the HTTP cache at all
+#### Bypassing the HTTP cache entirely
 
 Another approach to this problem would be to force-send all requests that were initiated even indirectly
 by a [prerendering browsing context][4] with the [cache mode][3] value of `"no-store"`. Fetching would behave
@@ -124,7 +124,7 @@ another kind of downgrade that we may want to consider, which is the scenario wh
 constraints, the UA opts not to create a [prerendering browsing context][4], but instead downgrade the whole prerender
 process to a simple prefetch. None of this is set in stone though.
 
-## Fetching with no credentials
+## Fetching With No Credentials
 
 Speculative navigation requests should in no way leak the user's identity to third-party origins. A consequence of
 this is that we must fetch all requests originating from a [prerendering browsing context][4] with no credentials.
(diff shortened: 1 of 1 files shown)


domfarolino:
PTAL! I think this is a good start on the fetching modes described in the OP. I do want to call out that I use the term "speculative navigation request" a good bit as opposed to "prerender", because I don't want to give the impression that I'm only talking about `<link rel=prerender>`, when I think much of this applies to portals and probably prefetch as well. I'm open to other suggestions though, as I'm not sure how general we're attempting to make some of the primitives in these explainers.

fetch.md
```diff
@@ -1,11 +1,204 @@
...
+
+Speculative navigation requests should in no way leak the user's identity to third-party origins. A consequence of
+this is that we must fetch all requests originating from a [prerendering browsing context][4] with no credentials.
+As with the HTTP cache partitioning restrictions, this must apply to all of the following:
+ - Top-level speculative resources
+ - Subresources under top-level speculative resources
+ - Nested browsing contexts and their subresources
+
+Concretely this can be done by overriding the [credentials mode][6] of all requests coming from a
+[prerendering browsing context][4] to the `"omit"` value.
```
jeremyroman, line 137: Given the separate network partition, what credentials are there likely to be to omit? Does this mean that multiple requests from the same prerendering browsing context (e.g. for subresources) cannot be correlated? Is that an explicit goal, or incidental?

I'd have expected it to be sufficient to start from an empty partition, so that the prefetching activity can't be correlated with any previous browsing activity.

fetch.md
```diff
@@ -1,11 +1,204 @@
...
+As mentioned, the list of sufficiently-strict referrer policies are the policies that are as or more
+strict compared to the platform [default][7] of `strict-origin-when-cross-origin`, which includes the
+following policies:
+ - `strict-origin-when-cross-origin`
+ - `same-origin`
+ - `strict-origin`
+ - `no-referrer`
+
+A speculative navigation request supplied with a referrer policy outside of this list will be silently
+cancelled.
```
jeremyroman, line 172: Or not issued in the first place? I'd have expected something like "ignored" rather than "cancelled", which to me suggests that we started it in some way. Disregard if this term is used for consistency with existing language.

jeremyroman (approved):
Makes sense to me, though I'm lacking in the details of fetch. :)

fetch.md
```diff
@@ -1,11 +1,204 @@
-# Prerendering fetching modes
+# Prerendering Fetching Modes
 
-Placeholder. TODO.
+## Related Issues and Discussion
```
domenic, line 3: Nit: heading casing is inconsistent. I prefer sentence-case and that's what some of the other documents in the repo do, but it doesn't really matter.

fetch.md
```diff
@@ -1,11 +1,204 @@
...
+process to a simple prefetch. None of this is set in stone though.
+
+## Fetching With No Credentials
+
+Speculative navigation requests should in no way leak the user's identity to third-party origins. A consequence of
+this is that we must fetch all requests originating from a [prerendering browsing context][4] with no credentials.
+As with the HTTP cache partitioning restrictions, this must apply to all of the following:
+ - Top-level speculative resources
+ - Subresources under top-level speculative resources
+ - Nested browsing contexts and their subresources
```
domenic, line 134: IIRC part of the plan is to delay loading nested browsing contexts until activation. I'm unsure whether we'll do that at the fetch level (i.e., just make the fetch hang) or as part of the spec for nested browsing contexts, hmm.

I don't think we need to even mention it here, but it's something for us to keep in mind...

domenic: Well, OK, upon further reflection, we probably want to do it as part of the spec for nested browsing contexts. Then the credentials just work, with no issues.

fetch.md
```diff
@@ -1,11 +1,204 @@
...
+
+> TODO: We can also consider dispatching a console warning or firing the element's error event in these
+cases to let developers know that the request was not able to be dispatched.
+
+Furthermore, since the activation of a speculative navigation request keys on the referrer policy, we
+guarantee that there is never a discrepancy between the referrer sent with the request, and the value
+of `document.referrer` on the prerender page that may eventually get activated and promoted to using
+first-party storage. For an example of this, see [this issue & comment][8].
+
+### Referrer Information After Activation
```
domenic, line 182: I don't think this is talking about after activation; it's talking about referrer information on subresource requests and nested browsing contexts. Right?

domenic (approved):
LGTM with nits

fetch.md
domfarolino: Yeah, I meant to represent the idea that the prerendering processing model will be aborted early, which really means that the request is ignored, not cancelled. I'll re-word this.

fetch.md
domfarolino: Thanks, I went with sentence-case.

fetch.md
domfarolino: I believe cookie jar / credential partitioning is different than network partitioning. This is my understanding from:
 - https://fetch.spec.whatwg.org/#ref-for-concept-request-current-url%E2%91%A1%E2%91%A3, as well as
 - What I've heard about some potential future efforts to work on cookie jar partitioning in the same way we do HTTP cache & connection partitioning, and the fact that this is a separate effort

So basically I think that credentials are not a benefactor of all of the partitioning work we're doing today with the network partition key, but I'd like to sanity check with @domenic to be sure.

fetch.md
domfarolino: OK, as per that and some offline discussion we'll leave this for now since it highlights the important constraint, and we'll figure out exactly how we achieve this constraint later through the spec design.

fetch.md
domfarolino: Hmm, yeah that's right. My intention was to say there is no referrer information change on any requests after activation, because there is no referrer information change on any requests period, after the initial top-level request. I didn't really capture that though, so I'll do that now.

fetch.md
domenic: I agree that they're separate efforts. However, we may also want to discuss a partitioned cookie jar approach? In particular if the main page does `Set-Cookie: ...`, maybe any same-origin `` requests should get that cookie.

domfarolino: Oh nevermind, I did include it in the form of a TODO in the credentials section. In that case, I think this is good to know since the task of elaborating on that alternate proposal is tracked there. I will work on further hashing that out asynchronously.

domfarolino:
I think everything has been addressed, so I'll go ahead and land this now. There are several TODOs in the document that have to do with further-explaining some alternate proposals which I will start working on next.

PR #31, opened 2021-02-02
Fix return and assert in "create a prerendering browsing context"

domfarolino:
We no longer require this algorithm to return anything, so this PR updates the existing spec text to reflect this.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -252,9 +252,9 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps 
 <div algorithm="create a prerendering browsing context">
   To <dfn export>create a prerendering browsing context</dfn> given a [=URL=] |startingURL|, a [=referrer policy=] |referrerPolicy|, and a {{Document}} |referrerDoc|:
 
-  1. Assert: |startingURL|'s [=url/scheme=] is a [=HTTP(S) scheme=].
+  1. If |startingURL|'s [=url/scheme=] is not a [=HTTP(S) scheme=], then return.
 
-  1. If |referrerDoc|'s [=Document/prerendering browsing contexts map=][(|startingURL|, |referrerPolicy|)] [=map/exists=], then return it.
+  1. If |referrerDoc|'s [=Document/prerendering browsing contexts map=][(|startingURL|, |referrerPolicy|)] [=map/exists=], then return.
 
   1. Let |bc| be the result of [=creating a new top-level browsing context=].
 


domfarolino:
This PR formally handles the rejection of non-HTTP(s) schemes in the prerender processing model, given the `Assert` in [create a prerendering browsing context].

However in general, I see three places to handle this, so I'd like to use this PR as an opportunity to as a more general question. Where in the spec should we do these kind of checks? For URLs with non-HTTPs schemes specifically, I see three places:
 1. In the earliest point of the processing model, like this PR
 1. In [process a navigate fetch](https://html.spec.whatwg.org/multipage/browsing-the-web.html#process-a-navigate-fetch)
     1. So we can catch all redirects
 1. In [process a navigate response](https://html.spec.whatwg.org/multipage/browsing-the-web.html#process-a-navigate-response)

It certainly makes sense to do it up-front like this PR does, since we know that non-HTTP(s) schemes are invalid right off the bat, we might as well act on it. Whether or not we want to do it in (2) and/or (3) seems to depend on how we want to handle redirects. The spec currently only does it in (3) (as per [this section](https://jeremyroman.github.io/alternate-loading-modes/index.html#no-bad-navs)), which is a little odd because as per my reading of HTML, this means we support the following: An HTTP(s) prerender that redirects to `ftp://`, and then later redirects to an HTTP(s) scheme.

If my reading is correct, we probably don't want to support that even accidentally, so I feel like our [check in (3)](https://jeremyroman.github.io/alternate-loading-modes/index.html#no-bad-navs) should be moved to (2), so we can assert this on ever redirect encountered.

Regardless, I think we want logic in (2) anways to handle redirects properly in general. For example, we probably want to:
 - Check that absolutely _whenever_ a prerender is redirected cross-origin, we set its `loading mode` to `uncredentialed-prerender`
 - ... as well as if it has gone cross-origin and a redirect has set a referrer policy that is not sufficiently strict, we cancel the request immediately

Given this, if others agree, I propose we land this PR, and then I follow up by introducing prerender-specific logic in the navigation redirect handling portion of the spec to cater to the above things as well as others we see in the future. WDYT?

domenic (approved):
Content LGTM, now to respond to your discussion topic...

domenic:
I don't think non-HTTP(S) URLs can redirect? In particular server-side redirecting is done via _HTTP_ status codes, which are not present for FTP or `file:` or similar.

Client-side redirecting to bad URLs is intended to be prevented via the check in (3), where we never actually replace the current document if the place we've ended up at is non-HTTP(S).

So I think the current architecture is fine for the HTTP(S)-ness. Adding this PRs saves us the work of doing a fetch, which is good.

Assuming that, for other checks besides HTTP(S)-ness, we want to be strict on redirects in the manner you describe, then I agree we need some checks in (2). And so yeah, moving some or all the checks from (3) to (2) makes sense.

domfarolino:
Great, thanks for the reply! I'll go ahead and land this.

> I don't think non-HTTP(S) URLs can redirect? In particular server-side redirecting is done via HTTP status codes, which are not present for FTP or file: or similar.

That's a good point, I didn't really think of that.

PR #33, opened 2021-02-02
Specify referrer policy redaction

domfarolino:
This PR specifies the referrer policy redaction proposed in https://github.com/jeremyroman/alternate-loading-modes/blob/main/fetch.md#stripping-referrer-information.

Currently we do the same-origin check and ignore-prerender-on-loose-referrer-policy work in the prerender processing model itself. I'm happy with this, but I'd also be happy doing it later on in process a navigate fetch, it's kind of a similar question to https://github.com/jeremyroman/alternate-loading-modes/pull/31#issuecomment-771538203 that asks where we should perform all of these checks.

Regardless, a later PR will patch the [process a navigate fetch](https://html.spec.whatwg.org/multipage/browsing-the-web.html#process-a-navigate-fetch) algorithm to handle redirects, where we'll need to perform another same-origin check and potentially set the loading mode to "uncredentialed-prerender" or cancel the prerender altogether if a too-loose referrer policy was set on a redirect.


index.bs
```diff
@@ -66,16 +66,20 @@ The appropriate times to <a spec=HTML>fetch and process the linked resource</a>
...
+
+  1. [=Create a prerendering browsing context=] with |url|, |referrerPolicy|, and |el|'s [=Node/node document=].
 </div>
 
 A user agent must not <a spec=HTML>delay the load event</a> of the <{link}> element's [=Node/node document=] for this link type.
 
 <p class="note">Note that this link type does not fire {{HTMLElement/error}} or {{HTMLElement/load}} events, unlike many other link types that create <a spec=HTML>external resource links</a>.
 
-<p class="issue">TODO: We should define a list of allowable referrer policies to prerender with, and early-return if the above referrer policy is not in this list.
+The <dfn>list of sufficiently-strict referrer policies</dfn> is a [=list=] containing the following subset of [=referrer policies=]: the empty string, "`strict-origin-when-cross-origin`", "`strict-origin`", "`same-origin`", "`no-referrer`".
```
domenic, line 82: You might be able to cross-link these, although getting them to be in code font is harder... I'm pretty sure `<a><code>strict-origin-when-cross-origin</code></a>` will work but I'm not sure about using `[= =]` or backtick shorthand.

domenic, line 82: Given the eventual plan to move this into HTML, we might want to give this a more specific name, e.g. "list of prerendering-allowed referrer policies"? Although, if we're planning on reusing this for prefetch then maybe that's not so good... we could leave it for now and figure it out when we merge into HTML.

index.bs
```diff
@@ -66,16 +66,20 @@ The appropriate times to <a spec=HTML>fetch and process the linked resource</a>
 
   1. [=Parse a URL=] given |el|'s <{link/href}> attribute, relative to |el|'s [=Node/node document=]. If that fails, then return. Otherwise, let |url| be the [=resulting URL record=].
 
-  1. Let |referrer policy| be  the current state of |el|'s <{link/referrerpolicy}> attribute.
+  1. Let |referrerPolicy| be  the current state of |el|'s <{link/referrerpolicy}> attribute.
 
-  1. [=Create a prerendering browsing context=] with |url|, |referrer policy|, and |el|'s [=Node/node document=].
+  1. If |url| is not [=same origin=] with [=Node/node document=]'s [=Document/url=], then:
```
domenic, line 71:
```suggestion
  1. If |url|'s [=url/origin=] is not [=same origin=] with |el|'s [=Node/node document=]'s [=Document/origin=], then:
```

index.bs
domfarolino: The latter is what I was thinking. We'll probably have to move this to a more general location anyways once we better-specify `<link rel=prefetch>`.

domenic (approved):

index.bs
domfarolino: Hmm, actually I've gone ahead and renamed this the `list of sufficiently-strict speculative navigation referrer policies`, which is a mouthful, but I think it gives us good "coverage" for future processing models that use this infra later.

domfarolino: Hmm, actually I've gone ahead and renamed this the `list of sufficiently-strict speculative navigation referrer policies`, which is a mouthful, but I think it gives us good "coverage" for future processing models that use this infra later.

PR #35, opened 2021-02-03
Allow for the activation of a prerendering browsing context upon navigation redirect

### LINKED ISSUE #17: Allow redirects to activate a prerender
As per the current spec's [Allowing activation in place of navigation](https://htmlpreview.github.io/?https://raw.githubusercontent.com/jeremyroman/alternate-loading-modes/spec/index.html#navigate-activation) section, we modify the HTML Standard's navigation algorithm to activate prerenders early on, however if the navigation later redirects, we should probably re-check the activation conditions? If that's the case I'm happy to update the spec accordingly.

We've also discussed more generally what effect redirects have on prerendering, and the current idea is that the URL used for activation of a prerendering browsing context will be the "initial" prerendering URL that was used to instantiate the prerendering browsing context, as per https://docs.google.com/document/d/1E82ZqXibjCvpPdOgMmOfPIatPMEBVAiJ2KQm6r1_E94/edit#heading=h.soo6uwz4i3m. This is orthogonal to what this issue tracks, which is allowing us to obtain the URL-to-activate via a navigation redirect. Also it appears that the spec already supports the activation URL being the prerendering browsing context's _`startingURL`_, so I think we're all set on that front.

domfarolino:
This PR:
 - Consolidates the conditions that are checked before activating a prerender to `<dfn>can activate a prerender</dfn>` which returns a true/false result as to whether or not we can activate a prerendering browsing context
 - Invokes these steps in two places: (1) In normal #navigate, like we do now, (2) In #process-a-navigate-fetch for redirect handling of non-prerendering navigations
    - I wrote the spec in such a way where only navigation redirects in non-prerendering-browsing-contexts can activate prerendering browsing contexts. This raises the more general question: should we let prerenders activate other prerenders /cc @jeremyroman. We should probably answer this question before landing this PR (see the review I left below)

Closes #17. /cc @mfalken

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/pull/35.html" title="Last updated on Feb 5, 2021, 5:46 PM UTC (d603169)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/35/fa427e0...d603169.html" title="Last updated on Feb 5, 2021, 5:46 PM UTC (d603169)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -343,19 +343,16 @@ The <dfn attribute for="Document">onprerenderingchange</dfn> attribute is an [=e
 <h3 id="navigate-activation">Allowing activation in place of navigation</h3>
 
 <div algorithm="can activate a prerender">
-  The <dfn>can activate a prerender</dfn> algorithm takes a [=browsing context=] |browsingContext|, a [=history handling behavior=] |historyHandling|, a string |navigationType|, and a [=request=] |request|, and runs the following steps:
+  We <dfn>can activate a prerender</dfn> given a [=browsing context=] |browsingContext|, a [=history handling behavior=] |historyHandling|, a string |navigationType|, and a [=request=] |request|, if the following steps return true:
 
-  1. If all of the following are true:
+  1. Return true if all of the following are true:
 
       * |browsingContext| is a [=top-level browsing context=]
+      * |browsingContext| is not a [=prerendering browsing context=]
       * |historyHandling| is "`default`" or "`replace`"
       * |navigationType| is "`other`"
       * |request|'s [=request/method=] is \``GET`\`
       * |browsingContext|'s [=active document=]'s [=Document/prerendering browsing contexts map=][(|request|'s [=request/URL=], |request|'s [=request/referrer policy=])] [=map/exists=] and is not [=prerendering browsing context/empty=]
-
-    then:
-
-    1. Return true.
 
   1. Otherwise, return false.
 
@@ -366,7 +363,7 @@ Patch the [=navigate=] algorithm to allow the [=prerendering browsing context/ac
 <div algorithm="navigate activate patch">
   In [=navigate=], append the following steps after the fragment navigation handling (currently step 6):
 
-  1. If |resource| is a [=request=] and [=can activate a prerender=] given |browsingContext|, |historyHandling|, <var ignore>navigationType</var>, and |resource| returns true, then:
+  1. If |resource| is a [=request=] and we [=can activate a prerender=] given |browsingContext|, |historyHandling|, <var ignore>navigationType</var>, and |resource|, then:
 
     1. [=prerendering browsing context/Activate=] <var ignore>successorBC</var> in place of |browsingContext| given |historyHandling|.
 
@@ -386,7 +383,7 @@ Patch the [=process a navigate fetch=] algorithm like so:
 <div algorithm="navigate activate patch redirect handling">
   In [=process a navigate fetch=], append the following steps after the first sub-step under "While true:" in order to handle redirects correctly:
 
-  1. If |browsingContext| is not a [=prerendering browsing context=] and [=can activate a prerender=] given |browsingContext| |historyHandling|, <var ignore>navigationType</var>, and <var ignore>request</var> returns true, then:
+  1. If we [=can activate a prerender=] given |browsingContext|, |historyHandling|, <var ignore>navigationType</var>, and <var ignore>request</var>, then:
 
     1. [=prerendering browsing context/Activate=] |browsingContext| in place of <var ignore>sourceBrowsingContext</var> given |historyHandling|.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -342,36 +342,55 @@ The <dfn attribute for="Document">onprerenderingchange</dfn> attribute is an [=e
...
+ * Logic allowing navigation redirects that are not inside of a [=prerendering browsing context=] to [=prerendering browsing context/activate=] a [=prerendering browsing context=].
+ * Logic that updates specific state of a [=prerendering browsing context=] when it redirects coss-origin.
+
+Patch the [=process a navigate fetch=] algorithm like so:
 
 <div algorithm="navigate activate patch redirect handling">
-  In [=process a navigate fetch=], append the following steps after the first sub-step under "While true:" in order to handle redirects correctly;
+  In [=process a navigate fetch=], append the following steps after the first sub-step under "While true:" in order to handle redirects correctly:
+
+  1. If |browsingContext| is not a [=prerendering browsing context=] and [=can activate a prerender=] given |browsingContext| |historyHandling|, <var ignore>navigationType</var>, and <var ignore>request</var> returns true, then:
```
domfarolino, line 389: We should either modify this PR to do one of the following, for completeness:
 - Restrict prerendering activation in the non-redirect nav case to non-prerendering browsing contexts (so that prerender requests cannot activate other prerender browsing contexts)
 - Lift this "is not a prerendering browsing context" condition, so that all navigation requests (including prerender browsing context navs) can activate activation prerendering browsing contexts

index.bs
```diff
@@ -342,36 +342,55 @@ The <dfn attribute for="Document">onprerenderingchange</dfn> attribute is an [=e
...
 <h3 id="navigate-activation">Allowing activation in place of navigation</h3>
 
-Patch the [=navigate=] algorithm to allow the [=prerendering browsing context/activate|activation=] of a [=prerendering browsing context=] in place of a normal navigation as follows:
-
-<div algorithm="navigate activate patch">
-  In [=navigate=], append the following steps after the fragment navigation handling (currently step 6):
+<div algorithm="can activate a prerender">
+  The <dfn>can activate a prerender</dfn> algorithm takes a [=browsing context=] |browsingContext|, a [=history handling behavior=] |historyHandling|, a string |navigationType|, and a [=request=] |request|, and runs the following steps:
 
   1. If all of the following are true:
```
domenic, line 348: "Return true if all of the following are true:", and remove the nested step? Seems slightly cleaner.

index.bs
```diff
@@ -342,36 +342,55 @@ The <dfn attribute for="Document">onprerenderingchange</dfn> attribute is an [=e
 
 <h3 id="navigate-activation">Allowing activation in place of navigation</h3>
 
-Patch the [=navigate=] algorithm to allow the [=prerendering browsing context/activate|activation=] of a [=prerendering browsing context=] in place of a normal navigation as follows:
-
-<div algorithm="navigate activate patch">
-  In [=navigate=], append the following steps after the fragment navigation handling (currently step 6):
+<div algorithm="can activate a prerender">
+  The <dfn>can activate a prerender</dfn> algorithm takes a [=browsing context=] |browsingContext|, a [=history handling behavior=] |historyHandling|, a string |navigationType|, and a [=request=] |request|, and runs the following steps:
```
domenic, line 346: An optional trick to make the English flow a bit better: "We **can activate a prerender**, given a browsing context..., if the following steps return true:"

Then later instead of "If can activate a prerender given ... returns true" you can say "If we can activate a prerender given ..."

index.bs
domenic, line 389:
```suggestion
  1. If |browsingContext| is not a [=prerendering browsing context=] and [=can activate a prerender=] given |browsingContext|, |historyHandling|, <var ignore>navigationType</var>, and <var ignore>request</var> returns true, then:
```

domenic: > Restrict prerendering activation in the non-redirect nav case to non-prerendering browsing contexts (so that prerender requests cannot activate other prerender browsing contexts)

Recursive prerendering seems complicated and scary to me, but I can't think of any concrete problems with it. This might be a thing where we disallow it for now, with a spec note that we could loosen it in the future? I guess it largely comes down to our implementation appetite.

domfarolino: That sounds good to me. I'm cool with preventing it for now, especially since it isn't really the intention of prerendering, and could potentially be a footgun anyways. OK I'll go for that then.

index.bs
domfarolino: Agreed, wasn't sure if the "Return true" if was "conventional" or weird to people. Done.

index.bs
domfarolino: Done locally (since I made other changes before I saw this, and it was simpler that way).

domenic (approved):

PR #36, opened 2021-02-05
Spec some initial permissions check delays

domenic:
Namely geolocation (special), serial (simple except for dedicated workers, left as a TODO), and notifications (simple except for service workers, which is left as a TODO).

I'd like to fix the dedicated workers TODO before merging this. (I ran out of time this Friday evening.) I think dedicated workers might grow their own list of post-prerendering steps...

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/pull/36.html" title="Last updated on Feb 12, 2021, 7:42 PM UTC (5ef770c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/36/3d42e1e...5ef770c.html" title="Last updated on Feb 12, 2021, 7:42 PM UTC (5ef770c)">Diff</a>


index.bs
```diff
@@ -499,6 +523,64 @@ Modify the <a spec=HTML>downloads a hyperlink</a> algorithm to ensure that downl
...
+</div>
+
+<div algorithm="Geolocation watchPosition patch">
+  Modify the {{Geolocation/watchPosition()}} method steps by prepending the following step:
+
+  1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then:
+
+    1. Let |watchProcessId| be a new unique <a spec=GEOLOCATION>watch process</a> ID, and note it as a <dfn>post-prerendering activation geolocation watch process ID</dfn>.
+
+    1. Append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=], with the modification that the <a spec=GEOLOCATION>watch process</a> generated must use |watchProcessId| as its ID, and return |watchProcessId|.
```
jeremyroman, line 545: This is slightly confusing in the subtlety of the change it applies to another step. Maybe that's fine for now because we want to keep the delta small, but I assume if this were to be merged we would refactor this?

index.bs
```diff
@@ -499,6 +523,64 @@ Modify the <a spec=HTML>downloads a hyperlink</a> algorithm to ensure that downl
...
+<h3 id="permissions">Waiting before permissions checks</h3>
+
+Many specifications need to be patched so that, if a given algorithm invoked in a [=prerendering browsing context=], most of its work is deferred until the browsing context is [=prerendering browsing context/activated=]. This is tricky to do uniformly, as many of these specifications do not have great hygeine around <a href="https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-for-spec-authors">using the event loop</a>. Nevertheless, the following sections give our best attempt.
+
+<h4 id="patch-geolocation">Geolocation API</h4>
+
+<div algorithm="Geolocation getCurrentPosition patch">
+  Modify the {{Geolocation/getCurrentPosition()}} method steps by prepending the following step:
+
+  1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=] and return.
```
jeremyroman, line 535: I wonder if we'd want to simply define a shorthand like "continue after prerendering" and use it similarly to how upstream geolocation API says "continue asynchronously". Is there some better spec analogue for async/await?

index.bs
jeremyroman: If we had infinite time, it might even be worth having the upstream thing have an explicit map of watch IDs to some structure that includes an abort signal (even if it's not web-exposed). But that's probably too much of a detour to ask for right now. :)

domenic: Yeah, the geolocation spec is so old-school, I'm not sure that even if we did incorporate it, I'd refactor as you suggest. If you look at how watch processes are currently associated and IDs are generated, it's so hand-wavy, that making this one thing precise seems like it might be more confusing than helpful.

I am tempted to refactor the geolocation spec entirely, but I feel like I'd quickly run into interop issues and then get stuck waiting for implementers to come agree on something.

index.bs
domenic: Yeah, I'm thinking that might be a good idea. I want to get 3-4 more long-handed samples done before doing such a refactoring, but ultimately this should be as easy as possible to use from other specs, and providing a macro of that sort seems like the right method.

mfalken:
I took a quick look at what the spec is and how it might be followed by our implementation. There's a disconnect between where the implementation will delay things and when the spec does. Due to our method of delaying at the level of Mojo/IPC,  the implementation will delay only at async steps (when the page asks the browser process to do work), while the spec delays at the beginning of the algorithm. Therefore our implementation is prone to do things like call the error callback with "insecure context" since it can do that without an IPC, while the spec is saying not to do any of that until activation.

This may be a common problem when writing the spec for these features.

I think @nhiroki is taking a deeper look, and may add more comments.

nhiroki:
I'm now thinking about how to implement the restriction for geolocation in Chromium. I haven't seen any issues other than @mfalken mentioned so far. When I find something, I'll share it here.

mfalken:
BTW my understanding is this is proposed spec text and it may go another direction depending on implementor feedback.  I do think it'd be great if we can just use the Mojo/IPC based delay mechanism as-is and not have to make many feature-specific changes in the implementation, but that may come at the cost of spec complexity. And I'm not sure if geolocation is or is not the only one that would need such changes if we go with the current spec text.

domenic:
Although we certainly want to be flexible to implementation feedback, I'd push back against basing the spec on implementation-specific mechanisms like Chromium's IPC framework. We need to achieve something that's interoperably implementable across all browsers, and the best way to do that is by picking a specific point in the algorithm to delay.

I'm not sure we should pick that point based on where Chromium starts doing IPC. That does not seem fair to other implementations or like it will lead to a good uniform delay mechanism across all the APIs. Thus my proposal of just starting the delay as soon as the algorithm starts.

As for other APIs, I suspect many of them will have similar checks which are possible to do in-process in Chrome but might be out-of-process in other browsers, or vice-versa. So I don't think this problem is limited to geolocation.

jeremyroman:
+1 domenic; there's no guarantee that other implementations would IPC in the same places or would use IPC at all. It might be acceptable in the short term for Chromium to behave slightly differently while we iron this out, and for IPC filtering to be a backstop against bugs, but I don't think it makes sense to authors or other implementers to have behavior too closely tied to where Chromium requests a Mojo interface from the browser process.

mfalken:
Thanks, that largely makes sense. I was imagining something like the spec being flexible in allowing the browser to process certain steps during prerendering, while ensuring that the actual concerning step of doing the geolocation is deferred, but I don't have a concrete idea of how to do that, if it's even possible (and as @domenic says, picking a specific point is more predictable). I agree we wouldn't want to require browsers to only defer where Chromium's IPC does (which may also change over time).

This does relate to the overall implementation maintenance cost of prerendering, i.e., the relevant features will need IsPrerendering() checks, which I thought we might be able to side-step with the IPC solution. But I agree it's tractable especially if the spec text is written for all the features.

domenic:
Alright, this is ready for re-review. I did a few more cases, although quite a number of them are covered by user activation.

As the comment threads above mention, at some point we'll probably want to create a spec macro for the very common pattern of delaying everything after promise creation. But for now, I'll keep going with the repetition, and refactor one we get a few more exceptional cases, so I can see whether such a macro should cover them or not.

I'm going to start a tracking issue that ports the appendix of https://docs.google.com/document/d/1zY15k_wFTik2EoxBf3_RT7YjYpFMDaeNspy15n0rtww/edit# onto GitHub and we can use to track the progress on specifying restrictions.

index.bs
```diff
@@ -499,6 +548,97 @@ Modify the <a spec=HTML>downloads a hyperlink</a> algorithm to ensure that downl
...
+  Modify the {{Serial/requestPort()}} method steps by inserting the following steps after the initial creation of |promise|:
+
+  1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=] and return |promise|.
+</div>
+
+<div algorithm="Serial getPorts patch">
+  Modify the {{Serial/getPorts()}} method steps by inserting the following steps after the initial creation of |promise|:
+
+  <!-- TODO link owner document once https://github.com/whatwg/html/pull/6379 lands -->
+  1. Let |document| be [=this=]'s [=relevant global object=]'s [=associated Document=], if [=this=]'s [=relevant global object=] is a {{Window}}, or [=this=]'s [=relevant global object=]'s owner document, if [=this=]'s [=relevant global object=] is a {{DedicatedWorkerGlobalScope}}.
```
jeremyroman, line 595: Is there any reason not to make the relevant the associated Document be the owner document for a window global? It feels like this complication need not live at the use of this term.

jeremyroman (approved):

index.bs
domenic: Hmm. I guess I think they're sufficiently different things that they shouldn't use the same term. A Window <-> Document relationship is rather special, but the Window -> DedicatedWorkerGlobalScope ownership feels like a different sort of thing. Also, since it would only work for Window and (some) DedicatedWorkerGlobalScopes, not other globals, the generalization doesn't buy us that much.

PR #37, opened 2021-02-10
document.prerendering should be based on top-level browsing context

bokand:
`document.prerendering` should accurately reflect whether a nested document is in a prerendering context or not. The current definition of "prerendering browsing context" is limited only to top-level browsing contexts so when called from a nested document the property should return the value for the top-level BC.

***
<a href="https://pr-preview.s3.amazonaws.com/bokand/alternate-loading-modes/pull/37.html" title="Last updated on Feb 24, 2021, 6:14 PM UTC (36d8fb3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/37/789b4c0...bokand:36d8fb3.html" title="Last updated on Feb 24, 2021, 6:14 PM UTC (36d8fb3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -309,7 +309,9 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps 
 
       1. Set |successorBC|'s [=browsing context/loading mode=] to "`default`".
 
-      1. [=Fire an event=] named {{Document/prerenderingchange}} at |successorBC|'s [=active document=].
+      1. Let |d| be |successorBC|'s [=active document=].
+
+      1. [=Fire an event=] named {{Document/prerenderingchange}} at |d| and at each [=active document=] of each of |d|'s [=list of the descendant browsing contexts=].
 
       1. [=list/For each=] |steps| in |successorBC|'s [=active document=]'s [=Document/post-prerendering activation steps list=]:
 
@@ -331,7 +333,7 @@ Amend the {{Document}} interface:
   };
 </pre>
 
-The <dfn attribute for="Document">prerendering</dfn> getter steps are to return true if [=this=] has a non-null [=Document/browsing context=] that is a [=prerendering browsing context=]; otherwise, false.
+The <dfn attribute for="Document">prerendering</dfn> getter steps are to return true if [=this=] has a non-null [=Document/browsing context=] whose [=top-level browsing context=] is a [=prerendering browsing context=]; otherwise, false.
 
 <p class="note">This attribute lets pages know when they're being presented in a non-interactive "prerendering-like" context. In the future, this would include a visible document in a `<portal>` element, both when loaded into it or via predecessor adoption.
 
--- a/index.bs
+++ b/index.bs
@@ -44,6 +44,38 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: script; url: concept-script
 </pre>
 
+<style>
+/* domintro from https://resources.whatwg.org/standard.css */
+.domintro {
+  position: relative;
+  color: green;
+  background: #DDFFDD;
+  margin: 2.5em 0 2em 0;
+  padding: 1.5em 1em 0.5em 2em;
+}
+
+.domintro dt, .domintro dt * {
+  color: black;
+  font-size: inherit;
+}
+.domintro dd {
+  margin: 0.5em 0 1em 2em; padding: 0;
+}
+.domintro dd p {
+  margin: 0.5em 0;
+}
+.domintro::before {
+  content: 'For web developers (non-normative)';
+  background: green;
+  color: white;
+  padding: 0.15em 0.25em;
+  font-style: normal;
+  position: absolute;
+  top: -0.8em;
+  left: -0.8em;
+}
+</style>
+
 <h2 id="link-rel-prerender">Link type "<dfn attr-value for="link/rel"><code>prerender</code></dfn>"</h2>
 
 <em>The following section would be added as a new sub-section of [[HTML]]'s <a href="https://html.spec.whatwg.org/multipage/links.html#linkTypes">Link types</a> section. This should replace the prerender documentation in [[RESOURCE-HINTS]]</em>
@@ -230,7 +262,26 @@ Periodically, for any [=document=] |document|, the user agent may [=queue a glob
 <p class="issue">
   We should also notice removals and consider cancelling speculated actions.
 
-<h2 id="prerendering-bcs">Prerendering browsing contexts</h2>
+<h2 id="prerendering-bcs">Prerendering browsing context infrastructure</h2>
+
+<h3 id="document-prerendering">Extensions to the {{Document}} interface</h3>
+
+We'd modify [[HTML]]'s centralized definition of {{Document}} as follows:
+
+<pre class="idl">
+  partial interface Document {
+      readonly attribute boolean prerendering;
+
+      // Under "special event handler IDL attributes that only apply to Document objects"
+      attribute EventHandler onprerenderingchange;
+  };
+</pre>
+
+The <dfn attribute for="Document">onprerenderingchange</dfn> attribute is an [=event handler IDL attribute=] corresponding to the <dfn event for="Document">prerenderingchange</dfn> [=event handler event type=]. (We would update the corresponding table in [[HTML]], which currently only contains {{Document/onreadystatechange}}.)
+
+As is customary for [[HTML]], the definition of {{Document/prerendering}} would be located in another section of the spec; we'd place it in the new section introduced below:
+
+<h3 id="prerendering-bcs-subsection">Prerendering browsing contexts</h3>
 
 <em>The following section would be added as a new sub-section of [[HTML]]'s <a href="https://html.spec.whatwg.org/multipage/browsers.html#windows">Browsing contexts</a> section.</em>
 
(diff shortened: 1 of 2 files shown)


bokand:
> LINK ERROR: No 'dfn' refs found for 'list of the descendant browsing contexts' that are marked for export.

I wasn't able to get autolinking to work for the [list of the descendant browsing contexts](https://html.spec.whatwg.org/multipage/browsers.html#list-of-the-descendant-browsing-contexts) definition. Does it need to be exported or is there a better way to do something for each nested browsing context? @domenic would probably know best.

index.bs
```diff
@@ -309,7 +309,9 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
 
       1. Set |successorBC|'s [=browsing context/loading mode=] to "`default`".
 
-      1. [=Fire an event=] named {{Document/prerenderingchange}} at |successorBC|'s [=active document=].
+      1. Let |d| be |successorBC|'s [=active document=].
+
+      1. [=Fire an event=] named {{Document/prerenderingchange}} at |d| and at each [=active document=] of each of |d|'s [=list of the descendant browsing contexts=].
```
domenic, line 314: It's not quite obvious from this phrasing that it's fired in order. I'd prefer "For each descendantBC of d's list of descendant browsing contexts, fire an event named prerenderingchange at descendantBC's active document."

Also, should we queue a separate task for each of these? That seems potentially necessary since the other BCs could be cross-process?

domenic (commented):
"list of descendant BCs" is indeed not exported. Since we're a monkeypatch spec, we use a lot of non-exported terms; the way I've been getting around that is to use `<a spec=HTML>list of descendant browsing contexts</a>` which punches through export/noexport distinctions.

index.bs
bokand: Done. And yes, I think separate tasks makes sense, though I only have a cursory knowledge about the details there so I just cargo cult-ed from above. Does this look ok?

domfarolino: Hmm, will any descendant browsing contexts be cross-process? My understanding is that we'll only load same-origin iframes whilst prerendering, in which case I think they are all in the same agent cluster and BCG? Does that impact any of this?

jeremyroman: I don't see why we couldn't load cross-origin subframes if they were themselves storageless or opted in. I don't think we should bake in an assumption that all nested BCs share an event loop, so queuing tasks makes sense to me.

bokand: Maybe not (I'm not caught up on that discussion), but it seems like something that could change? This still works in the case where we don't load cross-process iframes so it seems like a better approach to me. The only drawback I can see is that subframe's may still show prerendering state while we flush the `post-prerendering activation steps list` of the current document. That seems probably ok to me since this list is associated with specific document.

jeremyroman: That said this current description doesn't make it clear to me if these other contexts see `document.prerendering` change at the same time as they receive the event, which they should. This may require us to explicitly plumb a flag through to them. We probably should do this anyway, as synchronously asking for a property of a browsing context with another event loop (as we do here) is kinda a dodge.

domenic: That's a good point. However, I'm generally a bit scared about posting tasks into other globals from one global. So I slightly prefer individual tasks.

@bokand, how do you plan to implement this? Maybe that can be the tiebreaker.

domenic (approved):
Looks great, although now @domfarolino has me questioning my initial instinct on separate tasks, so your confirmation as implementer would be great.

index.bs
domenic: Oh, I see people were talking while I was reviewing. Yeah, I agree with @jeremyroman's concern that the event and `document.prerendering` updating will not necessarily happen at the same time with this spec text.

domfarolino: Yeah so just so that I'm clear, I think the concern here is that documents in the same agent that indeed share the same event loop and thus can synchronously access each other, have their `document.prerendering` updated at different times. Am I right in thinking that is the concern?

Of course documents across agent clusters will necessarily have their `document.prerendering` state updated at different times.

domenic: Hmm, that might be another concern. But Jeremy's concern is that as defined, the `prerendering` getter updates instantly (since the TLBC switches from prerendering to not-prerendering at some point in time). Whereas the event fires from a queued task.

I guess that's not so bad, compared to the reverse...

domfarolino: Ah, I see. One consequence of that is since we're planning on using the state of the TLBC when considering fetching restrictions for a given request, then it is possible that a web page could observe that a request was sent with no fetching restrictions _just shortly_ before the `prerendering` getting changes its value, but I agree that doesn't seem like a big deal.

My concern is that multiple documents sharing the same agent might be able to observe an inconsistency in their `prerendering` getter values, which really isn't a big deal, but I just want to make sure that it is known.

bokand: Thanks all.

Agree `document.prerendering` shouldn't change until the event is dispatched so I'll go ahead and fix that.

Yeah, this does mean that the `prerendering` value may be inconsistent between documents but I don't think that should be an issue. Especially since, within an agent, the ordering will be consistent (parent frames will always make and see the change before children). The expected use cases should really only be checking their own `document`'s value anyway.

Just to clarify:

> That's a good point. However, I'm generally a bit scared about posting tasks into other globals from one global. So I slightly prefer individual tasks.

I wasn't sure who this was in reply to, what do you mean by "posting tasks into other globals from one global" vs "individual tasks"? Is the current PR "posting into other globals" (since each posted task uses a different `window`?)

> @bokand, how do you plan to implement this? Maybe that can be the tiebreaker.

I hadn't thought that far ahead :). The prerequisite CL plumbing this state into the renderer just landed though so I'll go ahead and try this out before making further progress here.

domenic: > I wasn't sure who this was in reply to, what do you mean by "posting tasks into other globals from one global" vs "individual tasks"? Is the current PR "posting into other globals" (since each posted task uses a different window?)

I phrased this poorly. The current PR is "individual tasks".

The alternative I'm contrasting it with is posting a single task, from the parent global, which updates the prerendering state of all the same-agent child globals. There's nothing really wrong with that, but it's a bit scary to me.

index.bs
```diff
@@ -309,7 +309,13 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
 
       1. Set |successorBC|'s [=browsing context/loading mode=] to "`default`".
 
-      1. [=Fire an event=] named {{Document/prerenderingchange}} at |successorBC|'s [=active document=].
+      1. Let |d| be |successorBC|'s [=active document=].
+
+      1. [=Fire an event=] named {{Document/prerenderingchange}} at |d|.
+
+      1. [=list/For each=] |descendantBC| of |d|'s <a spec="HTML">list of the descendant browsing contexts</a>, [=queue a global task=] on the [=networking task source=], given |descendantBC|'s [=browsing context/active window=], with steps:
+
```
nhiroki, line 317: When do we run descendantBC's active document's post-prerendering activation steps list?

bokand: We currently do not. Per the spec, we only add to these steps for a prerendering browsing context which excludes subframes. It's a good point that's probably not the right behavior. IIUC, we currently use this to defer a download until activation - but I don't think this covers the case where a subframe in the prerender initiates a download.

domenic:
@bokan @nhiroki please take a look at my latest revision. It makes every browsing context have a loading mode, and explicitly propagates the change as part of activation.

The biggest question is probably around timing and ordering. I tried to follow what I understood of @nhiroki's [design doc](https://docs.google.com/document/d/1_l1LDUALf8PbZIz5y_UYNGZmYCMKNpePDLBHn-gTsTM/edit#) but I'm not sure I got it right.

bokand:
Thanks @domenic! This looks great to me and matches my understanding of @nhiroki's doc as well.

This does mean that:

  * The main frame will always activate before any nested frames, but...
  * Among nested frames, they may activate in any order (i.e. a parent may activate either before or after its child)

The asymmetry feels a little odd to me but I also don't see a real problem with it and enforcing an ordering (e.g. parent activates before child) could lead to performance bottlenecks. I wonder if we could just include the main frame when iterating and posting the activation tasks to frames so that there are no dependencies between activation at all? @nhiroki is there a technical reason to prefer the main frame or were you just going by my original text in this PR?

nhiroki:
Thank you! LGTM. This is aligned with my understanding.

I initially thought this PR doesn't ensure the order from parent subframe to child subframes, but as [replied in the doc](https://docs.google.com/document/d/1_l1LDUALf8PbZIz5y_UYNGZmYCMKNpePDLBHn-gTsTM/edit?disco=AAAALi9OLUY), this seems wrong.

In the activate algorithm, the `list of the descendant browsing contexts` lists frames from parent to children, and the `queue a global task on the networking task source` keeps the order to dispatch the event because these frames are in the same agent except for cross-origin iframes (OOPIF). Regarding cross-origin iframes, I think we don't have to ensure the order because they live in different agents that are not observable (right?).

> I wonder if we could just include the main frame when iterating and posting the activation tasks to frames so that there are no dependencies between activation at all? @nhiroki is there a technical reason to prefer the main frame or were you just going by my original text in this PR?

I think it's OK not to ensure the order among frames including the main frame, probably iterating them unintentionally keeps the order as above though. The doc just followed the original text :)

domenic:
> In the activate algorithm, the list of the descendant browsing contexts lists frames from parent to children, and the queue a global task on the networking task source keeps the order to dispatch the event because these frames are in the same agent except for cross-origin iframes (OOPIF). Regarding cross-origin iframes, I think we don't have to ensure the order because they live in different agents that are not observable (right?).

Yes, I agree with this. Is this OK? Or should we give implementations more freedom to pick an arbitrary order?

I generally prefer things to be more deterministic when possible, but if it'd be very hard to implement then we should reconsider.

bokand:
Guaranteed order only within same agent sounds fine to me.

In that case, lets avoid special-casing main, maybe something like:

- Let _bcList_ be _successorBC’s_ active document's  `list of the descendant browsing contexts`
- Prepend _successorBC_ to _bcList_
- For each _bc_ in _bcList_
  - Queue a global task....

WDYT?

nhiroki:
+1 to ensuring the order within same agent and avoiding special-casing main. These should be implementable.

domenic:
Great. @bokand want to do a final review?

bokand:
LGTM, thanks! (PR authors can't mark the PR approved but lgtm to merge)

jeremyroman:
I don't feel strongly, but we could also have considered updating the state and then firing the event in all windows with a common agent (i.e., those that can script each other anyway) synchronously, which would have the upside that those windows wouldn't be able to observe two windows in the same TLBC with different prerendering state. That might be slightly more predictable, though I'm not sure it's a big deal.

Do we know what other state updates do (i.e., those that were synchronous across all BCs before site isolation made that impossible in Chromium)?

domenic:
> Do we know what other state updates do (i.e., those that were synchronous across all BCs before site isolation made that impossible in Chromium)?

I think this is a good guiding question. Examples of such might be:

- `window.name`? But no event.
- `history.length`? But no event, I think.
- `navigator.languages` + `languagechange` event
- `navigator.onLine` + `online`/`offline` events

PR #45, opened 2021-02-19
Restrict more disurptive features

domenic:
This batch includes delaying clipboard, screen wake lock, and generic sensor APIs. It also notes that clipboard events and navigator.share() are activation-gated.

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/pull/45.html" title="Last updated on Feb 25, 2021, 6:54 PM UTC (1f7a61d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/45/789b4c0...1f7a61d.html" title="Last updated on Feb 25, 2021, 6:54 PM UTC (1f7a61d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -80,6 +80,10 @@ spec: clipboard-apis; urlPrefix: https://w3c.github.io/clipboard-apis/
 spec: screen-wake-lock; urlPrefix: https://w3c.github.io/screen-wake-lock/
   type: method; for: WakeLock
     text: request(); url: the-request-method
+spec: web-nfc; urlPrefix: https://w3c.github.io/web-nfc/
+  type: method; for: NDEFReader
+    text: write(); url: dom-ndefreader-write
+    text: scan(); url: dom-ndefreader-scan
 </pre>
 <pre class="biblio">
 {
@@ -672,6 +676,20 @@ TODO: what about the service worker API? Depends on what we're doing for service
   1. If [=this=].{{Sensor/[[state]]}} is "`idle`", then return.
 
       <p class="note">This ensures that if this portion of the algorithm was delayed due to prerendering, and in the meantime {{Sensor/stop()}} was called, we do nothing upon activating the prerender.
+</div>
+
+<h4 id="patch-web-nfc">Web NFC</h4>
+
+<div algorithm="NDEFReader write patch">
+  Modify the {{NDEFReader/write()}} method steps by inserting the following steps after the initial creation of |p|:
+
+  1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=] and return |p|.
+</div>
+
+<div algorithm="NDEFReader scan patch">
+  Modify the {{NDEFReader/scan()}} method steps by inserting the following steps after the initial creation of |p|:
+
+  1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=] and return |p|.
 </div>
 
 <h4 id="activation-gated">Activation-gated APIs</h4>
--- a/index.bs
+++ b/index.bs
@@ -681,6 +681,8 @@ TODO: what about the service worker API? Depends on what we're doing for service
   1. If [=this=].{{Sensor/[[state]]}} is "`idle`", then return.
 
       <p class="note">This ensures that if this portion of the algorithm was delayed due to prerendering, and in the meantime {{Sensor/stop()}} was called, we do nothing upon activating the prerender.
+
+  1. Assert: [=this=].{{Sensor/[[state]]}} is "`activating`".
 </div>
 
 <h4 id="patch-web-nfc">Web NFC</h4>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -627,7 +648,64 @@ TODO: what about the service worker API? Depends on what we're doing for service
...
+
+  1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=] and return |p|.
+</div>
+
+<h4 id="patch-wake-lock">Screen Wake Lock API</h4>
+
+<div algorithm="WakeLock request patch">
+  Modify the {{WakeLock/request()}} method steps by inserting the following steps after the mid-algorithm creation of |promise|:
+
+  1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=] and return |promise|.
```
mfalken, line 670: Just curious: The algorithm request() algorithm uses the browsing context of the "responsible document of the current settings object". Is this another way to get the same browsing context as "this's relevant global object's browsing context"? When we patch these specs should we make these consistent?

index.bs
```diff
@@ -627,7 +648,64 @@ TODO: what about the service worker API? Depends on what we're doing for service
...
+
+<h4 id="patch-generic-sensor">Generic Sensor API</h4>
+
+<div algorithm="Sensor start patch">
+  Modify the {{Sensor/start()}} method steps by inserting the following steps after the state is set to "`activating`":
+
+  1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=] and return.
+  1. If [=this=].{{Sensor/[[state]]}} is "`idle`", then return.
+
+      <p class="note">This ensures that if this portion of the algorithm was delayed due to prerendering, and in the meantime {{Sensor/stop()}} was called, we do nothing upon activating the prerender.
```
mfalken, line 683: I thought this may break if `start()` is called twice, but I think it works too. Should we add an ASSERT(state is 'activating') after the idle check?

index.bs
```diff
@@ -627,7 +648,64 @@ TODO: what about the service worker API? Depends on what we're doing for service
...
+
+  1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=] and return |p|.
+</div>
+
+<h4 id="patch-battery">Battery Status API</h4>
+
+<div algorithm="Navigator getBattery patch">
+  Modify the {{Navigator/getBattery()}} method steps by inserting the following after the initial secure context check:
+
+  1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=] and return [=this=]'s [=battery promise=].
```
mfalken, line 705: Check for understanding: We do the secure context check during prerendering and delay the "allowed to use" check because secure context is immutable while "allowed to use" is mutable? In the Geolocation spec I think we are delaying all checks until after prerendering.

index.bs
```diff
@@ -638,6 +716,8 @@ The following APIs do not need modifications, because they will automatically fa
 - {{PresentationRequest/start()|presentationRequest.start()}} [[PRESENTATION-API]]
 - {{Window/showOpenFilePicker()}}, {{Window/showSaveFilePicker()}}, and {{Window/showDirectoryPicker()}} [[FILE-SYSTEM-ACCESS]]
 - {{IdleDetector/requestPermission()|IdleDetector.requestPermission()}} [[IDLE-DETECTION]]
+- Firing of clipboard events, as well as {{Clipboard/read()|navigator.clipboard.read()}} and {{Clipboard/readText()|navigator.clipboard.readText()}} [[CLIPBOARD-APIS]]
```
mfalken, line 719: Ah, thanks for adding this. I was initially not understanding why write() is patched and read() isn't.

index.bs
domenic: Following the definitions, they are indeed the same, although it's not obvious. I can work on fixing Fetch to use this. (We'd eventually like to remove "responsible document" per https://github.com/whatwg/html/issues/4335.)

index.bs
domenic: This is a good point. My reasoning was indeed something like "if it's going to fail always (because of a non-secure context), we might as well fail early". But indeed, in geolocation we delay all the checks until afterward. This is mainly because the existing geolocation spec bundles up secure context and "allowed to use" into a single step, so it would have been more awkward to split them.

I see a few approaches:

- Split out all immutable checks. This requires more case-by-case analysis.
- Split out secure context checks specifically. These should be rare as most APIs use `[SecureContext]` these days; it's only very old APIs that have explicit secure context checks.
- Always defer all checks. Nice and simple and consistent.

To be clear, the first two options here would involve updating geolocation to make it more like getBattery().

Any thoughts? I think any of them is fine, personally.

mfalken: Hm, I agree any of them seem fine. If I were to choose, I'd go with "Always defer all checks", just because it seems the most straightforward and least error-prone.

mfalken (approved):
lgtm with the comments. There's a typo in the issue title "disurptive".

index.bs
domenic: OK, let's always defer all checks. On the spec side that'll depend on https://github.com/w3c/battery/pull/30, but I'll make that clear.

Issue #59, opened 2021-04-27
visibilityState for prerendering browsing contexts


mfalken:
Chromium's implementation is using `document.visibilityState == 'hidden'` for prerendering browsing contexts:
https://chromium-review.googlesource.com/c/chromium/src/+/2821563

We should make sure the impl and spec match here. cc @horo-t and @bokand does this match what you were planning here?

EDIT: typed too quickly, visibilityState == 'hidden' is what the test is using, not 'prerender'.

bokand:
Yup, I think `visibilityState == 'hidden'` in a prerender makes sense. Though in a portal it may be visible.

It's stated a bit awkwardly with two `IF` statements and no else, but I think prerender already does the right thing following [the specified steps](https://www.w3.org/TR/page-visibility/#dfn-steps-to-determine-the-visibility-state). At least, it definitely shouldn't return "visible". Maybe we should add an explicit statement to the "hidden" branch for prerendering? @mfalken is that what you meant about spec or in prerendering specific specs?

mfalken:
Makes sense, thanks. I mostly just wanted to make sure we're on the same page. Regarding spec language, if it's already implied by the Page Visibility spec, it could possibly just be a non-normative note in the prerendering spec. The Editor's Draft of the Page Visibility spec seems to have moved most of the cases to a non-normative note also: https://w3c.github.io/page-visibility/#visibility-states

mfalken:
Fixed in https://github.com/jeremyroman/alternate-loading-modes/pull/62

Issue #55, opened 2021-03-16
Clarify prerendering documents cannot have focus. 


Clqsin45:
Some APIs's specs[1][2] require documents to have focus[3] when using these APIs. So for prerendering,  maybe we should clarify that prerendering documents cannot be focused and executing `document.hasFocus()` on these pages always returns `false`[4].

Maybe we should also clarify that prerendering browsing context cannot be active?  

cc: @nhiroki 

[1] https://w3c.github.io/clipboard-apis/#privacy-async
[2] https://www.w3.org/TR/pointerlock/#extensions-to-the-element-interface
[3] https://html.spec.whatwg.org/multipage/interaction.html#focus 
[4] https://chromium-review.googlesource.com/c/chromium/src/+/2743865/5..9/third_party/blink/web_tests/wpt_internal/prerender/resources/document-has-focus.html#b17

domenic:
We only very recently added the concept of [system focus](https://html.spec.whatwg.org/#tlbc-system-focus) to the HTML spec. Although, we didn't hook it up to `document.hasFocus()`, hmm.

My take, based on previous discussions, is that prerendering browsing contexts can have their contents focused (e.g., they can have a nontrivial `document.activeElement`; `someElement.focus()` works; `autofocus=""` works). But, they cannot have system focus.

a4sriniv:
I agree with @domenic, prerendering browsing contexts should behave the same as portals do currently (they don't have system focus and only get it on activation). See WICG/portals#257 for the portals discussion on focus.

Clqsin45:
Thank you for discussing! 
I've added wpt-internal tests.

mfalken:
Thanks @Clqsin45 for raising this. We could probably extend the section on Activation-gated APIs to also cover System Focus-gated APIs:
https://jeremyroman.github.io/alternate-loading-modes/#activation-gated

mfalken:
It appears some APIs like Picture-In-Picture may also be gated on visibility state as well as user activation (e.g., https://w3c.github.io/picture-in-picture/#auto-pip can be used without user activation if visibility state changes??), so we may also add visibility state to this section. Filed related #59

mfalken:
This is fixed in https://github.com/jeremyroman/alternate-loading-modes/pull/62 and the Clipboard case specifically in https://github.com/jeremyroman/alternate-loading-modes/pull/69. Thanks @Clqsin45  for raising this!

PR #62, opened 2021-05-15
Introduce "implicitly restricted APIs".

mfalken:
This expands "Activated-Gated APIs" to cover other reasons that an API
can be implicitly restricted, namely, if it requires system focus or
page visibility.

It also clarifies that prerendering browsing contexts never have system
focus or page visibility, as per discussion in #59 and #55.

***
<a href="https://pr-preview.s3.amazonaws.com/mfalken/alternate-loading-modes/pull/62.html" title="Last updated on May 18, 2021, 11:44 PM UTC (cc0ad5c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/62/9983980...mfalken:cc0ad5c.html" title="Last updated on May 18, 2021, 11:44 PM UTC (cc0ad5c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -37,6 +37,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: prompt to unload; url: prompt-to-unload-a-document
       text: refused to allow the document to be unloaded; url: refused-to-allow-the-document-to-be-unloaded
       text: traverse the history; url: traverse-the-history
+    urlPrefix: interaction.html
+      text: system focus; url: tlbc-system-focus
     urlPrefix: urls-and-fetching.html
       text: parse a URL; url: parse-a-url
       text: resulting URL record
@@ -45,6 +47,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
 spec: page-visibility; urlPrefix: https://w3c.github.io/page-visibility/
   type: dfn
     text: hidden; for: Document; url: dfn-hidden
+    text: visible; for: Document; url: dfn-visible
+    text: visibility state; for: Document; url: dfn-visibility-states
 spec: geolocation; urlPrefix: https://w3c.github.io/geolocation-api/
   type: method; for: Geolocation
     text: getCurrentPosition(successCallback, errorCallback, options); url: getcurrentposition-method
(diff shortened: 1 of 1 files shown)


mfalken:
@domenic can you review this? While writing this I noticed the discussion in #59 is opposite what the current spec says, so we can also continue discussion on #59 if you think it should be "visible".

index.bs
```diff
@@ -602,6 +613,16 @@ Modify the [=discard a document|discard=] algorithm for {{Document}}s by appendi
   1. [=list/Empty=] <var ignore>document</var>'s [=Document/post-prerendering activation steps list=].
 </div>
 
+<h2 id="interaction-with-other-specs">Interaction with other specifications and concepts</h2>
+
+<h3 id="interaction-with-visibility-state">Interaction with Page Visibility</h3>
+
+Documents in [=prerendering browsing contexts=] are [=Document/hidden=].
```
domenic, line 620: Note that for portals this might not be true, and portals are envisioned to be a type of prerendering browsing context. We might want to state that subtlety here as a TODO or Note, even though portals spec work is kind of on hold while we get these foundations worked out...

index.bs
```diff
@@ -853,3 +870,9 @@ The following APIs do not need modifications, because they will automatically fa
 - Firing of clipboard events, as well as {{Clipboard/read()|navigator.clipboard.read()}} and {{Clipboard/readText()|navigator.clipboard.readText()}} [[CLIPBOARD-APIS]]
 - {{Navigator/share()|navigator.share()}} [[WEB-SHARE]]
 - {{Element/requestPointerLock()|element.requestPointerLock()}} [[POINTERLOCK]]
+
+APIs that require the [=Document/visible=] [=Document/visibility state=]:
+- {{WakeLock/request()}} [[WAKE-LOCK]]
```
domenic, line 875:
```suggestion
- {{WakeLock/request()|navigator.wakeLock.request()}} [[WAKE-LOCK]]
```

domenic (approved):
LGTM. I agree the existing text about "do not count as hidden" is wrong; maybe I was thinking of portals.

index.bs
mfalken: I made this change, also reverted to [SCREEN-WAKE-LOCK] which I didn't realize the original spec had instead of [WAKE-LOCK]. I don't know what source code that comes from.

index.bs
mfalken: Done, added it as an Issue which seemed consistent with the other Issues about portals.

mfalken:
Thanks, done! I added a commit though the diff still seems to show the old one. Would you be able to merge if it's ready? I don't have write access.

PR #63, opened 2021-05-18
Specify [DelayWhilePrerendering]

domenic:
This abstracts away some of the boilerplate involved in the simpler promise-returning methods, and improves uniformity by ensuring that we always delay all checks. Previously we would sometimes perform checks eagerly since it was easiest to modify the spec by inserting the delay right before promise creation.

@mfalken @jeremyroman curious on your thoughts! I kind of like it. And it also gives future spec authors something relatively easy to do.

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/pull/63.html" title="Last updated on Jun 11, 2021, 3:26 PM UTC (699b381)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/63/67eb340...699b381.html" title="Last updated on Jun 11, 2021, 3:26 PM UTC (699b381)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -17,6 +17,9 @@ spec:html; type:element; text:link
 spec:html; type:element; text:script
 </pre>
 <pre class="anchors">
+spec: ecma262; urlPrefix: https://tc39.es/ecma262/
+  type: dfn
+    text: current Realm; url: current-realm
 spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
     urlPrefix: browsers.html
@@ -694,6 +697,26 @@ Modify the <a spec=HTML>download the hyperlink</a> algorithm to ensure that down
 
 Many specifications need to be patched so that, if a given algorithm invoked in a [=prerendering browsing context=], most of its work is deferred until the browsing context is [=prerendering browsing context/activated=]. This is tricky to do uniformly, as many of these specifications do not have great hygeine around <a href="https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-for-spec-authors">using the event loop</a>. Nevertheless, the following sections give our best attempt.
 
+<h4 id="delay-while-prerendering">The {{[DelayWhilePrerendering]}} extended attribute</h4>
+
+To abstract away some of the boilerplate involved in delaying the action of asynchronous methods until [=prerendering browsing context/activated|activation=], we introduce the <dfn extended-attribute>[DelayWhilePrerendering]</dfn> Web IDL extended attribute. It indicates that when a given method is called in a [=prerendering browsing context=], it will immediately return a pending promise and do nothing else. Only upon activation will the usual method steps take place, with their result being used to resolve or reject the previously-returned promise.
+
+The {{[DelayWhilePrerendering]}} extended attribute must take no arguments, and must only appear on a <a lt="regular operation" spec="WEBIDL">regular</a> or <a spec="WEBIDL">static operation</a> whose <a spec="WEBIDL">return type</a> is a <a spec="WEBIDL">promise type</a> and whose <a spec="WEBIDL">exposure set</a> contains {{Window}}.
+
+<div algorithm="DelayWhilePrerendering method steps">
+  The method steps for any operation annotated with the {{[DelayWhilePrerendering]}} extended attribute are replaced with the following:
+
+  1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then:
+    1. Let |realm| be the [=current Realm=].
+    1. If the operation in question is a <a spec="WEBIDL">regular operation</a>, then set |realm| to the [=relevant Realm=] of [=this=].
+    1. Let |promise| be [=a new promise=], created in |realm|.
+    1. Append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=]:
+      1. Let |result| be the result of running the originally-specified steps for this operation, with the same [=this=] and arguments.
+      1. [=Resolve=] |promise| with |result|.
+    1. Return |promise|.
+  1. Otherwise, return the result of running the originally-specified steps for this operation, with the same [=this=] and arguments.
+</div>
+
 <h4 id="patch-geolocation">Geolocation API</h4>
 
 <div algorithm="Geolocation getCurrentPosition patch">
@@ -724,11 +747,9 @@ Many specifications need to be patched so that, if a given algorithm invoked in 
 
 <h4 id="patch-serial">Web Serial API</h4>
 
-<div algorithm="Serial requestPort patch">
-  Modify the {{Serial/requestPort()}} method steps by inserting the following steps after the initial creation of |promise|:
+Add {{[DelayWhilePrerendering]}} to {{Serial/requestPort()}}.
 
-  1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=] and return |promise|.
-</div>
+TODO: the below could probably be done by generalizing {{[DelayWhilePrerendering]}} to use owner document while in dedicated workers.
 
 <div algorithm="Serial getPorts patch">
   Modify the {{Serial/getPorts()}} method steps by inserting the following steps after the initial creation of |promise|:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -675,11 +698,9 @@ Many specifications need to be patched so that, if a given algorithm invoked in
 
 <h4 id="patch-serial">Web Serial API</h4>
 
-<div algorithm="Serial requestPort patch">
-  Modify the {{Serial/requestPort()}} method steps by inserting the following steps after the initial creation of |promise|:
+Add {{[DelayWhilePrerendering]}} to {{Serial/requestPort()}}.
 
-  1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=] and return |promise|.
-</div>
+TODO: the below could probably be done by generalizing {{[DelayWhilePrerendering]}} to use owner document while in dedicated workers.
```
domenic, line 703: I'm happy to spec this out but I wanted to give you all a chance to see the simple version first.

index.bs
```diff
@@ -645,6 +648,26 @@ Modify the <a spec=HTML>download the hyperlink</a> algorithm to ensure that down
...
 
+<h4 id="delay-while-prerendering">The {{[DelayWhilePrerendering]}} extended attribute</h4>
+
+To abstract away some of the boilerplate involved in delaying the action of asynchronous methods until [=prerendering browsing context/activated|activation=], we introduce the <dfn extended-attribute>[DelayWhilePrerendering]</dfn> Web IDL extended attribute. It indicates that when a given method is called in a [=prerendering browsing context=], it will immediately return a pending promise and do nothing else. Only upon activation will the usual method steps take place, with their result being used to resolve or reject the previously-returned promise.
+
+The {{[DelayWhilePrerendering]}} extended attribute must take no arguments, and must only appear on a <a lt="regular operation" spec="WEBIDL">regular</a> or <a spec="WEBIDL">static operation</a> whose <a spec="WEBIDL">return type</a> is a <a spec="WEBIDL">promise type</a> and whose <a spec="WEBIDL">exposure set</a> contains {{Window}}.
```
mfalken, line 655: Eventually we will probably want to support Workers/Worklets but it will need some sort of recursive definition that's more complex than Chromium has to deal with because according to the spec workers can spawn workers. Something like:
- a service worker is prerendering if all its clients are prerendering
- a shared worker is prerendering if all windows on the origin are prerendering
- a dedicated worker is prerendering if its creator is prerendering
- a worklet is prerendering if its creator is prerendering

and hopefully that eventually reaches the base case of prerendering browsing context, but I'm not sure yet. We can make this a TODO/issue.

index.bs
mfalken: Ah yep, I wrote the above before seeing this. For dedicated workers we should be able to do something like that, except shared workers can make dedicated workers.

For the initial origin trial in Chrome, we are planning to just silently defer the creation of dedicated workers until after activation.

mfalken (commented):
I like this direction at least spec-wise. Unfortunately I am not very familiar with bindings and how difficult it would be to implement in Blink (cc @kinu @bashi early on). Some comments on workers follow.

index.bs
jeremyroman: Might we we eventually want it on promise properties, too? Or would we rather handle that explicitly?

index.bs
```diff
@@ -645,6 +648,26 @@ Modify the <a spec=HTML>download the hyperlink</a> algorithm to ensure that down
...
+
+<div algorithm="DelayWhilePrerendering method steps">
+  The method steps for any operation annotated with the {{[DelayWhilePrerendering]}} extended attribute are replaced with the following:
+
+  1. Let |realm| be the [=current Realm=].
+  1. If the operation in question is a <a spec="WEBIDL">regular operation</a>, then set |realm| to the [=relevant Realm=] of [=this=].
+  1. If |realm|'s [=Realm/global object=]'s [=Window/browsing context=] is a [=prerendering browsing context=], then:
+    1. Let |promise| be [=a new promise=], created in |realm|.
+    1. Append the following steps to [=this=]'s [=platform object/post-prerendering activation steps list=]:
```
jeremyroman, line 664: What do you think about replacing the "post-prerendering activation steps list" construct with the machinery in https://heycam.github.io/webidl/#es-promise-manipulation?

Then we could use the WebIDL concept of reacting to a promise here instead:

```
1. Let |preconditions| be an empty [=list=].
1. If |realm|'s ... is a [=prerendering browsing context=], then:
  1. Append [=this=]'s [=prerendering promise=] to |preconditions|.
1. If |preconditions| is [=empty=], run the originally-specified steps for this operation, with the same [=this=] and arguments, and return the result.
1. [=Upon fulfillment=] of |preconditionPromise|, run the originally-specified steps for this operation, with the same [=this=] and arguments, and return the result.
```

(The silly step for the no preconditions case is probably necessary so that the immediate steps don't get delayed by until the microtask checkpoint.)

I personally like this construct better than data structures that store lists of steps, but your preference may vary, or I may be overthinking.

jeremyroman (commented):
Two natural questions:

- Is this the first extended attribute to modify method steps, as opposed to providing them or specifying requirements on them?
- Are we okay with letting script making potentially slightly less progress in error cases that could be detecting from early argument validation, in exchange for making this easier to sprinkle around?
- Will we still want to wait at other points in steps (besides the beginning of method steps)? If so, should we instead or also specify some sort of WHATWG-infra equivalent of the ECMAScript "await" construct to make it easy to wait for things without nesting another set of steps?

domenic:
> Is this the first extended attribute to modify method steps, as opposed to providing them or specifying requirements on them?

It is the second, after [`[CEReactions]`](https://html.spec.whatwg.org/#cereactions) (which also modifies attributes, setters, and deleters).

> Are we okay with letting script making potentially slightly less progress in error cases that could be detecting from early argument validation, in exchange for making this easier to sprinkle around?

That's the big question. IMO yes because it's a more predictable pattern for web developers: literally nothing will happen, including any error handling, before activation (for such APIs).

> Will we still want to wait at other points in steps (besides the beginning of method steps)? If so, should we instead or also specify some sort of WHATWG-infra equivalent of the ECMAScript "await" construct to make it easy to wait for things without nesting another set of steps?

Probably; indeed there are several cases still in the document. They use the pattern

> If this's relevant global object's browsing context is a prerendering browsing context, then append the following steps to this's post-prerendering activation steps list and return.

which is slightly verbose and could probably be abstracted into something like "Suspend if `this` is prerendering", but it doesn't have the nesting problem.

However what you can see from the remaining non-`[DelayWhilePrerendering]` steps is that this is the exception: most need more custom behavior. So such an abstraction wouldn't see too much use.

index.bs
domenic: There aren't many promise properties on the platform, so my plan would be to leave it out until we find one. (The only ones I know of are in streams.)

index.bs
domenic: I guess in your example `|preconditions|` is meant to be `|preconditionPromise|` and not be a list, but instead a promise-or-null or something like that? It seems reasonable, although I guess personally I get worried about using non-developer-exposed promises, as promises are _so_ JavaScriptey (e.g., realm specific) so they can be awkward to use as a spec-internal mechanism.

index.bs
```diff
@@ -645,6 +648,26 @@ Modify the <a spec=HTML>download the hyperlink</a> algorithm to ensure that down
...
 
+<h4 id="delay-while-prerendering">The {{[DelayWhilePrerendering]}} extended attribute</h4>
+
+To abstract away some of the boilerplate involved in delaying the action of asynchronous methods until [=prerendering browsing context/activated|activation=], we introduce the <dfn extended-attribute>[DelayWhilePrerendering]</dfn> Web IDL extended attribute. It indicates that when a given method is called in a [=prerendering browsing context=], it will immediately return a pending promise and do nothing else. Only upon activation will the usual method steps take place, with their result being used to resolve or reject the previously-returned promise.
+
+The {{[DelayWhilePrerendering]}} extended attribute must take no arguments, and must only appear on a <a lt="regular operation" spec="WEBIDL">regular</a> or <a spec="WEBIDL">static operation</a> whose <a spec="WEBIDL">return type</a> is a <a spec="WEBIDL">promise type</a> and whose <a spec="WEBIDL">exposure set</a> contains {{Window}}.
+
+<div algorithm="DelayWhilePrerendering method steps">
+  The method steps for any operation annotated with the {{[DelayWhilePrerendering]}} extended attribute are replaced with the following:
```
jeremyroman, line 658: I wonder if we'll eventually need to specify how this stacks with `[CEReactions]`. Not today, though. :)

jeremyroman (approved):
Sorry I haven't gotten back to this. I don't think I have any real reservations here so I'm okay with this if @mfalken is.

mfalken (approved):
lgtm to move forward! (while we might learn later we want to do a different model, either due to implementation constraints or behaviors we don't want to defer)

Issue #64, opened 2021-05-20
Expose the time of activation to the prerendered page


horo-t:
Currently, on prerendered pages, there is no way to know when the user clicked a link to activate the page.

So it is difficult to compare the performance difference between the normal page transition and the prerendered page transition.

We should have a way to expose the time of activation to the prerendered page.

My current idea is introducing PerformanceNavigationTiming.activationStart attribute:

```
  partial interface PerformanceNavigationTiming {
      readonly attribute DOMHighResTimeStamp activationStart;
  };
```

Chromium side doc: https://docs.google.com/document/d/1EXNO9BmqBvB0oe3EHuIxcUioVoMti8tAdMb8E8ACCW0/edit#

PR #65, opened 2021-05-20
PerformanceNavigationTiming.activationStart

horo-t:
This will fix the issue https://github.com/jeremyroman/alternate-loading-modes/issues/64 .

***
<a href="https://pr-preview.s3.amazonaws.com/horo-t/alternate-loading-modes/pull/65.html" title="Last updated on May 27, 2021, 12:15 AM UTC (00f3462)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/65/00e0523...horo-t:00f3462.html" title="Last updated on May 27, 2021, 12:15 AM UTC (00f3462)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -373,6 +373,22 @@ The <dfn attribute for="Document">onprerenderingchange</dfn> attribute is an [=e
 
 As is customary for [[HTML]], the definition of {{Document/prerendering}} would be located in another section of the spec; we'd place it in the new section introduced below:
 
+<h3 id="performance-navigation-timing-extension">Extensions to the {{PerformanceNavigationTiming}} interface</h3>
+
+We'd modify the definition of {{PerformanceNavigationTiming}} as follows:
+
+<pre class="idl">
+  partial interface PerformanceNavigationTiming {
+      readonly attribute DOMHighResTimeStamp activationStart;
+  };
+</pre>
+
+When getting the value of the <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> attribute, run the following steps:
+
+1. If the current [=browsing context=] has not been [=prerendering browsing context/activate|activatted=], return a {{DOMHighResTimeStamp}} with a time value equal to zero.
+
+1. Otherwise, return a {{DOMHighResTimeStamp}} with a time value equal to the time immediately after [=Prompt to unload|prompting to unload=] the previous document. 
+
 <h3 id="prerendering-bcs-subsection">Prerendering browsing contexts</h3>
 
 <em>The following section would be added as a new sub-section of [[HTML]]'s <a href="https://html.spec.whatwg.org/multipage/browsers.html#windows">Browsing contexts</a> section.</em>
--- a/index.bs
+++ b/index.bs
@@ -121,6 +121,9 @@ spec: picture-in-picture; urlPrefix: https://w3c.github.io/picture-in-picture/
     text: autoPictureInPicture; url: dom-htmlvideoelement-requestpictureinpicture
   type: method; for: HTMLVideoElement
     text: requestPictureInPicture(); url: dom-htmlvideoelement-autopictureinpicture
+spec: High Resolution Time; urlPrefix: https://w3c.github.io/hr-time/
+  type: dfn
+    text: relative high resolution coarse time; url: dfn-relative-high-resolution-coarse-time
 </pre>
 <pre class="biblio">
 {
@@ -412,7 +415,7 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps 
   :: 1. Return |platformObject|'s [=relevant global object=]'s [=associated Document=]'s [=Document/post-prerendering activation steps list=].
 </dl>
 
-Every {{Document}} has a <dfn for="Document">activation start time</dfn>, which is initially a {{DOMHighResTimeStamp}} with a time value equal to zero. This is used to keep the time of immediately after [=Prompt to unload|prompting to unload=] the [=active document=] of previous [=browsing context=].
+Every {{Document}} has a <dfn for="Document">activation start time</dfn>, which is initially a {{DOMHighResTimeStamp}} with a time value equal to zero. This is used to keep the time of immediately after [=Prompt to unload|prompting to unload=] the [=active document=] of the previous [=browsing context=].
 
 <div algorithm="create a prerendering browsing context">
   To <dfn export>create a prerendering browsing context</dfn> given a [=URL=] |startingURL|, a [=referrer policy=] |referrerPolicy|, and a {{Document}} |referrerDoc|:
@@ -447,13 +450,13 @@ Every {{Document}} has a <dfn for="Document">activation start time</dfn>, which 
 
   1. Assert: |predecessorBC| is a [=top-level browsing context=].
 
-  <!-- The following are copied from the navigate algorithm, and probably could benefit from some refactoring to deduplicate. -->
+  <!-- The following except for the step that sets the |activation start time| are copied from the navigate algorithm, and probably could benefit from some refactoring to deduplicate. -->
 
   1. Cancel any preexisting but not yet [=navigate/mature=] attempts to navigate |predecessorBC|, including canceling any instances of the [=fetch=] algorithm started by those attempts. If one of those attempts has already <a lt="create and initialize a Document object">created and initialized a new `Document` object</a>, [=Document/abort=] that {{Document}} also.
 
   1. [=Prompt to unload=] the [=active document=] of |predecessorBC|. If the user [=refused to allow the document to be unloaded=], then return.
 
-  1. Let |activation start time| be the present time.
+  1. Let |activation start time| be the [=coarsened shared current time=] given the |predecessorBC|'s [=active document=]'s [=relevant settings object=]'s [=environment settings object/cross-origin isolated capability=].
 
   1. [=Document/Abort=] the [=active document=] of |predecessorBC|.
 
@@ -474,7 +477,7 @@ Every {{Document}} has a <dfn for="Document">activation start time</dfn>, which 
 
       1. Let |doc| be |bc|'s [=active document=].
 
-      1. Set |doc|'s [=Document/activation start time=] to |activation start time|.
+      1. If |doc|'s [=Document/origin=] is the same as |predecessorBC|'s [=active document=]'s [=Document/origin=], then set |doc|'s [=Document/activation start time=] to the [=relative high resolution coarse time=] for |activation start time| and |doc|.
 
       1. [=Fire an event=] named {{Document/prerenderingchange}} at |doc|.
 
--- a/index.bs
+++ b/index.bs
@@ -124,6 +124,9 @@ spec: picture-in-picture; urlPrefix: https://w3c.github.io/picture-in-picture/
 spec: High Resolution Time; urlPrefix: https://w3c.github.io/hr-time/
   type: dfn
     text: relative high resolution coarse time; url: dfn-relative-high-resolution-coarse-time
+spec: Navigation Timing Level 2; urlPrefix: https://www.w3.org/TR/navigation-timing-2/
+  type: dfn
+    text: current document; url: dfn-current-document
 </pre>
 <pre class="biblio">
 {
@@ -415,7 +418,7 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps 
   :: 1. Return |platformObject|'s [=relevant global object=]'s [=associated Document=]'s [=Document/post-prerendering activation steps list=].
 </dl>
 
-Every {{Document}} has a <dfn for="Document">activation start time</dfn>, which is initially a {{DOMHighResTimeStamp}} with a time value equal to zero. This is used to keep the time of immediately after [=Prompt to unload|prompting to unload=] the [=active document=] of the previous [=browsing context=].
+Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which is initially a {{DOMHighResTimeStamp}} with a time value equal to zero. This is used to keep the time of immediately after [=Prompt to unload|prompting to unload=] the [=active document=] of the previous [=browsing context=].
 
 <div algorithm="create a prerendering browsing context">
   To <dfn export>create a prerendering browsing context</dfn> given a [=URL=] |startingURL|, a [=referrer policy=] |referrerPolicy|, and a {{Document}} |referrerDoc|:
@@ -646,7 +649,7 @@ Extend the {{PerformanceNavigationTiming}} interface as follows:
 
 When getting the value of the <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> attribute, run the following steps:
 
-1. Return a {{DOMHighResTimeStamp}} with a time value equal to the current document's [=Document/activation start time=].
+1. Return a {{DOMHighResTimeStamp}} with a time value equal to the [=current document=]'s [=Document/activation start time=].
 
 <h2 id="nonsense-behaviors">Preventing nonsensical behaviors</h2>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -373,6 +373,22 @@ The <dfn attribute for="Document">onprerenderingchange</dfn> attribute is an [=e
 
 As is customary for [[HTML]], the definition of {{Document/prerendering}} would be located in another section of the spec; we'd place it in the new section introduced below:
 
+<h3 id="performance-navigation-timing-extension">Extensions to the {{PerformanceNavigationTiming}} interface</h3>
```
mfalken, line 376: I think this belongs more in the "Interaction with other specifications and concepts" section.

index.bs
```diff
@@ -373,6 +373,22 @@ The <dfn attribute for="Document">onprerenderingchange</dfn> attribute is an [=e
...
+
+We'd modify the definition of {{PerformanceNavigationTiming}} as follows:
```
mfalken, line 378: I think we can say "Extend the {{PerformanceNavigationTiming}} interface as follows:"

index.bs
```diff
@@ -373,6 +373,22 @@ The <dfn attribute for="Document">onprerenderingchange</dfn> attribute is an [=e
...
+
+<pre class="idl">
+  partial interface PerformanceNavigationTiming {
+      readonly attribute DOMHighResTimeStamp activationStart;
+  };
+</pre>
+
+When getting the value of the <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> attribute, run the following steps:
+
+1. If the current [=browsing context=] has not been [=prerendering browsing context/activate|activatted=], return a {{DOMHighResTimeStamp}} with a time value equal to zero.
```
mfalken, line 388: This is a bit subtle with descendent browsing contexts. Only the top-level browsing context is the one where the activate steps run. And it's not clear at which point we call it "activated".

I think we need to set a timestamp at Step 5 "Prompt to unload" of the activate steps, and read that timestamp here.

mfalken (changes requested):
Overall looks good, some initial comments.

index.bs
mfalken: Actually I'm not sure what the intent is with iframes. Does each iframe's navigation entry also have a populated activationStart? For same-origin iframes, we plan to load them as normal; for cross-origin iframes, the load is delayed until after activation.

horo-t: Done.
Introduced Document's "activation start time", and the local variable "activation start time".

horo-t (commented):
Thank you very much for reviewing.

mfalken (changes requested):
Overall this looks good to me though I have remaining questions I don't think I can resolve on my own. cc @yoavweiss and @domenic for advice.

index.bs
```diff
@@ -625,6 +631,20 @@ Documents in [=prerendering browsing contexts=] are [=Document/hidden=].
...
+
+<pre class="idl">
+  partial interface PerformanceNavigationTiming {
+      readonly attribute DOMHighResTimeStamp activationStart;
+  };
+</pre>
+
+When getting the value of the <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> attribute, run the following steps:
+
+1. Return a {{DOMHighResTimeStamp}} with a time value equal to the current document's [=Document/activation start time=].
```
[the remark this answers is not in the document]
horo-t: No. Failed to make when I used [=current document=] here.

```shell
$ make
bikeshed --die-on=warning spec index.bs index.html
LINE ~649: No 'dfn' refs found for 'current document' that are marked for export.
 ✘  Did not generate, due to fatal errors
make: *** [Makefile:6: local] Error 2
```

index.bs
```diff
@@ -470,6 +474,8 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
 
       1. Let |doc| be |bc|'s [=active document=].
 
+      1. Set |doc|'s [=Document/activation start time=] to |activation start time|.
```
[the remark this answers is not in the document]
horo-t: Added the same origin check, same as [`update the session history with the new page`](https://html.spec.whatwg.org/multipage/browsing-the-web.html#update-the-session-history-with-the-new-page) logic does [here](https://html.spec.whatwg.org/multipage/browsing-the-web.html#:~:text=newdocument's%20origin%20is%20the%20same%20as%20previousdocument's%20origin) for unload timing.


> Also it means documents that didn't exist before prerendering don't have an activation start time, which might be a bit inconsistent/racy. It might be OK to just provide activation start time for the top level document and subframes don't have it to start with?

I don't understand what is problem here.
If my understanding is correct, `prerenderingchange` events is called if and only if the document has been created while prerendering. What is problem if we set the activation start time if and only if the document has been created while prerendering?

index.bs
```diff
@@ -412,6 +415,8 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
   :: 1. Return |platformObject|'s [=relevant global object=]'s [=associated Document=]'s [=Document/post-prerendering activation steps list=].
 </dl>
 
+Every {{Document}} has a <dfn for="Document">activation start time</dfn>, which is initially a {{DOMHighResTimeStamp}} with a time value equal to zero. This is used to keep the time of immediately after [=Prompt to unload|prompting to unload=] the [=active document=] of the previous [=browsing context=].
```
yoavweiss, line 418: s/ a / an / ?

index.bs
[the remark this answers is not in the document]
mfalken: Yeah I think this can be OK probably. I had just imagined it could appear racy: suppose a prerendered page makes many subframes dynamically. Whether they have an activation start time would depend on whether activation started quickly enough, which seemed inconsistent. However I see now the timestamp is relative to the  time origin of the document, so it'd be negative if the document wasn't created yet, which is strange also. So I can be OK with the current patch.

mfalken (approved):
Thanks, I think this works for me. Would still be good for @domenic or @yoavweiss to take a look.

index.bs
```diff
@@ -445,12 +450,14 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
...
   1. Assert: |predecessorBC| is a [=top-level browsing context=].
 
-  <!-- The following are copied from the navigate algorithm, and probably could benefit from some refactoring to deduplicate. -->
+  <!-- The following except for the step that sets the |activation start time| are copied from the navigate algorithm, and probably could benefit from some refactoring to deduplicate. -->
 
   1. Cancel any preexisting but not yet [=navigate/mature=] attempts to navigate |predecessorBC|, including canceling any instances of the [=fetch=] algorithm started by those attempts. If one of those attempts has already <a lt="create and initialize a Document object">created and initialized a new `Document` object</a>, [=Document/abort=] that {{Document}} also.
 
   1. [=Prompt to unload=] the [=active document=] of |predecessorBC|. If the user [=refused to allow the document to be unloaded=], then return.
 
+  1. Let |activation start time| be the [=coarsened shared current time=] given the |predecessorBC|'s [=active document=]'s [=relevant settings object=]'s [=environment settings object/cross-origin isolated capability=].
```
yoavweiss, line 459: I think you want to use [current high resolution time](https://w3c.github.io/hr-time/#dfn-current-high-resolution-time) here, to make the timestamps relative to the time origin.

index.bs
```diff
@@ -470,6 +477,8 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
 
       1. Let |doc| be |bc|'s [=active document=].
 
+      1. If |doc|'s [=Document/origin=] is the same as |predecessorBC|'s [=active document=]'s [=Document/origin=], then set |doc|'s [=Document/activation start time=] to the [=relative high resolution coarse time=] for |activation start time| and |doc|.
```
yoavweiss, line 480: When will the descendant `doc`'s `activation start time` be set? I see it set only for top-level browsing contexts..

yoavweiss:
> However I see now the timestamp is relative to the time origin of the document, so it'd be negative if the document wasn't created yet, which is strange also. So I can be OK with the current patch.

It may make sense to distinguish "no activation happened because this wasn't prerendered" from "activation happened before time origin started. But maybe `document.prerendering` would enable this distinction.

index.bs
```diff
@@ -451,6 +453,8 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
 
   1. [=Prompt to unload=] the [=active document=] of |predecessorBC|. If the user [=refused to allow the document to be unloaded=], then return.
 
+  1. Let |activation start time| be the present time.
+
   1. [=Document/Abort=] the [=active document=] of |predecessorBC|.
 
   <!-- End copied section. -->
```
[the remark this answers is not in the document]
horo-t: Done

horo-t (commented):
Thank you very much for review.

index.bs
yoavweiss: I misread `activation start time` as setting the document's one rather than a separate local variable. Apologies! 
(maybe rename it to make the distinction clearer?)

horo-t: Renamed `activation start time` local variable to `coarsened time`.

index.bs
```diff
@@ -412,6 +418,8 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
   :: 1. Return |platformObject|'s [=relevant global object=]'s [=associated Document=]'s [=Document/post-prerendering activation steps list=].
 </dl>
 
+Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which is initially a {{DOMHighResTimeStamp}} with a time value equal to zero. This is used to keep the time of immediately after [=Prompt to unload|prompting to unload=] the [=active document=] of the previous [=browsing context=].
```
yoavweiss, line 421: s/equal to/of/ ?

yoavweiss: Also nit: shorter lines would be easier to review

yoavweiss: Should the "This is used" part be in a note?

yoavweiss: Nit: s/time of/time/

index.bs
```diff
@@ -445,12 +453,14 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
...
   1. Assert: |predecessorBC| is a [=top-level browsing context=].
 
-  <!-- The following are copied from the navigate algorithm, and probably could benefit from some refactoring to deduplicate. -->
+  <!-- The following except for the step that sets the |activation start time| are copied from the navigate algorithm, and probably could benefit from some refactoring to deduplicate. -->
 
   1. Cancel any preexisting but not yet [=navigate/mature=] attempts to navigate |predecessorBC|, including canceling any instances of the [=fetch=] algorithm started by those attempts. If one of those attempts has already <a lt="create and initialize a Document object">created and initialized a new `Document` object</a>, [=Document/abort=] that {{Document}} also.
 
   1. [=Prompt to unload=] the [=active document=] of |predecessorBC|. If the user [=refused to allow the document to be unloaded=], then return.
 
+  1. Let |coarsened time| be the [=coarsened shared current time=] given the |predecessorBC|'s [=active document=]'s [=relevant settings object=]'s [=environment settings object/cross-origin isolated capability=].
```
yoavweiss, line 462: Can a cross-origin isolated document prerender a non-cross-origin-isolated one? If so, this can create a mismatch in timer resolution. Would be better to base this on the COI capability of the prerendered document

index.bs
```diff
@@ -625,6 +637,20 @@ Documents in [=prerendering browsing contexts=] are [=Document/hidden=].
...
+
+<pre class="idl">
+  partial interface PerformanceNavigationTiming {
+      readonly attribute DOMHighResTimeStamp activationStart;
+  };
+</pre>
+
+When getting the value of the <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> attribute, run the following steps:
+
+1. Return a {{DOMHighResTimeStamp}} with a time value equal to the [=current document=]'s [=Document/activation start time=].
```
yoavweiss, line 652: Might be worthwhile to add a note here saying that this value is 0 in cross-origin prerenders.

index.bs
```diff
@@ -470,6 +480,8 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
```
yoavweiss, line 480: Does `|inclusiveDescendants|` include the top-level document as well? https://html.spec.whatwg.org/multipage/browsers.html#list-of-the-descendant-browsing-contexts is not super clear on this

yoavweiss:
Thanks for renaming the local variable. A few more nits and questions :)

index.bs
horo-t: Done.

index.bs
horo-t: I see.

Changed to use `unsafe shared current time` here and `relative high resolution time` below.

index.bs
horo-t: Yes.
`|inclusiveDescendants|` includes the `|successorBC|`, because is was initialized by this line.

```
Let inclusiveDescendants be « successorBC » extended with successorBC’s active document's list of the descendant browsing contexts.
```

index.bs
horo-t: Done

horo-t (commented):
Thank you.

index.bs
yoavweiss: Oh, ok. Would be better IMO to clarify this and reduce type confusion (out of scope for this PR), e.g. "Let inclusiveDescendants be an empty list. Append successorBC to inclusiveDescendants. Extend inclusiveDescendants with the list of descendant browsing contexts"

yoavweiss (approved):
LGTM!

horo-t:
Thank you very much for reviewing!
I'll create another PR for `inclusiveDescendants`.

@mfalken 
Could you please approve this PR? This is my first PR on this repository.
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

mfalken:
I've already approved it. I think it needs @domenic or @jeremyroman to approve still.

horo-t:
I see.
@jeremyroman Could you please approve this PR?

index.bs
```diff
@@ -121,6 +121,9 @@ spec: picture-in-picture; urlPrefix: https://w3c.github.io/picture-in-picture/
     text: autoPictureInPicture; url: dom-htmlvideoelement-requestpictureinpicture
   type: method; for: HTMLVideoElement
     text: requestPictureInPicture(); url: dom-htmlvideoelement-autopictureinpicture
+spec: Navigation Timing Level 2; urlPrefix: https://www.w3.org/TR/navigation-timing-2/
```
domenic, line 124:
```suggestion
spec: navigation-timing; urlPrefix: https://w3c.github.io/navigation-timing/
```

index.bs
```diff
@@ -625,6 +635,22 @@ Documents in [=prerendering browsing contexts=] are [=Document/hidden=].
...
+
+Extend the {{PerformanceNavigationTiming}} interface as follows:
+
+<pre class="idl">
+  partial interface PerformanceNavigationTiming {
+      readonly attribute DOMHighResTimeStamp activationStart;
+  };
+</pre>
+
+When getting the value of the <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> attribute, run the following steps:
```
domenic, line 648:
```suggestion
The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> getter steps are:
```

domenic (approved):
LGTM with nits

jeremyroman (approved):
lgtm

index.bs
horo-t: Done

index.bs
horo-t: Done.

horo-t (commented):
Thank you!

PR #70, opened 2021-06-04
prompt()'s "soft block" return value should be null

mfalken:
Per https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-prompt

Credit to https://chromium-review.googlesource.com/c/chromium/src/+/2927630


domenic (approved):
Oops, thanks!

Issue #42, opened 2021-02-12
Tracking issue for speccing disallowed prerendering BC features


domenic:
The prerendering browsing contexts explainer, supplemented by [this Google doc](https://docs.google.com/document/d/1zY15k_wFTik2EoxBf3_RT7YjYpFMDaeNspy15n0rtww/edit#), lists all the APIs we want to restrict. This issue is meant to track actually speccing those restrictions, in checklist form.

For some of these (e.g. cross-origin isolated) the outcome may be "don't restrict"; we can check the box when we explicitly make that decision and include a note in the spec about it being permissions-controlled but still not restricted.

* Storage access
  * Synchronous key-value stores
    - [ ] Cookies
    - [ ] localStorage
    - [ ] sessionStorage
  * Asynchronous APIs
    - [ ] IndexedDB
    - [ ] Cache storage
    - [ ] File System Access's origin-private file system
    - [ ] Storage Foundation API
* Complicated not-quite storage things
  - [ ] Shared workers
  - [ ] Service workers
  - [ ] BroadcastChannel
  - [ ] Web locks
* Permissions API-controlled features ([registry](https://w3c.github.io/permissions/#permission-registry), but the registry is not up to date)
  - [x] Geolocation #36
  - [x] Notifications #36
  - [ ] Notifications from within a service worker
  - [ ] Push
  - [x] MIDI #36
  - [ ] Camera
  - [ ] Microphone
  - [ ] Speaker selection
  - [ ] Device info
  - [ ] Background fetch
  - [ ] Background sync
  - [ ] Bluetooth
  - [ ] Persistent storage
  - [ ] Ambient light sensor
  - [ ] Accelerometer
  - [ ] Gyroscope
  - [ ] Magnetometer
  - [ ] Clipboard read
  - [ ] Clipboard write
  - [ ] Display capture
  - [ ] NFC
  - [ ] Web USB
  - [ ] WebXR Spatial Tracking
  - [x] Idle detection #36
  - [x] File System Access (not the origin-private subset) #36
* Not-allowed-by-default permissions policy features ([registry](https://github.com/w3c/webappsec-permissions-policy/blob/master/features.md)). Omitting ones already in permissions API list.
  * Asynchronous (despite not being permission-gated)
    - [ ] Battery status
    - [x] Fullscreen #36
    - [ ] Picture-in-picture
    - [ ] Encrypted media extensions
    - [ ] Web Authentication public key credentials get
    - [ ] Wake lock
[...]

PR #71, opened 2021-06-11
Specify EME delay

### LINKED ISSUE #61: Prerendering restriction: EME
At https://github.com/jeremyroman/alternate-loading-modes/pull/45 we previously decided to always defer reject/resolve of a promise, even for immutable checks that could be run before prerendering.

I started looking at this for EME: https://w3c.github.io/encrypted-media/#requestMediaKeySystemAccess

In Blink it looks like it would be slightly inconvenient to take that approach as seen in this CL: https://chromium-review.googlesource.com/c/chromium/src/+/2874764

Blink currently does all the sync checks first, returning a new promise or throwing an exception. Only after that is done does 
it makes a promise to be resolved after the async work. 
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/encryptedmedia/navigator_request_media_key_system_access.cc;l=128;drc=7fcda909a4f438f251cf98d93cf494efe35b1bd7

This actually matches the spec text fairly closely: https://w3c.github.io/encrypted-media/#requestMediaKeySystemAccess

For this case, I wonder if it'd make sense to just defer after the "Let promise be a new promise." point. But this would be inconsistent with the earlier decision in the above PR.

I'm also wondering if there is a difference between `throw a [exception] and abort these steps` and `return a promise rejected with a newly created [error]` as in the Step 1 vs Step 2 of the requestMediaKeySystemAccess() algorithm. In Blink it looks identical, both cases do: 

```javascript
 exception_state.Throw*Error();
 return ScriptPromise();
```

Any thoughts on this? @domenic @jeremyroman

domenic:
Part of #42. Closes #61.

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/pull/71.html" title="Last updated on Jun 11, 2021, 3:43 PM UTC (f67cca0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/jeremyroman/alternate-loading-modes/71/1108ba0...f67cca0.html" title="Last updated on Jun 11, 2021, 3:43 PM UTC (f67cca0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -124,9 +124,12 @@ spec: picture-in-picture; urlPrefix: https://w3c.github.io/picture-in-picture/
     text: autoPictureInPicture; url: dom-htmlvideoelement-requestpictureinpicture
   type: method; for: HTMLVideoElement
     text: requestPictureInPicture(); url: dom-htmlvideoelement-autopictureinpicture
-spec: spec: navigation-timing; urlPrefix: https://w3c.github.io/navigation-timing/
+spec: navigation-timing; urlPrefix: https://w3c.github.io/navigation-timing/
   type: dfn
     text: current document; url: dfn-current-document
+spec: encrypted-media; urlPrefix: https://w3c.github.io/encrypted-media/
+  type: method; for: Navigator
+    text: requestMediaKeySystemAccess(keySystem, supportedConfigurations); url: requestMediaKeySystemAccess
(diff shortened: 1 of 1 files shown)


mfalken (approved):
lgtm

PR #74, opened 2021-08-13
Same origin explainer

mfalken:
Explainer to help with TAG review and may serve as developer document for an origin trial.

### DIFF
--- a/same-origin-explainer.md
+++ b/same-origin-explainer.md
@@ -108,7 +108,7 @@ The primary advantage of prerendering over prefetching is that users perceive ne
 
 The primary disadvantage is that feature is much more complex for browser implementors, and some pages may need to adjust to detect prerendering by using the `document.prerendering` API if needed.
 
-### <link rel=prerender>
+### `<link rel=prerender>`
 
 An existing API `<link rel=prerender>` is specified today but it is not widely supported. While Chromium is listed as supporting this API, it performs a NoState Prefetch rather than a prerender.
 
--- a/same-origin-explainer.md
+++ b/same-origin-explainer.md
@@ -77,7 +77,7 @@ The user agent may also discard a prerendered page at any time, and some impleme
 
 Generally, prerendered pages can load resources like a normal page. As mentioned earlier, cookies are included on these requests as usual.
 
-However, cross-origin iframes delay loading until after activation. This is partially to avoid breakage caused by loading a cross-origin site that is unaware of prerendering.
+However, loading of cross-origin iframes within the prerendered page is delayed until after activation. This is partially to avoid breakage caused by loading a cross-origin site that is unaware of prerendering.
 
 ### Timing APIs
 
--- a/same-origin-explainer.md
+++ b/same-origin-explainer.md
@@ -1,0 +1,118 @@
+# Same-origin prerendering explainer
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


mfalken:
@domenic @jeremyroman would either or both of you like to review this?

same-origin-explainer.md
```diff
@@ -0,0 +1,119 @@
...
+
+Some restrictions apply to prerendered pages, to prevent intrusive behaviors such as popping up windows, playing audio, etc.
+
+Many APIs are simply deferred in a prerendered page, that is, the processing does not occur until activation. If the API is promise-based, the promise is not resolved until activation. Some examples of these APIs include Geolocation, Web Serial, Notifications, Web MIDI, and Idle Detection.
+
+It is worth noting that APIs gated on user interaction, system focus, or visibilty are implicitly restricted in a prerendered page, since prerendered pages never have those properties. This includes `window.open()`, `element.requestFullscreen()`, and more.
+
+See the [Preventing intrusive behaviors](https://jeremyroman.github.io/alternate-loading-modes/#intrusive-behaviors) section of the specification for a list of restrictions, while this is still a work-in-progress.
+
+The user agent may also discard a prerendered page at any time, and some implementations may do this when they do not yet support deferring or handling an API in a graceful manner. For example, Chromium's implementation discards prerendered pages that use plugins.
```
domenic, line 74: Just curious, by this I guess you mean PDFs? That's the only plugin we support.

same-origin-explainer.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Resource requests
+
+Generally, prerendered pages can load resources like a normal page. As mentioned earlier, cookies are included on these requests as usual.
+
+However, cross-origin iframes delay loading until after activation. This is partially to avoid breakage caused by loading a cross-origin site that is unaware of prerendering.
```
domenic, line 80:
```suggestion
However, loading of cross-origin iframes within the prerendered page is delayed until after activation. This is partially to avoid breakage caused by loading a cross-origin site that is unaware of prerendering.
```

domenic (approved):

domenic:
Also I fixed the main branch so if you rebase or just squash and merge then the CI build should be fixed.

same-origin-explainer.md
mfalken: Hm, I'm not actually sure. Code-wise, we cancel when a "PepperHost" is requested by the renderer process. This code is only reached when the ENABLE_PLUGINS buildflag is enabled. A testcase for this uses `<embed id="testPlugin" type="application/chromium-test-default-plugin" width="42" height="42"></embed>` to cause cancellation.

I guess any "embed/object" that Chromium thinks is a plugin gets here. i'm not sure if the type is special but searching for `chromium-test-default-plugin` in codesearch doesn't bring up anything.

I'll see if I can find a better example later. I'm also thinking of adding an "Implementation-specific quirks" section for any significant Chromium behaviors that might not be intended to be specced.

mfalken (commented):
Thank you! I took your suggestion about iframes.

mfalken:
Thanks... I'll go ahead and merge this.. @jeremyroman or others feel free to add post-merge comments. I expect I'll continue adding to this iteratively.

PR #78, opened 2021-09-03
Add Speculation Rules security and privacy considerations, and refactor into two bikeshed files

jeremyroman:
This change:

* adds security and privacy considerations related to speculation rules
* so that this doesn't get mixed in, `index.bs` is moved to `prerendering.bs`, which is unmodified except for making the speculation rules section just an outbound link
* `speculation-rules.bs` has a new abstract and security and privacy considerations, but is otherwise copied from `index.bs`
* the `Makefile` is modified to support this
* a basic `index.html` is created to bounce people to one of the two output files


domenic:
The split looks solid. I'll try to review the S&P additions on Tuesday.

I guess we should delete `.pr-preview.json` since it doesn't seem to support this use case?

jeremyroman:
That's unfortunate. I guess we could still mark one of the files as supporting it, but ideally tobie/pr-preview#18 would be fixed. :)

speculation-rules.bs
```diff
@@ -0,0 +1,275 @@
...
+
+<h2 id="privacy-considerations">Privacy considerations</h2>
+
+<h3 id="privacy-heuristics">Heuristics</h3>
+
+Because the candidate prefetches and other actions are not required, the user agent can use heuristics to determine which actions would be best to execute. Because it may be observable to the document whether actions were executed, user agents should take care to protect privacy when making such decisions -- for instance by only using information which is already available to the origin. To the extent these heuristics depend on any persistent state, that state should be erased whenever other site data is erased, possibly including time-based expiry.
```
domenic, line 242:
```suggestion
Because the candidate prefetches and other actions are not required, the user agent can use heuristics to determine which actions would be best to execute. Because it may be observable to the document whether actions were executed, user agents should take care to protect privacy when making such decisions — for instance by only using information which is already available to the origin. To the extent these heuristics depend on any persistent state, that state should be erased whenever other site data is erased, possibly including time-based expiry.
```

domenic, line 242: "possibly including time-based expiry" doesn't make too much sense to me. In particular it doesn't seem to be "included" in erasing site data.

Maybe "Additionally, the user agent might want to clear such state after some time" or similar?

domenic, line 242: I think we've settled on a privacy boundary of "site", not "origin".

speculation-rules.bs
```diff
@@ -0,0 +1,275 @@
...
+
+Examples of inputs which would be already known to the document:
+* author-supplied scores
```
domenic, line 245: These aren't in the spec, right? Maybe mention "(in a future version of this specification which allows specifying them)"?

speculation-rules.bs
domenic, line 242: Unsure on "should" vs. "must" here and below... might be better to lean toward "must"?

speculation-rules.bs
```diff
@@ -0,0 +1,275 @@
...
+
+<div class="issue">Expand this section once more detail on prefetch and prerender partitioning mechanism is specified.</div>
+
+<h3 id="privacy-identity-joining">Identity joining</h3>
+
+This specification describes a mechanism through which HTTP requests for later top-level navigation (in the case of prefetching) can be made without a user gesture. It is natural to ask whether it is possible for two coordinating sites to connect user identities.
+
+Since existing [=credentials=] for the destination origin are not sent (assuming it is not [=same origin=] with the referrer), that site is limited in its ability to identify the user before navigation in a similar way to if the referrer site had simply used [[FETCH]] to make an uncredentialed request. Upon navigation, this becomes similar to ordinary navigation (e.g., by clicking a link that was not prefetched).
+
+To the extent that user agents attempt to mitigate identity joining for ordinary fetches and navigations, they can apply similar mitigations to prefetched navigations.
```
domenic, line 275: I think @jyasskin might have a document you could link to on this subject.

domenic (approved):

speculation-rules.bs
jeremyroman: Hmm, how about:

```suggestion
Because the candidate prefetches and other actions are not required, the user agent can use heuristics to determine which actions would be best to execute. Because it may be observable to the document whether actions were executed, user agents should take care to protect privacy when making such decisions -- for instance by only using information which is already available to the origin. If these heuristics depend on any persistent state, that state should be erased whenever the user erases site data or the user agent clears it after some time.
```

speculation-rules.bs
jeremyroman:
```suggestion
* author-supplied scores (if future version of this specification allows specifying them)
```

speculation-rules.bs
jeremyroman: IIUC we have a security boundary of origin and a privacy boundary of site. I think origin is appropriate here because two same-site origins which are not trying to collude are generally isolated on the web, and we don't want user agents to expose information about activity on one origin to another, even if that origin would be free to share that information if it wanted, because it may in fact not want to.

jeremyroman: Is there a general principle here? A UA that didn't follow these principles would still work and be interoperable. Assuming the user was comfortable with making those tradeoffs (even if we don't think they should be the norm) it doesn't seem to be a problem. But if there's a precedent for using "must" in such cases then I'm not averse to changing.

domenic: "that state should be erased whenever ... the user agent clears it after some time" seems tautological.

domenic: Agreed, good analysis.

domenic: I think the general principle would be that it's better to have "must" requirements than "should" requirements when trying to protect users from harm.

jeremyroman: Added this as a note since it's not obvious.

jeremyroman: Done.

jeremyroman: Split it into two sentences to try and make it clearer, hopefully. Coarsely this is trying to say:
- other site data is cleared by user => this site data must be cleared
- other site data is cleared by expiry => this site data must be cleared

domenic: Looks good!

PR #81, opened 2021-09-15
Sketch out prefetching

jeremyroman:
I've tried to sketch out (and mark TODOs on some obvious things that need further work) how we could spec prefetching, both a "partitioned" version suitable for same-site URLs (where credentials apply as long as no redirect leaves the network partition) and an "uncredentialed" version which attempts to describe the isolation from the ordinary partition to not send any prior credentials (though this is a little rough in part because the underlying pieces are still evolving to deal with partitioning).

There are also some bits where I haven't yet investigated the decisions behind what Chromium currently does for private prefetching and so some reconciliation would be needed.

Broadly this is "copy 'process a navigate fetch'" twice and tweak, since we're focused on pre-navigational fetches at the moment.


jeremyroman:
Ninja preview: https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jeremyroman/alternate-loading-modes/prefetch-spec/prefetch.bs

prefetch.bs
```diff
@@ -0,0 +1,219 @@
...
+        :  [=request/use-URL-credentials flag=]
+        :: (set)
+        :  [=request/client=]
+        :: |document|'s [=relevant settings object=]
+        :  [=request/header list=]
+        ::
+            *  `` `Sec-Purpose` ``/`` `Prefetch` ``
+    1. Let |locationURL| be null.
+    1. Let |currentURL| be |request|'s [=request/current URL=].
+    1. While true:
```
domenic, line 104: Oof, duplicating all the navigation-redirect-handling logic here is unfortunate, especially since we keep finding bugs in it and fixing it. (E.g. https://github.com/whatwg/html/commit/431e8584c95d4a5483cb4ed592e65c668d701984)

I'd suggest either trying to create a refactoring so it can be shared (i.e. spec a shared algorithm, then spec the modifications to process a navigate fetch to use your shared algorithm), or at least have a TODO to deduplicate.

prefetch.bs
```diff
@@ -0,0 +1,219 @@
...
+        1. TODO: this is where [=process a navigate fetch=] checks the `navigate-to` CSP
+        1. If |response| is null, [=fetch=] |request|.
+        1. Otherwise, perform [=HTTP-redirect fetch=] using |request| and |response|.
+        1. Wait for the [=task=] on the [=networking task source=] to [=process response=] and set |response| to the result.
+        1. TODO: Figure out if the coop enforcement logic is necessary here; it probably is
+        1. TODO: Figure out if the corp enforcement logic is necessary here; it probably is
+        1. Set |locationURL| to |response|'s [=response/location URL=].
+        1. If |locationURL| is not a [=URL=] whose [=url/scheme=] is an [=HTTP(S) scheme=], then [=iteration/break=].
+    1. If |locationURL| is failure or a [=URL=] whose [=url/scheme=] is not an [=HTTP(S) scheme=], then set |response| to a [=network error=].
+    1. TODO: navigate-to, frame-src, XFO enforcement should probably be left to navigation, but what about status codes and Content-Disposition?
```
domenic, line 129: Would it be observable either way? I guess maybe if a server was going to give a 204 the first time (during prefetch) and a 200 the second time (during navigation), then caching the 204 in the prefetch buffer would be a bit suboptimal?

I mean, do we even want to cache 404s in the prefetch buffer?

prefetch.bs
```diff
@@ -0,0 +1,219 @@
...
+    1. If |response| is a [=network error=], then return.
+    1. [=Store a prefetch record=] given |document|, |url|, |referrerPolicy| and |response|.
+</div>
+
+The <dfn>list of sufficiently strict speculative navigation referrer policies</dfn> is a list containing the following: "", "`strict-origin-when-cross-origin`", "`strict-origin`", "`same-origin`", "`no-referrer`".
+
+<div algorithm="uncredentialed prefetch">
+    To <dfn export>uncredentialed prefetch</dfn> given a [=document=] |document|, [=URL=] |url| and [=referrer policy=] |referrerPolicy|, perform the following steps.
```
domenic, line 137: Any way to share more logic between this and the previous, so that it's obvious what the differences are?

domenic (commented):
Very cool to see some work starting on this.

prefetch.bs
jeremyroman: Would have to poke at what classic prefetch does. At the moment the /prefetch_proxy/ implementation requires a 2XX status code and text/html mime type. It requires that SimpleURLLoader successfully produce a body, but it's not obvious to me without trying it whether a 204/205 results in an empty or absent body string in that code path.

My gut is that we probably do want to require an OK status and possibly just 200. OTOH, I don't see as much reason to require `text/html`, but Ryan would know whether there was a concrete reason behind that.

Non-HTML responses make `Content-Disposition` a little more interesting even though it's not really the marquee use case. I suspect actually handing off to the Chromium download manager would be a mild implementation headache, but I don't really see a conceptual reason it shouldn't be legal.

prefetch.bs
jeremyroman: I can spend some more time trying to. I'm trying to keep it vaguely readable as opposed to "like process a navigate fetch, but with tons of diffs".

I'm worried that if I try to abstract to deduplicate it may harm readability further, but it's worth a shot.

prefetch.bs
jeremyroman: Maybe; I can make another attempt. I'm a little worried that if it becomes too littered with "if you're in mode X, do Y" checks then it will become harder to reason about the properties of each.

prefetch.bs
```diff
@@ -0,0 +1,219 @@
...
+
+A <dfn>prefetch record</dfn> is a [=struct=] with the following [=struct/items=]:
+* <dfn export for="prefetch record">URL</dfn>, a [=URL=]
+* <dfn export for="prefetch record">referrer policy</dfn>, a [=referrer policy=]
+* <dfn export for="prefetch record">response</dfn>, a [=response=]
+* <dfn export for="prefetch record">expiry time</dfn>, a {{DOMHighResTimeStamp}}
+
+<p class="issue">Does the prefetch record need to contain redirect info, e.g. to verify that all redirects would pass `navigate-to`?</p>
+
+The user agent may remove elements from the [=prefetch buffer=] even if they are not expired, e.g., due to resource constraints.
```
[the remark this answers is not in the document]

prefetch.bs
```diff
@@ -0,0 +1,219 @@
...
+  type: dfn
+    text: process response; url: process-response
+    text: network partition key; url: network-partition-key
+</pre>
+
+<h2 id="concepts">Concepts</h2>
+
+In light of <a href="https://privacycg.github.io/storage-partitioning/">storage partitioning</a>, this specification defines prefetch for navigations which would occur within the same partition (for example, top-level navigations within the same site) and for navigations which would occur in a separate partition (for example, top-level navigations to a different site).
+
+Each [=document=] has a <dfn export>prefetch buffer</dfn>, which is a [=list=] of [=prefetch records=].
```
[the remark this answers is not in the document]
jeremyroman: Done globally.

prefetch.bs
domenic: I think OK status that is not 204 or 205 is probably the right call. Content type... hmm. I mean, I like enforcing good content type hygiene in general, but we should at least support `application/xhtml+xml` or whatever the XHTML content type is.

I think avoiding downloads or non-HTML pages is reasonable just as a simplification effort, since the benefits of prerendering them are pretty marginal.

prefetch.bs
domenic: Yeah, very fair concern. It might be that just getting the redirect stuff factored out is enough.

prefetch.bs
jeremyroman: Especially in the prefetch case, it doesn't seem unreasonable that a user might want a tax form PDF or something like that to load quickly, for instance.

jeremyroman:
ptal; factored out a "perform a common navigational fetch" which takes algorithms for: "will redirect", "should block navigation request" and "should block navigation response"

prefetch.bs
```diff
@@ -0,0 +1,315 @@
...
+</div>
+
+<h2 id="html-patches">HTML Patches</h2>
+
+<div algorithm="perform a common navigational fetch">
+    <div class="note">This is an abstraction of the existing [=process a navigate fetch=]. It includes behavior like redirect handling that is particular to navigational fetches, including those that are relate to a speculate future navigation, rather than an immediate one.</div>
+
+    To <dfn export>perform a common navigational fetch</dfn> given a [=request=] |request|, [=string=] |navigationType|, [=browsing context=] |browsingContext|, [=environment=] or null |forceEnvironment|, algorithm |willRedirect| (which takes [=URLs=] currentURL and locationURL) algorithm |shouldBlockNavigationRequest| (which takes a [=request=], navigation type [=string=], and [=environment=], and returns "`Blocked`" or "`Allowed`"), algorithm |shouldBlockNavigationResponse| (which takes a [=request=] and a [=response=], and returns "`Blocked`" or "`Allowed`"), perform the following steps.
```
domenic, line 108: "willRedirect" name is confusing; I thought it'd return a boolean. Maybe "processRedirect"?

domenic (approved):
This looks great. Definitely a number of TODOs, but you've noted them all.

prefetch.bs
jeremyroman: Hmm, the calling code still does the redirect processing (at least, I would interpret that to mean the updated URL computation and fetch). I was hoping to follow the common-ish (at least in code) `willDoFoo`/`didDooFoo` pattern for pre/post hooks. But if this is unclear (I can see how it has a boolean sense), how about `preRedirectHook`?

domenic: Yeah, `preRedirectHook` seems good too.

PR #84, opened 2021-10-13
Update text around Sec-Purpose header.

jeremyroman:
Correct the case on `prefetch`, and expressly suggest browsers may continue sending the vendor-specific equivalents for now.


toyoshim:
LGTM. Thank you for working on this!

PR #86, opened 2021-10-21
Explain why keyboard lock doesn't require prerendering integration.

### LINKED ISSUE #82: Refer Keyboard Lock API in Prerender Revamp section 7.4
https://wicg.github.io/keyboard-lock/#h-keyboard-lock
navigator.keyboard.lock() requires active top-level browsing context.

So, we would list this API in the "APIs that require transient activation or sticky activation:" block in the section 7.4.

jeremyroman:
Resolves #82.

@toyoshim @domenic ptal


jeremyroman:
Do you think it's worth noting? The same is true of a non-prerendered browsing context -- you can get keyboard lock but it won't have any effect until you go fullscreen. (This took me awhile to figure out.)

prerendering.bs
```diff
@@ -734,6 +734,9 @@ APIs that require [=transient activation=] or [=sticky activation=]:
 - {{Window/open(url, target, features)|window.open()}} [[HTML]]
 - The prompt generated by the {{Window/beforeunload}} event [[HTML]]
 - {{Element/requestFullscreen()|element.requestFullscreen()}} [[FULLSCREEN]]
+  - {{Keyboard/lock()|navigator.keyboard.lock()}} [[KEYBOARD-LOCK]], which requires fullscreen
+
+    <div class="note">[[KEYBOARD-LOCK]] allows browsing contexts to enable keyboard lock easily but it has no effect except in fullscreen, which requires a user gesture. If that changes then this should be revisited.</div>
```
domenic, line 739: should -> needs to be to avoid using special keywords

jeremyroman: Bah, I ran the build the first time but thought I could get away with amending the note without messing this up. Fixed.

toyoshim:
sorry for checking after the merge. it looks good to me.
thank you for the update!

jeremyroman:
Filed https://bugs.chromium.org/p/chromium/issues/detail?id=1268179 for WPT

PR #89, opened 2021-11-03
Update links and spec boilerplate to reflect the repository move into WICG.

jeremyroman:


domfarolino (approved):
Other Dominic but this popped up in my inbox and LGTM

fetch.md
```diff
@@ -238,8 +238,8 @@ proxy server or virtual private network, if available)
 [1]: https://html.spec.whatwg.org/multipage/webappapis.html#concept-environment-top-level-origin
 [2]: https://fetch.spec.whatwg.org/#network-partition-key
 [3]: https://fetch.spec.whatwg.org/#concept-request-cache-mode
-[4]: https://htmlpreview.github.io/?https://raw.githubusercontent.com/jeremyroman/alternate-loading-modes/spec/index.html#prerendering-browsing-context
+[4]: https://htmlpreview.github.io/?https://raw.githubusercontent.com/WICG/nav-speculation/spec/index.html#prerendering-browsing-context
```
domenic, line 241: This link seems subpar and maybe broken; should we fix it while here?

domenic (approved):
LGTM as a strict improvement, although I suspect there may have been some broken links due to the split of the specs that we could fix at the same time.

fetch.md
jeremyroman:
```suggestion
[4]: https://wicg.github.io/nav-speculation/prerendering.html#prerendering-browsing-context
```

jeremyroman:
Fixed that one and a couple more.

PR #93, opened 2021-11-25
Lift CORP checks to "perform a common navigational fetch".

jeremyroman:
Though they only apply to child browsing contexts, it seems appropriate
to do these at prefetch time as long as the prefetch buffer is scoped to
the document that requested the prefetch (otherwise, the CORP checks
would occur against the wrong policy container).

In particular, this means that it is not possible to, inside a frame
which has Cross-Origin-Embedder-Policy: require-corp, navigationally
prefetch a document (or redirect to a document) which does not have a
compatible Cross-Origin-Resource-Policy header.

This has no effect in top-level browsing contexts, but does apply to
both partitioned (N.B., due to the network partition key, this is not
actually cross-partition even if it is cross-origin) and uncredentialed
prefetch -- if either is useful in child browsing contexts, anyway.


jeremyroman:
@camillelamy FYI (and checking that I'm not grossly misunderstanding COEP/CORP)

jeremyroman:
Build issue should be fixed by #94.

camillelamy:
Ok so I am not very familiar with the spec you are modifying. If I read the change correctly, it means that we will also apply COEP+CORP checks to subframes in the page being prerendered. Without this change, we would not do it. Is that correct? If so, yes we should definitely do that.

jeremyroman:
This applies to prefetching (when prerendering doesn't take place). Rather, it's intended to specify that CORP checks should apply to prefetches for navigation in the same way they would apply to a non-prefetched navigation in the same frame.

camillelamy:
Ok. Well that should be the case.

jeremyroman:
bump @domenic

prefetch.bs
```diff
@@ -141,6 +141,9 @@ The user agent may remove elements from the [=prefetch buffer=] even if they are
...
         1. Wait for the [=task=] on the [=networking task source=] to [=process response=] and set |response| to the result.
         1. If the result of |shouldBlockNavigationResponse| given |request| and |response| is "`Blocked`", then set |response| to a [=network error=] and [=iteration/break=].
+        1. If |response| is not a [=network error=], |browsingContext| is a [=child browsing context=], and the result of performing a [=cross-origin resource policy check=] with |browsingContext|'s [=browsing context/container document=]'s [=Document/origin=], |browsingContext|'s [=browsing context/container document=]'s [=relevant settings object=], |request|'s [=request/destination=], |response| and true is <strong>blocked</strong>, then set |response| to a [=network error=] and [=iteration/break=].
+
+            <div class="note">Here we're running the [=cross-origin resource policy check=] against the [=parent browsing context=] rather than sourceBrowsingContext. This is because we care about the same-originness of the embedded content against the parent context, not the navigation source.</div>
```
domenic, line 146:
```suggestion
            <div class="note">Here we're running the [=cross-origin resource policy check=] against the [=parent browsing context=] rather than |sourceBrowsingContext|. This is because we care about the same-originness of the embedded content against the parent context, not the navigation source.</div>
```

prefetch.bs
```diff
@@ -141,6 +141,9 @@ The user agent may remove elements from the [=prefetch buffer=] even if they are
         1. Otherwise, perform [=HTTP-redirect fetch=] using |request| and |response|.
         1. Wait for the [=task=] on the [=networking task source=] to [=process response=] and set |response| to the result.
         1. If the result of |shouldBlockNavigationResponse| given |request| and |response| is "`Blocked`", then set |response| to a [=network error=] and [=iteration/break=].
+        1. If |response| is not a [=network error=], |browsingContext| is a [=child browsing context=], and the result of performing a [=cross-origin resource policy check=] with |browsingContext|'s [=browsing context/container document=]'s [=Document/origin=], |browsingContext|'s [=browsing context/container document=]'s [=relevant settings object=], |request|'s [=request/destination=], |response| and true is <strong>blocked</strong>, then set |response| to a [=network error=] and [=iteration/break=].
```
domenic, line 144:
```suggestion
        1. If |response| is not a [=network error=], |browsingContext| is a [=child browsing context=], and the result of performing a [=cross-origin resource policy check=] with |browsingContext|'s [=browsing context/container document=]'s [=Document/origin=], |browsingContext|'s [=browsing context/container document=]'s [=relevant settings object=], |request|'s [=request/destination=], |response|, and true is <strong>blocked</strong>, then set |response| to a [=network error=] and [=iteration/break=].
```

domenic (approved):
LGTM with nits; sorry for the delay. Took me a while to think through what was going on here, but the commit message was quite helpful.

prefetch.bs
jeremyroman: Is there an official position on the Oxford comma? :) I don't mind so I'll just commit it.

PR #97, opened 2021-11-30
Disallow prerendered page activation in fenced frames

horo-t:
Fixes WICG/nav-speculation#96


domfarolino (commented):
Could we also add an "issue" tag perhaps like the following:

`<p class="issue">The concept of a fenced frame browsing context is not yet defined but we need to link to it once it exists.</p>`

horo-t:
> Could we also add an "issue" tag perhaps like the following:
> 
> `<p class="issue">The concept of a fenced frame browsing context is not yet defined but we need to link to it once it exists.</p>`

Done.
Thank you.

domfarolino (approved):

PR #99, opened 2021-12-09
Explain client IP anonymization and how it relates to nav speculation.

jeremyroman:
This is a key capability that isn't well-explained in the explainer yet and deserves some expansion.

@kenjibaheux @buettner fyi


domenic (approved):
Very nice

PR #107, opened 2021-12-16
Fix visibility references to ref HTML spec

noamr:


prerendering.bs
```diff
@@ -45,16 +45,13 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: traverse the history; url: traverse-the-history
     urlPrefix: interaction.html
       text: system focus; url: tlbc-system-focus
+      text: visibility state; for: Document; url: dom-document-visibilitystate
+      text: hidden; for: Document; url: dom-document-hidden
```
domenic, line 49: IIUC, this will make https://wicg.github.io/nav-speculation/prerendering.html#interaction-with-visibility-state link to the `document.hidden` DOM property. Instead I think we should have it state something like `... always have a visibility state of "hidden"`.

prerendering.bs
```diff
@@ -45,16 +45,13 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: traverse the history; url: traverse-the-history
     urlPrefix: interaction.html
       text: system focus; url: tlbc-system-focus
+      text: visibility state; for: Document; url: dom-document-visibilitystate
```
domenic, line 48: This should be exported, so this line shouldn't be necessary?

domenic (approved):

PR #109, opened 2021-12-17
Add autoplay patch

noamr:
When media is playing, it should still appear to be playing in the context of the prerendered page.
However, the monotonic time progression of the media element would be stalled until the page is active.

This is different from the behavior of opening the page in a new tab, as in the case of new tabs the video starts running﻿
while in the case of a prerendered page you don't want prerendering to change the behavior of media elements in that way.


jeremyroman (approved):

noamr:
Thanks! Should I have merge rights on this repo, or do you want to click merge @jeremyroman? :)

PR #110, opened 2021-12-20
A batch for delaying permission-gated features

noamr:
﻿Added a delay-while-prerender gate to the following permission-gated features:

1. Media capture (camera/microphone)
2. Audio output (speaker selection)
3. Push API
4. Background Fetch
5. Persistent Storage
6. Screen Capture
7. WebUSB
8. WebBluetooth
9. WebHId
10. WebXR
11. Credential Management (which includes WebAuth public key management)

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -131,6 +131,33 @@ spec: navigation-timing; urlPrefix: https://w3c.github.io/navigation-timing/
 spec: encrypted-media; urlPrefix: https://w3c.github.io/encrypted-media/
   type: method; for: Navigator
     text: requestMediaKeySystemAccess(keySystem, supportedConfigurations); url: requestMediaKeySystemAccess
+spec: mediacapture-main; urlPrefix: https://w3c.github.io/mediacapture-main/
+  type: dfn
+    text: device change notification steps; url: dfn-device-change-notification-steps
+spec: hid; urlPrefix: https://wicg.github.io/webhid/
+  type: interface
+    text: HID; url: hid-interface
+  type: method; for: HID
+    text: getDevices(); url: dom-hid-getdevices
+  type: method; for: HID
+    text: requestDevice(); url: dom-hid-requestdevice
+spec: backgroud-fetch; urlPrefix: https://wicg.github.io/background-fetch/
+  type: interface
+    text: BackgroundFetchManager; url: backgroundfetchmanager
+  type: method; for: BackgroundFetchManager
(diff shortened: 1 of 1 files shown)


noamr:
Should allow ticking some boxes at https://github.com/WICG/nav-speculation/issues/42

prerendering.bs
```diff
@@ -721,6 +748,56 @@ Add {{[DelayWhilePrerendering]}} to {{ScreenOrientation/unlock()}}.
 
 Add {{[DelayWhilePrerendering]}} to {{Navigator/requestMediaKeySystemAccess()}}.
 
+<h4 id="media-capture-patch">Media Capture and Streams</h4>
+
+Add {{[DelayWhilePrerendering]}} to {{Navigator/getUserMedia()}}, {{MediaDevices/getUserMedia()}} and {{MediaDevices/enumerateDevices()}}.
```
jeremyroman, line 753: I don't think this works for Navigator/getUserMedia(), since its API is callback-based rather than promise-based. (Well, it would work except that it would return a promise which it shouldn't.)

I'd suggest modifying the definition of `[DelayWhilePrerendering]` to only return the promise if the operation otherwise returns a promise, and otherwise to require that the operation returns `undefined`. (Or perhaps you have another alternative.)

prerendering.bs
```diff
@@ -721,6 +748,56 @@ Add {{[DelayWhilePrerendering]}} to {{ScreenOrientation/unlock()}}.
...
+
+<h4 id="webhid-patch">WebHID API</h4>
+
+Add {{[DelayWhilePrerendering]}} to {{HID/getDevices()}} and {{HID/requestDevice()}}.
+
+<h4 id="webxr-patch">WebXR Device API</h4>
+
+Add {{[DelayWhilePrerendering]}} to {{XRSystem/requestSession()}}.
+
+<h4 id="credential-manamgenet-patch">Credential Manamgenet</h4>
```
jeremyroman, line 797:
```suggestion
<h4 id="credential-management-patch">Credential Management</h4>
```

prerendering.bs
noamr: Yea, I think that would be a good way to go with `DelayWhilePrerendering`.

noamr: @jeremyroman: changed `DelayWhilePrerendering` in new revision to also work for void functions.

jeremyroman (approved):

PR #112, opened 2021-12-20
Patch service worker notifications

noamr:
Only allow showing a notification if the SW has at least one non-prerendered client.
See https://github.com/WICG/nav-speculation/issues/44

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -642,7 +642,17 @@ Add {{[DelayWhilePrerendering]}} to {{Notification/requestPermission()}}.
   1. Otherwise, <a spec="NOTIFICATIONS">get the notifications permission state</a> and return it.
 </div>
 
-TODO: what about the service worker API? Depends on what we're doing for service workers in prerendering BCs...
+<div algorithm="Service worker notifications patch">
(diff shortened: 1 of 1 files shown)


jeremyroman (commented):
Hmm, I don't think this works. Previously a ServiceWorker would be able to show a notification when no clients at all existed (e.g., in response to a push notification) but with this patch such notifications would be dropped.

e.g.

```js
// sw.js
self.onpush = e => {
  self.registration.showNotification(`got push ${e.data.text()}`);
};
```

noamr:
> Hmm, I don't think this works. Previously a ServiceWorker would be able to show a notification when no clients at all existed (e.g., in response to a push notification) but with this patch such notifications would be dropped.
> 
> e.g.
> 
> ```js
> // sw.js
> self.onpush = e => {
>   self.registration.showNotification(`got push ${e.data.text()}`);
> };
> ```

You're right.
I'm wondering if this direction is correct at all. Will comment on the issue.

noamr:
Closing for now. See issue, proposing to have the restriction elsewhere.

PR #115, opened 2022-01-11
Trigger prefetch within the speculation rules spec.

jeremyroman:
Now that there's a definition of prefetch to link to, link to it.

The partitioned version is necessary for this to be useful for same-site cases in top-level browsing contexts; the uncredentialed version is used for cross-site cases. This meshes slightly awkwardly with the same-origin concept used in other places.

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -407,5 +407,11 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
     1. If |partitionKey| is equal to (|topLevelSite|, |secondKey|), then [=partitioned prefetch=] given |document|, |url| and |referrerPolicy|.
     1. Otherwise, [=uncredentialed prefetch=] given |document|, |url| and |referrerPolicy|.
 
-    <div class="note">This determines whether the navigation would use the same network partition key. If it would, the prefetch is restricted to the same partition. Otherwise it is uncredentialed.</div>
+    <div class="note">
+      This determines whether the navigation would use the same network partition key.
+
+      If it would, the prefetch is restricted to the same partition, and redirects which would leave the partition cause the prefetch to fail.
+
+      Otherwise it is uncredentialed, and redirects which would return to the original partition (thus ought to have credentials) will cause the prefetch to fail.
+    </div>
 </div>
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -401,7 +401,7 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
     To <dfn export>prefetch</dfn> given a {{Document}} |document|, [=URL=] |url| and [=referrer policy=] |referrerPolicy|, perform the following steps.
 
     1. Let |partitionKey| be the result of [=determining the network partition key=] given |document|'s [=relevant settings object=].
-    1. Let |topLevelOrigin| be |document|'s [=relevant settings object=]'s [=environment/top-level origin=] if |document|'s [=Document/browsing context=] is not a [=top-level browsing context=], and |url|'s [=origin=] otherwise.
+    1. Let |topLevelOrigin| be |url|'s [=url/origin=] if |document|'s [=Document/browsing context=] is a [=top-level browsing context=], and |document|'s [=relevant settings object=]'s [=environment/top-level origin=] otherwise.
     1. Let |topLevelSite| be the result of [=obtaining a site=], given |topLevelOrigin|.
     1. Let |secondKey| be null or an [=implementation-defined=] value.
     1. If |partitionKey| is equal to (|topLevelSite|, |secondKey|), then [=partitioned prefetch=] given |document|, |url| and |referrerPolicy|.
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -396,3 +396,16 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
 
     <div class="issue">Further review is needed of the cookies-present handling and its interaction with redirects and caches.</div>
 </div>
+
+<div algorithm>
+    To <dfn export>prefetch</dfn> given a {{Document}} |document|, [=URL=] |url| and [=referrer policy=] |referrerPolicy|, perform the following steps.
+
+    1. Let |partitionKey| be the result of [=determining the network partition key=] given |document|'s [=relevant settings object=].
+    1. Let |topLevelOrigin| be |document|'s [=relevant settings object=]'s [=environment/top-level origin=] if |document|'s [=Document/browsing context=] is not a [=top-level browsing context=], and |url|'s [=origin=] otherwise.
+    1. Let |topLevelSite| be the result of [=obtaining a site=], given |topLevelOrigin|.
+    1. Let |secondKey| be null or an [=implementation-defined=] value.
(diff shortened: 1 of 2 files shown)


speculation-rules.bs
```diff
@@ -23,6 +23,9 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
     urlPrefix: webappapis.html
       text: script; url: concept-script
+spec: nav-speculation; urlPrefix: prefetch.html
```
domenic, line 26: Hmm, probably this indicates the linking database needs updating to have two separate specs (speculation-rules and prefetch) instead of one (nav-speculation). Might be worth filing an issue on tabatkins/bikeshed for that.

speculation-rules.bs
```diff
@@ -204,8 +207,7 @@ Periodically, for any [=document=] |document|, the user agent may [=queue a glob
     1. [=list/For each=] |rule| of |ruleSet|'s [=speculation rule set/prefetch rules=]:
       1. Let |requiresAnonymousClientIPWhenCrossOrigin| be true if |rule|'s [=speculation rule/requirements=] [=set/contains=] "`anonymous-client-ip-when-cross-origin`", and false otherwise.
       1. [=list/For each=] |url| of |rule|'s [=speculation rule/URLs=]:
-        1. The user agent may prefetch |url| given |requiresAnonymousClientIPWhenCrossOrigin|.
-           <p class="issue">TODO: expand this to actually elaborate on how prefetch works, once initiated, and to incorporate the |requiresAnonymousClientIPWhenCrossOrigin| flag. We may wish to include language about when the UA should deduplicate requests.
+        1. The user agent may [=prefetch=] |url| given |document|, |url| and "`strict-origin-when-cross-origin`". If |requiresAnonymousClientIPWhenCrossOrigin| is true, it must not do so unless it can obscure the user's client IP address for any request for a URL whose [=url/origin=] is not [=same origin=] to |document|'s [=Document/origin=].
```
domenic, line 210: I would think that long term we would want to flow this flag down into https://fetch.spec.whatwg.org/#http-network-or-cache-fetch , probably step 9.5. So leaving a TODO might be a good idea, or maybe passing the boolean to prefetch and moving the TODO there.

prefetch.bs
```diff
@@ -396,3 +396,16 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
 
     <div class="issue">Further review is needed of the cookies-present handling and its interaction with redirects and caches.</div>
 </div>
+
+<div algorithm>
+    To <dfn export>prefetch</dfn> given a {{Document}} |document|, [=URL=] |url| and [=referrer policy=] |referrerPolicy|, perform the following steps.
+
+    1. Let |partitionKey| be the result of [=determining the network partition key=] given |document|'s [=relevant settings object=].
+    1. Let |topLevelOrigin| be |document|'s [=relevant settings object=]'s [=environment/top-level origin=] if |document|'s [=Document/browsing context=] is not a [=top-level browsing context=], and |url|'s [=origin=] otherwise.
```
domenic, line 404: This line is a bit confusing because of the negative. Maybe

> Let topLevelOrigin be url's origin if document's browsing context is a top-level browsing context; otherwise let it be document's relevant settings object's origin.

With that formulation I think I can see the logic more clearly: if we're navigating a TLBC, the _new_ top-level origin will be related to url, whereas if we're navigating a nested BC, the top-level origin will stay the same.

prefetch.bs
```diff
@@ -396,3 +396,16 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
...
+    1. Let |topLevelSite| be the result of [=obtaining a site=], given |topLevelOrigin|.
+    1. Let |secondKey| be null or an [=implementation-defined=] value.
+    1. If |partitionKey| is equal to (|topLevelSite|, |secondKey|), then [=partitioned prefetch=] given |document|, |url| and |referrerPolicy|.
+    1. Otherwise, [=uncredentialed prefetch=] given |document|, |url| and |referrerPolicy|.
+
+    <div class="note">This determines whether the navigation would use the same network partition key. If it would, the prefetch is restricted to the same partition. Otherwise it is uncredentialed.</div>
```
domenic, line 410: So in the redirect case the resulting partition key might not end up the same, since |url| is not the same as the final URL. Is that an issue?

speculation-rules.bs
jeremyroman: I'm not sure I follow. It's at least as likely that I don't know what I'm doing when I want to link between two .bs files in one repository. This seemed to work, but I may be holding it wrong.

speculation-rules.bs
jeremyroman: Hmm, all the way to https://fetch.spec.whatwg.org/#concept-connection-obtain? Even if we do plumb it there in the future it will probably remain a handwave there until/unless the mechanism for anonymizing the client IP becomes universal (e.g., everyone implements MASQUE proxies or whatever). For the time being this being implementation-defined seems appropriate.

Still, happy to leave a TODO here.

prefetch.bs

prefetch.bs
jeremyroman: This is certainly an awkward case. At the moment the Chromium new-prefetch path doesn't allow this at all; what I've suggested in spec text is the following resolution (open to better ones):

- if it starts with a URL in the same partition, the "partitioned prefetch" algorithm is used which allows following redirects as long as they stay within the partition, and bails otherwise (at the top level, this basically means same-site)
- if it starts with a URL in another partition, the "uncredentialed prefetch" algorithm is used which allows redirects as long as they remain outside the requesting page's partition (i.e., they remain cross-site) and isolates the storage

This obviously doesn't deal with redirect chains which include both same- and cross-site URLs, but it doesn't preclude adding specific cases of that which are safe at a later point.

speculation-rules.bs
domenic: So in general you should only need to add things to the `<pre class="anchors">` block if they're not in the linking database, either because they're not exported, or because they're not in the linking database.

I was hypothesizing that the linking database only contains https://wicg.github.io/nav-speculation/, and might not contain https://wicg.github.io/nav-speculation/prefetch.html .

However perhaps the issue is just that the linking database doesn't get updated until after the PR is merged. So this is definitely necessary for now. The question is whether, ~24 hours after merging, you can remove it and just use `[=prefetch=]` with nothing in `<pre class=anchors>`.

speculation-rules.bs
domenic: Yeah. Making it clear that it normatively impacts the connection-obtaining part and nothing else would be ideal. I wouldn't expect anything more than a handwave there.

Right now it's essentially monkey-patching "obtain a connection" from a distance, which is not great.

In other words I think it's fine for there to be implementation-defined behavior at that level, we should just avoid the monkeypatching approach.

Anyway, the issue text you have below looks good to me for now.

prefetch.bs
domenic: Oh, that sounds great. Maybe just add that explanation in a note here, since it's pretty easy to say "wait, |url| doesn't look like the right way to choose a partition key", without realizing that if you follow all the algorithms you'll end up in a safe place.

domenic (approved):

PR #116, opened 2022-01-11
Remove "prefetch_with_subresources" from speculation rules.

jeremyroman:
It's somewhat complicated to specify, no browser engine that I'm aware of has near-term plans to ship it, and there isn't much spec text for it yet anyway.


domenic (approved):
Hmm, I thought we were already shipping this in some form? I guess not speculation-rules driven; maybe omnibox?

jeremyroman:
Yes, NoStatePrefetch is shipping with other triggers. But at the moment the mentions of it in WICG/nav-speculation are a stub and since we don't have plans to spec this in the short term, we can always add those stubs back when they become useful.

PR #117, opened 2022-01-18
Outline the Sec-Purpose HTTP request header field.

jeremyroman:
Preview: https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jeremyroman/nav-speculation/sec-purpose/prefetch.bs#sec-purpose-header

This actually defines the `Sec-Purpose` header and describes how it can have a parameter which indicates that the prefetch has been IP-anonymized. Plumbing is still required to populate that parameter, but this seems a useful intermediate state to confirm that this shape makes sense.

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -61,6 +61,11 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
   type: dfn
     text: process response; url: process-response
     text: network partition key; url: network-partition-key
+spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
+  type: dfn
+    text: Item; url: name-items
+    text: List; url: name-lists
+    text: Token; url: name-tokens
 </pre>
 
 <h2 id="concepts">Concepts</h2>
@@ -290,6 +295,8 @@ These algorithms are based on [=process a navigate fetch=].
                 <div class="note">
                     Implementations might also send vendor-specific headers, like Chromium's `` `Purpose` ``/`` `prefetch` ``, Mozilla's `` `X-moz` ``/`` `prefetch` ``, and WebKit's `` `X-Purpose` ``/`` `preview` ``, for compatibility with existing server software. Over time we hope implementers and server software authors will adopt a standard header.
                 </div>
+
+                <div class="issue">TODO: Emit the "`anonymous-client-ip`" parameter when applicable.</div>
     1. Let |redirectChain| be an empty [=list=].
     1. Let |shouldBlockNavigationRequest| be the following steps, given [=request=] |request|, [=string=] |navigationType| and [=environment=] |environment|:
         1. [=Assert=]: |navigationType| is "`other`".
@@ -358,6 +365,8 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
                     Implementations might also send vendor-specific headers, like Chromium's `` `Purpose` ``/`` `prefetch` ``, Mozilla's `` `X-moz` ``/`` `prefetch` ``, and WebKit's `` `X-Purpose` ``/`` `preview` ``, for compatibility with existing server software. Over time we hope implementers and server software authors will adopt a standard header.
                 </div>
 
+                <div class="issue">TODO: Emit the "`anonymous-client-ip`" parameter when applicable.</div>
+
     1. Let |originsWithConflictingCredentials| be an empty [=ordered set=].
     1. Let |redirectChain| be an empty [=list=].
     1. Let |shouldBlockNavigationRequest| be the following steps, given [=request=] |request|, [=string=] |navigationType| and [=environment=] |environment|:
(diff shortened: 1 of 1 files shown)


prefetch.bs
```diff
@@ -415,3 +424,40 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
       Otherwise it is uncredentialed, and redirects which would return to the original partition (thus ought to have credentials) will cause the prefetch to fail.
     </div>
 </div>
+
+<h2 id="sec-purpose-header">The `Sec-Purpose` HTTP request header</h2>
+
+The `` `Sec-Purpose` `` HTTP request header specifies that the request serves one or more purposes other than requesting the resource for immediate use by the user.
```
domenic, line 430: Add a `<dfn http-header>`?

prefetch.bs
```diff
@@ -415,3 +424,40 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
...
+
+It may contain a <a spec="RFC8941">Item</a> member which is the <a spec="RFC8941">Token</a> "`prefetch`". If so, this indicates the request's purpose is to download a resource it is anticipated will be fetched shortly.
+
+<div class="issue">TODO: Are there normative implications of this that should be specified here?</div>
+
+The following parameters is defined for the "`prefetch`" member:
+
+* A parameter whose key is "`anonymous-client-ip`".
+
+  If present with a value other than boolean false, this parameter indicates that the prefetch request is being made using an anonymous client IP. Consequently, servers should not rely on it matching, or sharing a geographic location or network operator with, the client's IP address from which a non-prefetch request would have been made.
```
domenic, line 446: I always like to clarify booleans with structured headers, e.g. something like "value other than the structured header boolean false (i.e. `?0`)..."

prefetch.bs
```diff
@@ -415,3 +424,40 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
...
+
+  If a suitable response is not possible, for example because the resource depends on the client's geographic location, there is no other means of determining the location (e.g., the <a href="https://www.ietf.org/archive/id/draft-geohash-hint-00.html">Geohash client hint</a>), and no location-agnostic response is available, then the server should respond with an appropriate HTTP status code and response headers which mark the response as not suitable for caching.
+
+  <div class="note">
+    A future specification might define assign more specific meaning to non-boolean values. For now, they are treated the same as true. Implementations are advised not to emit such values.
+  </div>
+
+<h2 id="iana-considerations">IANA Considerations</h2>
+
+This document registers the "`Sec-Purpose`" header in the <a href="https://www.iana.org/assignments/http-fields/http-fields.xhtml">Hypertext Transfer Protocol (HTTP) Field Name Registry</a>:
```
domenic, line 456: We could maybe skip straight to https://github.com/protocol-registries/http-fields these days. Recently HTML and Fetch have removed all their header registration appendices in favor of that.

prefetch.bs
```diff
@@ -415,3 +424,40 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
...
+
+<div class="issue">TODO: Are there normative implications of this that should be specified here?</div>
+
+The following parameters is defined for the "`prefetch`" member:
+
+* A parameter whose key is "`anonymous-client-ip`".
+
+  If present with a value other than boolean false, this parameter indicates that the prefetch request is being made using an anonymous client IP. Consequently, servers should not rely on it matching, or sharing a geographic location or network operator with, the client's IP address from which a non-prefetch request would have been made.
+
+  If a suitable response is not possible, for example because the resource depends on the client's geographic location, there is no other means of determining the location (e.g., the <a href="https://www.ietf.org/archive/id/draft-geohash-hint-00.html">Geohash client hint</a>), and no location-agnostic response is available, then the server should respond with an appropriate HTTP status code and response headers which mark the response as not suitable for caching.
```
domenic, line 448: Do we have any idea what the appropriate status code would be? I could imagine 400, 403, 406.

On the other hand, maybe leaving this vague is better since it would avoid anyone depending on specific responses in some way.

prefetch.bs
```diff
@@ -415,3 +424,40 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
...
+
+```
+Sec-Purpose = sf-list
+```
+
+It may contain a <a spec="RFC8941">Item</a> member which is the <a spec="RFC8941">Token</a> "`prefetch`". If so, this indicates the request's purpose is to download a resource it is anticipated will be fetched shortly.
+
+<div class="issue">TODO: Are there normative implications of this that should be specified here?</div>
+
+The following parameters is defined for the "`prefetch`" member:
```
domenic, line 442: member -> token?

prefetch.bs
```diff
@@ -415,3 +424,40 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
...
+
+* A parameter whose key is "`anonymous-client-ip`".
+
+  If present with a value other than boolean false, this parameter indicates that the prefetch request is being made using an anonymous client IP. Consequently, servers should not rely on it matching, or sharing a geographic location or network operator with, the client's IP address from which a non-prefetch request would have been made.
+
+  If a suitable response is not possible, for example because the resource depends on the client's geographic location, there is no other means of determining the location (e.g., the <a href="https://www.ietf.org/archive/id/draft-geohash-hint-00.html">Geohash client hint</a>), and no location-agnostic response is available, then the server should respond with an appropriate HTTP status code and response headers which mark the response as not suitable for caching.
+
+  <div class="note">
+    A future specification might define assign more specific meaning to non-boolean values. For now, they are treated the same as true. Implementations are advised not to emit such values.
```
domenic, line 451: Maybe either now or in the future mention something like "(The [=partitioned prefetch=] and [=uncredentialed prefetch=] algorithms conform to this advice.)"

domenic (approved):
LGTM with minor suggestions

prefetch.bs
jeremyroman: Done.

prefetch.bs
jeremyroman: This scans weirdly to me because RFC 8941 uses the term "boolean false" to refer to the structured data value, and `?0` is the serialization of that (admittedly the only one, disregarding whitespace). Added a variant of that nonetheless.

prefetch.bs
jeremyroman: It seems slightly odd to mention it before actually making it true. :)

In the future we could if I remember to do so, though it kinda just says "By the way, we follow our own advice." so I'm uncertain how much value it adds.

prefetch.bs
jeremyroman: Yeah. I could also see 500, maybe. I don't want to overspecify and also we're kinda discussing in parallel whether prefetch discarding the response should be keyed off the status code or the cache control headers. (Presently at least Chromium actually does treat 4XX/5XX responses as potentially usable for prefetch.) If that settles down and the resolution seems interoperable I'm happy to narrow this.

prefetch.bs
jeremyroman: So the modern thing to do is to not include the registration appendix at all, and merely to file a GitHub issue there at an appropriate time? (When is an appropriate time for "wide review"? When we've resolved a couple of the TODOs added in this PR? Immediately upon this landing?)

prefetch.bs
domenic: Not done?

prefetch.bs
domenic: I feel like it clarifies things a bit because upon first reading I'm like "wait, why are you giving such weaksauce advice in a non-normative note; don't you control the actual specification for when this header is sent?" Such an expanded note makes it clearer you're talking about other potential implementers besides prefetch.

prefetch.bs
domenic: I'm not really sure what the protocol is for new headers like this. I think it hasn't really been tried; we've only converted old established headers. So I'd say it's up to you to attempt registration whenever you feel it's worthwhile.

domenic (approved):
LGTM with member -> token

prefetch.bs
jeremyroman: Let's try this again. Done. Hopefully.

prefetch.bs
jeremyroman: Done.

prefetch.bs
jeremyroman: Alright, filed #118 to do this in the future and removed the IANA considerations section from here for now.

prefetch.bs
noamr: I think that for the purpose of prerendering it would be good to align on a single status code as "discard prerender" (I think 205 as in the prerender spec is the best option, to me 204 feels less descriptive). It would help guide server devs to be consistent by sending this status code rather than sending any 2xx/5xx and expecting the real navigation to re-fetch.

PR #119, opened 2022-01-21
Plumb IP anonymization into prefetch, and use it to set Sec-Purpose.

jeremyroman:
This causes this header to be sent with requests which do not require anonymization:

```
Sec-Purpose: prefetch
```

And this header to be sent with those which do:

```
Sec-Purpose: prefetch;anonymous-client-ip
```

Preview: https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jeremyroman/nav-speculation/7d3fa30f431dabb2b5f027630f30790333538df1/speculation-rules.bs

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -158,7 +158,7 @@ A <dfn export>prefetch IP anonymization policy</dfn> is either null or a [=tuple
 
   1. If |policy| is a tuple and |policy|[0] is "`anonymous-client-ip-when-cross-origin`", then:
       1. Let |origin| be |policy|[1].
-      1. Use |request|'s [=request/URL=]'s [=url/origin=] is the [=same origin|same=] as |origin|, then return false.
+      1. If |request|'s [=request/URL=]'s [=url/origin=] is the [=same origin|same=] as |origin|, then return false.
       1. Return true.
   1. [=Assert=]: |policy| is null.
   1. Return false.
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -312,9 +312,7 @@ These algorithms are based on [=process a navigate fetch=].
             1. Add a parameter whose key is "`anonymous-client-ip`" and whose value is true to the "`prefetch`" token in |purpose|.
             1. The user agent must use a [=connection=] which anonymizes the client IP address (e.g., using a proxy) when fetching |request|, or return "`Blocked`".
 
-                <div class="issue">
-                  Ideally this concept would be plumbed through to the [=obtain a connection=] algorithm eventually.
-                </div>
+                <p class="issue">At the moment, how IP anonymization is achieved is handwaved. This will probably be done in an [=implementation-defined=] manner using some kind of proxy or relay. Ideally this would be plumbed down to [=obtain a connection=], and possibly even the mechanism could be further standardized.</p>
         1. [=header list/Set a structured field value=] given (<a http-header>`` `Sec-Purpose` ``</a>, |purpose|) in |request|'s [=request/header list=].
 
             <div class="note">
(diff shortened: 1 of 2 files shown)


prefetch.bs
```diff
@@ -151,6 +151,19 @@ The user agent may remove elements from the [=prefetch buffer=] even if they are
     1. Return |navigationParams|.
 </div>
 
+A <dfn export>prefetch IP anonymization policy</dfn> is either null or a [=tuple=] consisting of the string "`anonymous-client-ip-when-cross-origin`" and an [=origin=].
```
domenic, line 154: This seems like some attempt to emulate a tagged union? Could we instead do something like:

- A prefetch anonymization policy is either null or an IP anonymization policy.
- An IP anonymization policy has an origin, which is an origin.

Then say things like "If policy is an IP anonymization policy, let origin be policy's origin".

prefetch.bs
```diff
@@ -151,6 +151,19 @@ The user agent may remove elements from the [=prefetch buffer=] even if they are
...
+
+<div algorithm>
+  A [=prefetch IP anonymization policy=] |policy| <dfn for="prefetch IP anonymization policy">requires anonymity</dfn> for [=request=] |request| if the following steps return true:
+
+  1. If |policy| is a tuple and |policy|[0] is "`anonymous-client-ip-when-cross-origin`", then:
+      1. Let |origin| be |policy|[1].
+      1. Use |request|'s [=request/URL=]'s [=url/origin=] is the [=same origin|same=] as |origin|, then return false.
```
domenic, line 161:
```suggestion
      1. If |request|'s [=request/URL=]'s [=url/origin=] is the [=same origin|same=] as |origin|, then return false.
```

prefetch.bs
```diff
@@ -289,18 +302,22 @@ These algorithms are based on [=process a navigate fetch=].
...
-                </div>
-
-                <div class="issue">TODO: Emit the "`anonymous-client-ip`" parameter when applicable.</div>
     1. Let |redirectChain| be an empty [=list=].
     1. Let |shouldBlockNavigationRequest| be the following steps, given [=request=] |request|, [=string=] |navigationType| and [=environment=] |environment|:
         1. [=Assert=]: |navigationType| is "`other`".
         1. [=Assert=]: |request|'s [=request/reserved client=] is |environment|.
+        1. Let |purpose| be a <a spec="RFC8941">List</a> containing the <a spec="RFC8941">Token</a> "`prefetch`".
+        1. If |anonymizationPolicy| [=prefetch IP anonymization policy/requires anonymity=] for |request|, then:
+            1. Add a parameter whose key is "`anonymous-client-ip`" and whose value is true to the "`prefetch`" token in |purpose|.
```
domenic, line 312: Link parameter?

domenic (commented):
> `Sec-Purpose: prefetch;anonymous-client-ip`

That's not correct, right? It'll be `Sec-Purpose: prefetch;anonymous-client-ip=?1`.

jeremyroman:
> > `Sec-Purpose: prefetch;anonymous-client-ip`
> 
> That's not correct, right? It'll be `Sec-Purpose: prefetch;anonymous-client-ip=?1`.

The = and then value is only appended if the value isn't true.
https://www.rfc-editor.org/rfc/rfc8941.html#ser-params

domenic:
🤯

prefetch.bs
```diff
@@ -289,18 +303,22 @@ These algorithms are based on [=process a navigate fetch=].
...
-                </div>
 
-                <div class="issue">TODO: Emit the "`anonymous-client-ip`" parameter when applicable.</div>
     1. Let |redirectChain| be an empty [=list=].
     1. Let |shouldBlockNavigationRequest| be the following steps, given [=request=] |request|, [=string=] |navigationType| and [=environment=] |environment|:
         1. [=Assert=]: |navigationType| is "`other`".
         1. [=Assert=]: |request|'s [=request/reserved client=] is |environment|.
+        1. Let |purpose| be a <a spec="RFC8941">List</a> containing the <a spec="RFC8941">Token</a> "`prefetch`".
+        1. If |anonymizationPolicy| [=prefetch IP anonymization policy/requires anonymity=] for |request|, then:
+            1. Add a parameter whose key is <a for="Sec-Purpose prefetch" lt="anonymous-client-ip">"`anonymous-client-ip`"</a> and whose value is true to the "`prefetch`" token in |purpose|.
```
domenic, line 313: I *think* if you move the `<a>` inside the quotation marks you could get rid of the `lt`, but I don't think you could go all the way to autolinking syntax...

domenic (approved):

PR #122, opened 2022-01-25
User-agent initiated prerendering context

noamr:
Specify an algorithm that can be called by a UA to initiate
prerendering without the need for a predecessor document.

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -309,6 +309,21 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps 
 Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which is initially a {{DOMHighResTimeStamp}} with a time value of zero.
 <p class="note">This is used to keep the time immediately after [=Prompt to unload|prompting to unload=] the [=active document=] of the previous [=browsing context=].
 
+
+<div algorithm="User-agent initiated prerendering">
+  User-agents may choose to initiate prerendering without a predecessor document, for example as a result of the address bar or other browser user interactions.
+
+  To <dfn export>Start user-agent initiated prerendering</dfn> given a [=URL=] |url|:
+
+  1. Let |blankBC| be the result of [=creating a new top-level browsing context=].
+
+  1. Let |prerenderBC| be the result of [=create a prerendering browsing context|creating a prerendering browsing context=] given |url|, "`no-referrer`", and |blankBC|'s [=active document=].
+
+  1. Let |activate| be to [=prerendering browsing context/activate=] |prerenderBC| given |blankBC| and "`replace`".
+
+  1. Return |activate|.
+</div>
+
 <div algorithm="create a prerendering browsing context">
   To <dfn export>create a prerendering browsing context</dfn> given a [=URL=] |startingURL|, a [=referrer policy=] |referrerPolicy|, and a {{Document}} |referrerDoc|:
 
@@ -320,7 +335,7 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
 
   1. Set |bc|'s [=browsing context/loading mode=] to "`prerender`".
 
-  1. If |startingURL|'s [=url/origin=] is not [=same origin=] with |referrerDoc|'s [=Document/origin=], run these steps:
+  1. If |referrerDoc| is not the initial `about:blank` and |startingURL|'s [=url/origin=] is not [=same origin=] with |referrerDoc|'s [=Document/origin=], run these steps:
 
     1. Set |bc|'s [=browsing context/loading mode=] to "`uncredentialed-prerender`".
 
@@ -331,6 +346,8 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
   1. Let |request| be a new [=request=] whose [=request/URL=] is |startingURL| and [=request/referrer policy=] is |referrerPolicy|.
 
   1. [=Navigate=] |bc| to |request| with the [=source browsing context=] set to |referrerDoc|'s [=Document/browsing context=].
+
+  1. Return |bc|.
 </div>
 
 <div algorithm>


nyaxt (approved):

prerendering.bs
```diff
@@ -309,6 +309,21 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
 Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which is initially a {{DOMHighResTimeStamp}} with a time value of zero.
 <p class="note">This is used to keep the time immediately after [=Prompt to unload|prompting to unload=] the [=active document=] of the previous [=browsing context=].
 
+
+<div algorithm="User-agent initiated prerendering">
+  User-agents may choose to initiate prerendering without a predecessor document, for example as a result of the address bar or other browser user interactions.
```
domenic, line 314:
```suggestion
  [=User agents=] may choose to initiate prerendering without a predecessor document, for example as a result of the address bar or other browser user interactions.
```

prerendering.bs
```diff
@@ -309,6 +309,21 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
...
+
+  To <dfn export>Start user-agent initiated prerendering</dfn> given a [=URL=] |url|:
```
domenic, line 316:
```suggestion
  To <dfn export>start user-agent initiated prerendering</dfn> given a [=URL=] |url|:
```

prerendering.bs
```diff
@@ -309,6 +309,21 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
...
+
+  1. Let |blankBC| be the result of [=creating a new top-level browsing context=].
+
+  1. Let |prerenderBC| be the result of [=create a prerendering browsing context|creating a prerendering browsing context=] given |url|, "`no-referrer`", and |blankBC|'s [=active document=].
```
domenic, line 320:
```suggestion
  1. Let |prerenderBC| be the result of [=creating a prerendering browsing context=] given |url|, "`no-referrer`", and |blankBC|'s [=active document=].
```

prerendering.bs
```diff
@@ -309,6 +309,21 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
...
+
+  1. Let |activate| be to [=prerendering browsing context/activate=] |prerenderBC| given |blankBC| and "`replace`".
+
+  1. Return |activate|.
```
domenic, line 324:
```suggestion
  1. Return the result of [=prerendering browsing context/activating=] |prerenderBC| given |blankBC| and "`replace`".
```

prerendering.bs
```diff
@@ -320,7 +335,7 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
 
   1. Set |bc|'s [=browsing context/loading mode=] to "`prerender`".
 
-  1. If |startingURL|'s [=url/origin=] is not [=same origin=] with |referrerDoc|'s [=Document/origin=], run these steps:
+  1. If |referrerDoc| is not the initial `about:blank` and |startingURL|'s [=url/origin=] is not [=same origin=] with |referrerDoc|'s [=Document/origin=], run these steps:
```
domenic, line 338:
```suggestion
  1. If |referrerDoc|'s <a spec="HTML">is initial about:blank</a> is false and |startingURL|'s [=url/origin=] is not [=same origin=] with |referrerDoc|'s [=Document/origin=], run these steps:
```

prerendering.bs
```diff
@@ -331,6 +346,8 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
   1. Let |request| be a new [=request=] whose [=request/URL=] is |startingURL| and [=request/referrer policy=] is |referrerPolicy|.
 
   1. [=Navigate=] |bc| to |request| with the [=source browsing context=] set to |referrerDoc|'s [=Document/browsing context=].
+
+  1. Return |bc|.
```
domenic, line 350: Why add this?

domenic (commented):
Added editorial comments inline.

However the approach in general seems a bit off. Currently all top-level browsing contexts are visible to the user. So creating a new one in step 1 is basically opening a new tab. I don't think that's the right sequence?

Maaaaybe it could work for something like Ctrl-clicking on a bookmark or Shift-Entering on the URL bar. (Although even in that case we'd have to explain that the new TLBC that is created is somehow hidden from the user until activation time.) But for the mainline case, there will be an existing browsing context (even if it's on the new tab page), and the user is typing in the URL bar of it, and then when the user presses enter, we activate the prerendering BC with the existing BC as the predecessor.

This is why I think we probably need an alternate version of the "create a prerendering browsing context" algorithm. All of the steps related to referrerDoc in that algorithm are not really applicable when the "referrer" is the user typing.

noamr:
> Added editorial comments inline.
> 
> However the approach in general seems a bit off. Currently all top-level browsing contexts are visible to the user. So creating a new one in step 1 is basically opening a new tab. I don't think that's the right sequence?
> 
> Maaaaybe it could work for something like Ctrl-clicking on a bookmark or Shift-Entering on the URL bar. (Although even in that case we'd have to explain that the new TLBC that is created is somehow hidden from the user until the user commits to the navigation, at which point we activate.) But for the mainline case, there will be an existing browsing context (even if it's on the new tab page), and the user is typing in the URL bar of it, and then when the user presses enter, we activate the prerendering BC with the existing BC as the predecessor.
> 
> This is why I think we probably need an alternate version of the "create a prerendering browsing context" algorithm. All of the steps related to referrerDoc in that algorithm are not really applicable when the "referrer" is the user typing.

Gotcha, let me rewrite rather than go with the editorial remarks.

noamr:
> > Added editorial comments inline.
> > However the approach in general seems a bit off. Currently all top-level browsing contexts are visible to the user. So creating a new one in step 1 is basically opening a new tab. I don't think that's the right sequence?
> > Maaaaybe it could work for something like Ctrl-clicking on a bookmark or Shift-Entering on the URL bar. (Although even in that case we'd have to explain that the new TLBC that is created is somehow hidden from the user until the user commits to the navigation, at which point we activate.) But for the mainline case, there will be an existing browsing context (even if it's on the new tab page), and the user is typing in the URL bar of it, and then when the user presses enter, we activate the prerendering BC with the existing BC as the predecessor.
> > This is why I think we probably need an alternate version of the "create a prerendering browsing context" algorithm. All of the steps related to referrerDoc in that algorithm are not really applicable when the "referrer" is the user typing.
> 
> Gotcha, let me rewrite rather than go with the editorial remarks.

Refactored to separate the two ways of creating the context, sharing some parts of "activate" (moved to "finalize activation")

prerendering.bs
```diff
@@ -309,6 +309,35 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
...
+
+    1. Let |unsafe time| be the [=unsafe shared current time=].
+
+    1. Update the user agent's user interface to present |bc|, e.g., by updating the tab/window contents and the browser chrome.
+
+    1. [=prerendering browsing context/Finalize activation=] for |bc| given |url|'s [=url/origin=] and |unsafe time|.
+
+  1. Return |activate|.
+
+    <p class="note">The returned algorithm steps can be used by the user-agent when the user is ready to "commit" the navigation.
```
domenic, line 338: Since there are no actual callers, I think it's a bit clearer to change this so that there's no return value, but instead the step that currently reads "Let activate be to do the following in parallel:" to instead be "When the user indicates they wish to commit the navigation:"

We could also add some notes like "NOTE: an example of indicating commitment is pressing the Enter key in the URL bar", and "NOTE: the user might never indicate such a commitment, or might take long enough to do so that the user agent needs to reclaim the resources used by the prerender for some more immediate task. In that case the user agent can TODO" where the TODO stands in for our future definition of discarding prerendering BCs.

domenic (commented):
This looks great, with just a few suggestions about fleshing out the commitment process.

prerendering.bs
noamr: Amended

domenic (approved):
Nice!

noamr:
I can't merge this... can someone do that and/or grant me write permissions? @jeremyroman / @nyaxt ?

PR #125, opened 2022-01-28
Block requests and redirects over insecure transports and where implementation details require it.

### LINKED ISSUE #124: Prefetching and HTTPS
Chromium's implementation of speculation rules prefetch currently requires that URLs have the `https` scheme.

This requirement seems quite sensible in the case where IP anonymization is achieved with a proxy service (to avoid the traffic being visible to the egress node), but does this make sense for prefetch in general?

A few options:

**HTTPS only**: HTTPS is always required, even for non-anonymized prefetching. Authors should move to HTTPS going forward.

**Potentially trustworthy URLs only**: A [potentially trustworthy origin](https://www.w3.org/TR/secure-contexts/#is-origin-trustworthy) is required (or hypothetically a potentially trustworthy URL, if we want to enable pointless prefetching of `about:blank` and `data` URLs). Authors should move to HTTPS going forward, but localhost for development is fine too.

**No HTTPS restriction per se**: Prefetching from non-potentially trustworthy origins is acceptable if it is requested. User agents might, however, use this as an implementation-defined condition for dropping prefetches. For example, a UA that anonymizes IPs using a proxy service might only permit HTTPS traffic for public hosts, and nothing else.

My default preference is for no HTTPS restriction per se, I think. I could see an argument for potentially trustworthy origins only, because moving to that is a good idea regardless and while all plaintext HTTP traffic over the network is a potential security/privacy threat, this is an opportunity to remove some of it that doesn't break sites.

There's a related question of what should happen if a [localhost URL](https://datatracker.ietf.org/doc/html/draft-west-let-localhost-be-localhost) or a host that is on a [private network](https://wicg.github.io/private-network-access/#ip-address-space-heading) when resolved locally is requested with IP anonymity. I'd argue that certainly for localhost and possibly for private IPs, anonymity isn't possible if the anonymizing technology is a proxy service.

jeremyroman:
An attempt to resolve #124.

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -323,6 +323,12 @@ These algorithms are based on [=process a navigate fetch=].
         1. If |partitionKey| is not equal to |proposedPartitionKey|, then return "`Blocked`".
 
             <div class="issue">It might be possible to "downgrade" to [=uncredentialed prefetch=] in this case.</div>
+        1. If |request|'s [=request/URL=] is not [=potentially trustworthy URL|potentially trustworthy=], then return "`Blocked`".
+
+            <div class="note">This is intended to both reduce the likelihood of prefetch traffic being visible to an on-path attacker, and to encourage the use of cryptographic schemes over public networks.</div>
+        1. If |request| cannot be fetched given |anonymizationPolicy| for an [=implementation-defined=] reason, then return "`Blocked`".
+
+            <div class="note">This explicitly acknowledges that implementations might have additional restrictions. For instance, anonymized traffic might not be possible to some hosts, such as those that are not publicly routable and those that have <a href="https://buettner.github.io/private-prefetch-proxy/traffic-advice.html">traffic advice</a> declining private prefetch traffic.
         1. Otherwise, return "`Allowed`".
 
     1. Let |shouldBlockNavigationResponse| be the following steps, given [=request=] |request| and [=response=] |response|:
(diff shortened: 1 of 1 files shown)


prefetch.bs
```diff
@@ -323,6 +323,12 @@ These algorithms are based on [=process a navigate fetch=].
         1. If |partitionKey| is not equal to |proposedPartitionKey|, then return "`Blocked`".
 
             <div class="issue">It might be possible to "downgrade" to [=uncredentialed prefetch=] in this case.</div>
+        1. If |request|'s [=request/URL=] is not [=potentially trustworthy URL|potentially trustworthy=], then return "`Blocked`".
+
+            <div class="note">This is intended to both reduce the likelihood of prefetch traffic being visible to an on-path attacker, and to encourage the use of cryptographic schemes over public networks.</div>
+        1. If |request| cannot be fetched given |anonymizationPolicy| for an [=implementation-defined=] reason, then return "`Blocked`".
+
+            <div class="note">This explicitly acknowledges that implementations might have additional restrictions. For instance, anonymized traffic might not be possible to some hosts, such as those that are not publicly routable and those that have <a href="https://buettner.github.io/private-prefetch-proxy/traffic-advice.html">traffic advice</a> declining private prefetch traffic.
```
domenic, line 331: IMO we should make traffic advice normative at some point?

domenic (approved):
It might be worth factoring out a "should block navigation request" algorithm at some point to reduce duplication... will leave it up to you whether you want to do that in this PR or later.

prefetch.bs
jeremyroman: Maybe.

On the one hand, consistency is good.

On the other hand, it's more descriptive of the proxy than of the web browser, and it's possible we'd get bogged down in the details of how anonymization happens in the process.

PR #128, opened 2022-02-01
Suspend worker execution until activation

noamr:
For the first phase of prerendering, anything that is not commonly part of the loading process is deferred until activation.
Worker script should be loaded though, to avoid that delay when activating.

This is close to the current Chrome implementation. 

See https://github.com/WICG/nav-speculation/issues/42#issuecomment-1025091364

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -545,6 +545,15 @@ Patch the [=navigate=] algorithm to prevent certain navigations in a [=prerender
   1. Otherwise, if <var ignore>browsingContext</var> is a [=prerendering browsing context=], then return.
 </div>
 
+<div algorithm=The worker's lifetime patch">
+  In <a spec=HTML>The worker's lifetime</a>, modify the definition of [=suspendable worker=], to count workers with a [=prerendering browsing context=] owner as [=suspendable worker|suspendable=]:
+
+  A worker is said to be a [=suspendable worker=] if all of its [=WorkerGlobalScope/owner set|owners=] are {{Document/prerendering}} {{Document|Documents}}, or if it is not an [=active needed worker=] but it is a [=permissible worker=].
+
+  <p class="note">This means that the worker's script would load, but the execution would be suspended until the document is activated.
+
+</div>
+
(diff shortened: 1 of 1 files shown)


noamr:
See WPTs: https://github.com/web-platform-tests/wpt/pull/32636

prerendering.bs
```diff
@@ -545,6 +550,15 @@ Patch the [=navigate=] algorithm to prevent certain navigations in a [=prerender
   1. Otherwise, if <var ignore>browsingContext</var> is a [=prerendering browsing context=], then return.
 </div>
 
+<div algorithm=The worker's lifetime patch">
```
domenic, line 553: This should get a new `<h3>` instead of being part of "Preventing forbidden navigations"

prerendering.bs
```diff
@@ -545,6 +550,15 @@ Patch the [=navigate=] algorithm to prevent certain navigations in a [=prerender
...
+  In <a spec=HTML>The worker's lifetime</a>, modify the definition of [=suspendable worker=], to count workers with a [=prerendering browsing context=] owner as [=suspendable worker|suspendable=]:
+
+  A worker is said to be a [=suspendable worker=] if all of its [=WorkerGlobalScope/owner set|owners=] are {{Document/prerendering}} {{Document|Documents}}, or if it is not an [=active needed worker=] but it is a [=permissible worker=].
```
domenic, line 556: "`prerendering` `Documents`" is not a sufficiently-precise way to say this. I think it'd be better to say "if all of its owners are `Document`s whose browsing contexts are prerendering browsing contexts".

At that point we might want two bullet points instead of a big sentence.

domenic (commented):
This looks like the right place to hook, nice find.

I think it might be equivalent to modify the definition of "active needed worker". I'm not sure which would be clearer; both definitions get pretty complicated with this modification.

prerendering.bs
noamr: Done

prerendering.bs
noamr: Done

nyaxt (approved):
Thank you. This looks good to me.

domenic (approved):

PR #130, opened 2022-02-02
Copy cookies after uncredentialed prefetch.

jeremyroman:
Chromium's logic is currently somewhat simpler since it doesn't currently follow redirects (and so can only have cookies for a single request URL). Still, copying all cookies in the isolated partition (which should only have activity from the single prefetch) seems the natural thing to do here.

There is some handwaving mostly around the fact that where a cookie store is obtained from and how it works isn't really written down, other than RFC6265 specifying that you can insert and remove cookies from them, and the fact that half the definition of a network partition key is implementation-defined in Fetch.

Preview: https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jeremyroman/nav-speculation/2d5ef60217c925dbaf3998ba6a93b6f509a55b4e/prefetch.bs

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -241,7 +241,7 @@ Given this, the non-prefetch case becomes, with the small addition of prefetch l
         <div class="note">This step, and the following condition, is added by this specification.</div>
     1. If |prefetchRecord| is not null, then:
         1. Let |navigationParams| be the result of [=performing prefetch response checks=] given |navigationId|, |request|, |prefetchRecord|, |sourceBrowsingContext|, |browsingContext|, |sandboxFlags|, |historyPolicyContainer|, |initiatorPolicyContainer|, |incumbentNavigationOrigin|, and |historyHandling|.
-        1. If |prefetchRecord|'s [=prefetch record/isolated partition key=] is not null, then [=copy prefetch cookies=] given |prefetchRecord|'s [=prefetch record/isolated partition key=] and |navigationParams|' [=navigation params/reserved environment=].
+        1. If |prefetchRecord|'s [=prefetch record/isolated partition key=] is not null, then [=copy prefetch cookies=] given |prefetchRecord|'s [=prefetch record/isolated partition key=] and |navigationParams|'s [=navigation params/reserved environment=].
 
             <div class="note">This copy is complete before continuing, in the sense that subresource fetches, {{Document/cookie|document.cookie}}, etc. can observe the cookies.</div>
         1. [=Process a navigate response=] with |navigationType|, |allowedToDownload|, |hasTransientActivation|, and |navigationParams|.
@@ -486,7 +486,7 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
           <div class="note">|secondKey| is expected to match the value it would have had if this response had been processed as part of an ordinary navigation in |environment|'s [=environment/target browsing context=].</div>
       1. Let |destinationPartitionKey| be (|topLevelSite|, |secondKey|).
       1. Let |cookieStore| be the cookie store associated with |destinationPartitionKey|.
-      1. Let |newCookie| be a new cookie identical to |cookie|.
+      1. Let |newCookie| be a copy of |cookie|.
       1. Set |newCookie|'s creation-time and last-access-time to the current date and time.
       1. If |cookieStore| contains a cookie |existingCookie| with the same name, domain and path as the newly created cookie:
           1. Set the creation-time of |newCookie| to |existingCookie|'s creation-time.
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -66,6 +66,10 @@ spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
     text: Item; url: name-items
     text: List; url: name-lists
     text: Token; url: name-tokens
+spec: COOKIES; urlPrefix: https://httpwg.org/specs/rfc6265.html
+  type: http-header; text: Set-Cookie; url: set-cookie
+  type: dfn; text: cookie; url: storage-model
+  type: dfn; text: receive a cookie; url: storage-model
 </pre>
 
 <h2 id="concepts">Concepts</h2>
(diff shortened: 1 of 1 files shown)


jeremyroman:
This is the major state we have to worry about. Caches and Clear-Site-Data would be the others that come to mind, though I don't think Chromium expressly tries to handle either right now.

prefetch.bs
```diff
@@ -445,6 +454,46 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
...
+
+  1. Let |isolatedCookieStore| be the cookie store associated with |isolatedPartitionKey|.
+  1. For each <a spec="COOKIES">cookie</a> |cookie| in |isolatedCookieStore|.
+      1. Remove |cookie| from |isolatedCookieStore|.
+      1. A user agent may ignore a cookie in its entirety. If so, continue.
+
+          <div class="note">This is consistent with [[COOKIES]] expressly permitting this when [=receiving a cookie=].</div>
+      1. Let |topLevelSite| be null.
+      1. If |environment|'s [=environment/target browsing context=] is a [=top-level browsing context=]:
+          1. Set |topLevelSite| to the result of [=obtaining a site=] given [=tuple origin=] ("`https`", |cookie|'s domain, null, null).
```
domenic, line 476: Feels like there should be a simpler way to do this by directly constructing the site scheme-and-host tuple, but I guess that would duplicate the registrable domain logic which is probably not desirable. Maybe that could be factored out...

prefetch.bs
```diff
@@ -445,6 +454,46 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
...
+  </div>
+
+  1. Let |isolatedCookieStore| be the cookie store associated with |isolatedPartitionKey|.
+  1. For each <a spec="COOKIES">cookie</a> |cookie| in |isolatedCookieStore|.
+      1. Remove |cookie| from |isolatedCookieStore|.
+      1. A user agent may ignore a cookie in its entirety. If so, continue.
+
+          <div class="note">This is consistent with [[COOKIES]] expressly permitting this when [=receiving a cookie=].</div>
+      1. Let |topLevelSite| be null.
+      1. If |environment|'s [=environment/target browsing context=] is a [=top-level browsing context=]:
```
domenic, line 475: I'm probably missing something, but why is this special-cased? Couldn't we get the information directly from _environment_ like the "Otherwise:" case does?

prefetch.bs
```diff
@@ -445,6 +454,46 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
...
+          1. Let |topLevelOrigin| be |environment|'s [=environment/top-level origin=].
+          1. If |topLevelOrigin| is null, then set |topLevelOrigin| to |environment|'s [=environment/top-level creation URL=]'s [=url/origin=].
+          1. [=Assert=]: |topLevelOrigin| is an [=origin=].
+          1. Set |topLevelSite| be the result of [=obtaining a site=] given |topLevelOrigin|.
+      1. Let |secondKey| be null or an [=implementation-defined=] value.
+
+          <div class="note">|secondKey| is expected to match the value it would have had if this response had been processed as part of an ordinary navigation in |environment|'s [=environment/target browsing context=].</div>
+      1. Let |destinationPartitionKey| be (|topLevelSite|, |secondKey|).
+      1. Let |cookieStore| be the cookie store associated with |destinationPartitionKey|.
+      1. Let |newCookie| be a new cookie identical to |cookie|.
```
domenic, line 489: I think "a clone" or "a copy" of |cookie| would probably be the better underdefined term here, instead of "a new cookie identical to".

prefetch.bs
```diff
@@ -445,6 +454,46 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
...
+      1. Set |newCookie|'s creation-time and last-access-time to the current date and time.
+      1. If |cookieStore| contains a cookie |existingCookie| with the same name, domain and path as the newly created cookie:
+          1. Set the creation-time of |newCookie| to |existingCookie|'s creation-time.
+          1. Remove |existingCookie| from |cookieStore|.
+      1. Insert |newCookie| into |cookieStore|.
```
domenic, line 494: Removal and reinsertion changes the ordering; do you know if the ordering matters for cookie stores?

domenic (commented):
https://github.com/WICG/nav-speculation/blob/main/fetch.md#cache-partitioning-after-activation I think needs updating, although maybe we should roll that into a general "update the explainers" task (which could involve rewriting them to be focused around what we've actually settled on instead of our initial exploratory sketches).

There's an issue 9 "Further review is needed of the cookies-present handling and its interaction with redirects and caches." which is probably obsolete. Maybe issue 7 as well?

prefetch.bs
```diff
@@ -236,6 +241,9 @@ Given this, the non-prefetch case becomes, with the small addition of prefetch l
         <div class="note">This step, and the following condition, is added by this specification.</div>
     1. If |prefetchRecord| is not null, then:
         1. Let |navigationParams| be the result of [=performing prefetch response checks=] given |navigationId|, |request|, |prefetchRecord|, |sourceBrowsingContext|, |browsingContext|, |sandboxFlags|, |historyPolicyContainer|, |initiatorPolicyContainer|, |incumbentNavigationOrigin|, and |historyHandling|.
+        1. If |prefetchRecord|'s [=prefetch record/isolated partition key=] is not null, then [=copy prefetch cookies=] given |prefetchRecord|'s [=prefetch record/isolated partition key=] and |navigationParams|' [=navigation params/reserved environment=].
```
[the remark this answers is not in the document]
jeremyroman: Done

prefetch.bs
jeremyroman: Because the top-level site at the time the cookie is assigned would have been determined not by wherever the navigation ended up (which is what environment reflects at this time), but by the URL that was being fetched then. The two are not necessarily equal, because if we're following redirects (the spec assumes we do), then we might have picked up cookies from those origins too (e.g., we might have followed a redirect from example.com to www.example.com, and if the former set a cookie it should be copied faithfully).

prefetch.bs
jeremyroman: Yeah, it would require either duplicating that (which makes it less obviously equivalent) or splitting the HTML obtaining a site algorithm into two two-line algorithms, which it's not clear is worthwhile.

If you'd rather duplicate, I can do that, but I tried and it seemed messier to me. :/

prefetch.bs

prefetch.bs
jeremyroman: RFC 6265 defines the ordering. The simplest way to argue this is correct is that it's what the "receive a cookie" algorithm does:
https://httpwg.org/specs/rfc6265.html#:~:text=Remove%20the%20old%2Dcookie,into%20the%20cookie%20store.

I'd considered using that instead but arranging for its parameters to be available is even more work, so I've just duplicated the "remove existing cookie, insert new cookie" bit.

jeremyroman (commented):
sorry for dropping this for a bit; ptal

jeremyroman:
> https://github.com/WICG/nav-speculation/blob/main/fetch.md#cache-partitioning-after-activation I think needs updating, although maybe we should roll that into a general "update the explainers" task (which could involve rewriting them to be focused around what we've actually settled on instead of our initial exploratory sketches).

That's a good point. I think it's probably easier done as a separate task.

> There's an issue 9 "Further review is needed of the cookies-present handling and its interaction with redirects and caches." which is probably obsolete. Maybe issue 7 as well?

Resolved both of these.

jeremyroman:
Specifically, filed #139.

prefetch.bs
domenic: Maybe add a note, saying how this remove-then-insert instead of update-in-place is similar to "receive a cookie".

prefetch.bs
domenic: So to confirm I understand, this means what we have here only works for prefetches (where child browsing contexts are not fetched before the redirect kicks in), or prerenders where we avoid loading child browsing contexts until activation. If we ever used this for a version of cross-origin prerendering which loaded child browsing contexts before activation, then we'd need to do a similar dance for establishing the right top-level site for their cookies. Right?

This feels a bit delicate and maybe worth explaining in a couple notes. Here's my attempt:

> When copying the top-level browsing context's cookies, we use the cookie's domain to establish _topLevelSite_ and thus _destinationPartitionKey_. The cookie might have been delivered at any point in the redirect chain while fetching the top-level document, so its domain is not necessarily equal to the site corresponding to _environment_'s top-level origin.

> When copying nested browsing contexts' cookies, we know that _environment_'s top-level origin, which reflects the final destination of the top-level browsing context, is sufficient to establish _topLevelSite_ and thus _destinationPartitionKey_. This is because we only fetch nested browsing context documents after any top-level document redirects have ceased.

domenic (approved):
OK, just a couple of requests for extra notes. If I totally misunderstood the top-level site thing then maybe don't merge until I do :)

prefetch.bs
jeremyroman: Well, close. Since this is for prefetch, either the whole prefetch relates to a top-level navigation (in which case every fetch, redirect or no, is top-level and so the top-level site matches that of the cookie's domain) or the whole prefetch occurs in some nested browsing context (in which case in which case the top-level site is that of the top-level browsing context).

Because subframes of the prefetched resource don't exist until document parsing (and nothing currently parses the document), there are no cookies thereto appertaining.

But yes the logic for merging cookies for prerendering including subframes (in fact, including any subresources) would need more verbiage to do this correctly. Much of this complexity just arises from the fact that these abstractions don't really exist in the cookie spec and only kinda exist in Fetch.

Added two notes (one rather long) in the hopes of clarifying.

jeremyroman:
added more notes; ptal

domenic (approved):
Great, thanks for the extra iteration to make things extra-clear. I'm quite sure it will pay off in the future :)

PR #131, opened 2022-02-02
Defer new SW registration from a prerendered page until it's activated

noamr:
This prevents nonsensical behavior such as a prerendering context creating a service-worker that can create notifications or cache in a way that affects non-prerendering contexts.

An attempt at https://github.com/WICG/nav-speculation/issues/44

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -173,7 +173,8 @@ spec: speech-api; urlPrefix: https://wicg.github.io/speech-api/
     text: stop(); url: dom-speechrecognition-stop
     text: abort(); url: dom-speechrecognition-abort
 spec: service-workers; urlPrefix: https://w3c.github.io/ServiceWorker/
-  text: Service worker installation algorithm; url: installation-algorithm 
+  type: dfn
+    text: Service worker installation algorithm; url: installation-algorithm
 </pre>
 <pre class="biblio">
 {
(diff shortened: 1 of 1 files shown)


nyaxt:
@nhiroki What do you think of this PR?

noamr:
Hmm I guess this makes less sense for the omnibox case, where the SW would usually be brand new. maybe it's better to just defer registration instead.

noamr:
> Hmm I guess this makes less sense for the omnibox case, where the SW would usually be brand new. maybe it's better to just defer registration instead.

Refactored to defer instead of discard. This is better for the user-agent initiated prerendering case, where otherwise any page with SW would not work.

nhiroki:
LGTM. Deferring service worker sounds good to me.

noamr:
> LGTM. Deferring service worker registration sounds good to me.

Great! I think the way this PR works needs more baking, but I'll follow that path and also create the appropriate WPTs.

noamr:
@domenic, WDYT?

domenic (approved):
LGTM with nit but I will continue discussion in the issue.

noamr:
Can someone please merge?

nhiroki:
LGTM, but I don't have permission to merge.

nyaxt:
@jeremyroman Would you merge this PR?

jeremyroman:
Sorry for not catching this when I was pinged; I've fallen behind on email. Added @nyaxt and @nhiroki to the merge ACL to remove this obstacle in the future.

PR #134, opened 2022-02-08
Add explainer for the first phase

noamr:
Added an explainer, derived from previous explainer and current state of affairs, that focuses solely on the omnibox case and the set of specified features today.


nyaxt:
Thanks. This LGTM

PR #137, opened 2022-02-09
Add opt-out text instead of opt-in

### LINKED ISSUE #136: UA-initiated prerendering does not have an opt-in
https://github.com/WICG/nav-speculation/blob/b6e7cf048f15d763c922b2a96cdb3348fb491c82/ua-initiated-prerendering.md says

> This navigation takes place using [special fetch modes](https://github.com/WICG/nav-speculation/blob/main/fetch.md), which ensure that https://b.example/ has [opted in](https://github.com/WICG/nav-speculation/blob/main/opt-in.md) to being prerendered.
>
> Within this prerendering browsing context, assuming the opt-in check passes, ...

I don't believe there is any opt-in for UA initiated prerendering, so this should be removed. We could mention #133 (once it's specced) as a way to do server-side opt-outs.

noamr:
Note that there is no consensus on the solution yet so it has a TODO

Closes #136

### DIFF
--- a/ua-initiated-prerendering.md
+++ b/ua-initiated-prerendering.md
@@ -74,7 +74,7 @@ When a document is fetched for the purpose of prerendering, the user-agent sends
 
 The server may decide at this point to cancel the prerendering, which would cause a full load of the document once the user performs an actual to the URL, by responsing [as described here](https://wicg.github.io/nav-speculation/prerendering.html#no-bad-navs). (TODO: put actual description here once we reach consensus)
 
-Developers might decide to implement such response, for example, in order to reduce server load in case where there are too many unfulfilled prerendered, or if prerendering may cause the page to reach some error condition.
+Developers might decide to implement such response, for example, in order to reduce server load in case where there are too many unfulfilled prerenders, or if prerendering may cause the page to reach some error condition.
 
 ## Restrictions
 
--- a/ua-initiated-prerendering.md
+++ b/ua-initiated-prerendering.md
@@ -70,7 +70,7 @@ This completes the journey to a fully-rendered view of `https://b.example/`, in 
 
 ## Opting Out
 
-When a document is fetched for the purpose of prerendering, the user-agent sends an additional header: `Sec-Purpose: prefetch; prerender`. See [here]([`Sec-Purpose` header](https://wicg.github.io/nav-speculation/prefetch.html#sec-purpose-header)) for more details.
+When a document is fetched for the purpose of prerendering, the user-agent sends an additional header: `Sec-Purpose: prefetch; prerender`. See [the spec](https://wicg.github.io/nav-speculation/prefetch.html#sec-purpose-header) for more details.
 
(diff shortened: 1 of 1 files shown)


ua-initiated-prerendering.md
```diff
@@ -67,6 +68,14 @@ Notification.requestPermission().then(() => {
 
 This completes the journey to a fully-rendered view of `https://b.example/`, in a user-visible top-level browsing context.
 
+## Opting Out
+
+When a document is fetched for the purpose of prerendering, the user-agent sends an additional header: `Sec-Purpose: prefetch; prerender`. See [here]([`Sec-Purpose` header](https://wicg.github.io/nav-speculation/prefetch.html#sec-purpose-header)) for more details.
```
domenic, line 73:
```suggestion
When a document is fetched for the purpose of prerendering, the user-agent sends an additional header: `Sec-Purpose: prefetch; prerender`. See [the spec](https://wicg.github.io/nav-speculation/prefetch.html#sec-purpose-header) for more details.
```

ua-initiated-prerendering.md
```diff
@@ -67,6 +68,14 @@ Notification.requestPermission().then(() => {
...
+
+The server may decide at this point to cancel the prerendering, which would cause a full load of the document once the user performs an actual to the URL, by responsing [as described here](https://wicg.github.io/nav-speculation/prerendering.html#no-bad-navs). (TODO: put actual description here once we reach consensus)
+
+Developers might decide to implement such response, for example, in order to reduce server load in case where there are too many unfulfilled prerendered, or if prerendering may cause the page to reach some error condition.
```
domenic, line 77:
```suggestion
Developers might decide to implement such response, for example, in order to reduce server load in case where there are too many unfulfilled prerenders, or if prerendering may cause the page to reach some error condition.
```

ua-initiated-prerendering.md
```diff
@@ -67,6 +68,14 @@ Notification.requestPermission().then(() => {
 
 This completes the journey to a fully-rendered view of `https://b.example/`, in a user-visible top-level browsing context.
 
+## Opting Out
+
+When a document is fetched for the purpose of prerendering, the user-agent sends an additional header: `Sec-Purpose: prefetch; prerender`. See [here]([`Sec-Purpose` header](https://wicg.github.io/nav-speculation/prefetch.html#sec-purpose-header)) for more details.
+
+The server may decide at this point to cancel the prerendering, which would cause a full load of the document once the user performs an actual to the URL, by responsing [as described here](https://wicg.github.io/nav-speculation/prerendering.html#no-bad-navs). (TODO: put actual description here once we reach consensus)
```
domenic, line 75:
```suggestion
The server may decide at this point to cancel the prerendering, which would cause a full load of the document once the user performs an actual to the URL, by responsing [as described here](https://wicg.github.io/nav-speculation/prerendering.html#no-bad-navs). (TODO: we should pick one recommended response strategy from among the linked response strategies, and expand on it here.)
```

domenic (approved):

PR #140, opened 2022-02-15
Delay BroadcastChannel and ServiceWorker postMessage

noamr:
Closes https://github.com/WICG/nav-speculation/issues/141
Delay posting messages to broadcast channels and service workers, to avoid nonsensical behavior such as prerendering tabs notifying other tabs that they are active.

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -188,8 +188,13 @@ spec: service-workers; urlPrefix: https://w3c.github.io/ServiceWorker/
     text: unregister(); url: dom-serviceworkerregistration-unregister
   type: interface
     text: ServiceWorkerContainer; url: serviceworkercontainer-interface
-  type: method; for: ServiceWorkerRegistration
+  type: method; for: ServiceWorkerContainer
     text: register(scriptURL, options); url: dom-serviceworkercontainer-register
+  type: interface
+    text: ServieWorker; url: serviceworker-interface
+  type: method: for: ServiceWorker
+    text: postMessage(message, transfer); url: dom-serviceworker-postmessage
+    text: postMessage(message, options); url: service-worker-postmessage-options
 </pre>
 <pre class="biblio">
 {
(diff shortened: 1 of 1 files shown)


jeremyroman:
Is this sufficient? What about `message` and `messageerror` events firing? This makes pages able to receive but not send messages, and buffers outgoing messages. Is that desirable?

It's not obvious to me whether it's better to make the BroadcastChannel essentially unplugged (drop incoming and outgoing messages while prerendering), or buffered (for both incoming and outgoing messages), or something else (if we don't want it to function), but we should probably write down the reasoning for our choice.

noamr:
> Is this sufficient? What about `message` and `messageerror` events firing? This makes pages able to receive but not send messages, and buffers outgoing messages. Is that desirable?
> 
> It's not obvious to me whether it's better to make the BroadcastChannel essentially unplugged (drop incoming and outgoing messages while prerendering), or buffered (for both incoming and outgoing messages), or something else (if we don't want it to function), but we should probably write down the reasoning for our choice.

Gotcha, this is not fully thought out. Will start an issue to discuss this.

noamr:
As I mentioned in the discussion, the main reasoning for this is to avoid nonsensical behavior in one of the main cases of `BroadcastChannel`, which is to detect "other open tabs", scenarios like "we see that you've logged in from another tab, do you want to log out here?". Deferring outgoing messages can achieve that without disrupting other flows.

prerendering.bs
```diff
@@ -186,8 +188,13 @@ spec: service-workers; urlPrefix: https://w3c.github.io/ServiceWorker/
     text: unregister(); url: dom-serviceworkerregistration-unregister
   type: interface
     text: ServiceWorkerContainer; url: serviceworkercontainer-interface
-  type: method; for: ServiceWorkerRegistration
+  type: method; for: ServiceWorkerContainer
     text: register(scriptURL, options); url: dom-serviceworkercontainer-register
+  type: interface
+    text: ServieWorker; url: serviceworker-interface
```
nyaxt, line 194: typo: ServiceWorker

nyaxt (changes requested):

jeremyroman:
lgtm2 fwiw; thanks for following up

PR #144, opened 2022-02-18
Drop prefetch/prerender responses without an ok status.

jeremyroman:


jeremyroman:
A slightly interesting consequence of this is that this also affects navigations initiated within the prerendering browsing context after the first one. We could not do that if we think it's better, but this is consistent with the versions already there.

noamr (approved):

prerendering.bs
```diff
@@ -573,6 +576,15 @@ Patch the [=navigate=] algorithm to prevent certain navigations in a [=prerender
 
 <p class="issue">Portals might need an extra hook to close the portal in these cases. Or should we reconsider and just do nothing for portals too? That might be more elegant. I think it just requires portals to not be so zealous about clearing the host element/browsing context link, which isn't observable anyway?
 
+<div algorithm>
```
noamr, line 579: No need for a new algorithm, this should be incorporated into https://wicg.github.io/nav-speculation/prerendering.html#no-bad-navs

domenic: +1, I think just adding "_response_ does not support prefetch" to that list would be pretty nice and minimal. Although I appreciate the motivation to have a clearly-named "supports prerender" algorithm, all the other stuff in the no-bad-navs section arguably also would belong there...

noamr (approved):

PR #154, opened 2022-04-07
Perform navigate-to CSP checks when activating

### LINKED ISSUE #43: Prerendering and CSP
How should the CSP of a page affect a prerendered page that was triggered by that page?

One option is to say the CSP of the triggering page should propagate to the prerendered page, since the prerendered page is not something that will necessarily be navigated to, so it should still be subject to the constraints of the triggering page. However I feel that's problematic:

- The CSP can break the page in a way that's not reproducible when the page is navigated to normally without prerendering.
- When the page is activated, the CSP would presumably need to be change to become the normal CSP of the page.

Admittedly both of those arguments hold for any restrictions on prerendered content, so they may not be convincing. However, I'm tending to think that as its own browsing context, the prerendered page should be subject to only its own CSP and not the triggering page's. With that goal, here is a first cut at proposed behavior for relevant CSP directives:

### navigate-to

Background: CSP [navigate-to](https://w3c.github.io/webappsec-csp/#directive-navigate-to) restricts what URLs a document can initiate navigation to via <a>, window.location, etc. The CSP of the initiator document, not the target document, specifies the policy. From my reading of the spec, navigate-to applies to redirects of the navigation request too.

Proposal: This should behave mostly like a link click. Apply navigate-to checks to the prerender initial URL and to any redirects. For example, suppose a document has `navigate-to: a.test`, and it triggers a prerender to `a.test` which redirects to `b.test`. The document shouldn't be able to  later initiate a navigation to `a.test` which activates the prerendered page, resulting in a navigation to `b.test`. However, if the prerendered page does a client-side navigation from `a.test` to `b.test`, it would be fine for the document to initiate a navigation to `a.test` which activates the prerendered page, since this is the behavior I expect a link click to have under this CSP.

The easiest way to enforce this would be to cancel prerendering if the initial URL or any redirect violates `navigate-to`, and once the first response is committed let the prerendering page continue without constraints of the initiator page's CSP. (Otherwise, on activation, replay the redirect chain of prerendering and block the activation navigation if it would result in a violation of the directive.)

### connect-src

Background: CSP [connect-src](https://w3c.github.io/webappsec-csp/#directive-connect-src) restricts which requests a document can make by script.

Proposal: APIs that trigger prerendering will be constrained by this directive for the prerendering navigation request and its redirects. Subresource requests from the prerendered page will not be affected by the connect-src of the triggering document. Furthermore, once the prerendered page commits, any client-side navigation is not affected by connect-src.

### prefetch-src

Background: CSP [prefetch-src](https://w3c.github.io/webappsec-csp/#directive-prefetch-src) restricts which resources can be prefetched/prerendered from a document.

Proposal: Same as connect-src: it only acts on the main resource and not on the subresources of the prerendered page. 

What do people think? cc @domenic @jeremyroman @domfarolino

noamr:
- When prerendering, the referring doc's policy container should be used to ensure its `navigate-to` is respected
- When activating, an additional `navigate-to` check should be made, in case it had changed in the meantime
  (e.g. by adding a `<meta http-equiv="Content-Security-Policy">` element dynamically)

Closes #43

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -403,9 +403,7 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
 
      <p class="note">As with regular [=browsing contexts=], the [=prerendering browsing context=] can be [=discarded=] for any reason, for example if it becomes unresponsive, performs a restricted operation, or if the user agent believes prerendering takes too much resources.
 
-  1. Let |request| be a new [=request=] whose [=request/URL=] is |startingURL|, [=request/referrer policy=] is |referrerPolicy|, and [=request/policy container=] is |referrerDoc|'s [=Document/policy container=], [=clone a policy container|cloned=].
-
-     <p class="note">The [=policy container=] is cloned to avoid a situation where changes in [=Content Security Policy=] affect a prerender that had already started.
+  1. Let |request| be a new [=request=] whose [=request/URL=] is |startingURL|, and [=request/referrer policy=] is |referrerPolicy|.
 
   1. [=Navigate=] |bc| to |request| with the [=source browsing context=] set to |referrerDoc|'s [=Document/browsing context=].
 </div>
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -403,7 +403,7 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
 
      <p class="note">As with regular [=browsing contexts=], the [=prerendering browsing context=] can be [=discarded=] for any reason, for example if it becomes unresponsive, performs a restricted operation, or if the user agent believes prerendering takes too much resources.
 
-  1. Let |request| be a new [=request=] whose [=request/URL=] is |startingURL|, and [=request/referrer policy=] is |referrerPolicy|.
+  1. Let |request| be a new [=request=] whose [=request/URL=] is |startingURL| and [=request/referrer policy=] is |referrerPolicy|.
 
   1. [=Navigate=] |bc| to |request| with the [=source browsing context=] set to |referrerDoc|'s [=Document/browsing context=].
 </div>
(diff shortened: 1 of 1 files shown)


prerendering.bs
```diff
@@ -403,7 +403,9 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
 
      <p class="note">As with regular [=browsing contexts=], the [=prerendering browsing context=] can be [=discarded=] for any reason, for example if it becomes unresponsive, performs a restricted operation, or if the user agent believes prerendering takes too much resources.
 
-  1. Let |request| be a new [=request=] whose [=request/URL=] is |startingURL| and [=request/referrer policy=] is |referrerPolicy|.
+  1. Let |request| be a new [=request=] whose [=request/URL=] is |startingURL|, [=request/referrer policy=] is |referrerPolicy|, and [=request/policy container=] is |referrerDoc|'s [=Document/policy container=], [=clone a policy container|cloned=].
```
hiroshige-g, line 406: |request|'s policy container is overwritten to a clone of [source browsing context](https://html.spec.whatwg.org/multipage/browsing-the-web.html#source-browsing-context)'s [active document](https://html.spec.whatwg.org/multipage/browsers.html#active-document)'s [policy container](https://html.spec.whatwg.org/multipage/dom.html#concept-document-policy-container) in Steps 10-11 of https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigate. So if |referrerDoc| is the active document of its browsing context, then this step is not needed.

noamr: You're right! So only the activation part is needed.

nyaxt:
@hiroshige-g What do you think of the latest uploaded version? Would you `approve` if it lgty?

hiroshige-g:
LGTM.
It's probably better to rename the title, e.g. to "Perform navigate-to CSP checks when activating", because this PR no longer touches the CSP on prererndering time.

nyaxt (approved):

PR #156, opened 2022-04-12
Attempt to spell out history merger in more detail

domenic:
/cc @altimin @rakina @jakearchibald @noamr; not sure if you're the best reviewers but I thought you might be interested.

Before merging, we should get at least one an implementer review to make sure what I'm saying here seems reasonable from the point of view of our current implementation. @nyaxt maybe?

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -214,7 +214,7 @@ spec: nav-speculation; urlPrefix: prefetch.html
 </pre>
 
 <style>
-/* domintro from https://resources.whatwg.org/standard.css */
+/* domintro and XXX from https://resources.whatwg.org/standard.css */
 .domintro {
   position: relative;
   color: green;
@@ -242,6 +242,12 @@ spec: nav-speculation; urlPrefix: prefetch.html
   position: absolute;
   top: -0.8em;
   left: -0.8em;
+}
+
+.XXX {
+  color: #D50606;
+  background: white;
+  border: solid #D50606;
 }
 </style>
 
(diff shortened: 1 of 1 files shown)


prerendering.bs
```diff
@@ -429,7 +435,29 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
...
+
+       - Move |successorBC| from |successorTraversable| into |predecessorTraversable|.
+
+       - Move the session history entry |successorEntry| from |successorTraversable| into |predecessorTraversable|, as follows:
+
+          - If |historyHandling| is "`replace`", delete the currently-active session history entry of |predecessorTraversable|.
+
+          - If |historyHandling| is "`default`", delete every entry after the the currently-active session history entry of |predecessorTraversable|.
+
+          - Append |successorEntry| into the |predecessorTraversable|'s session history list, and make it the active session history entry. (As a side effect, this then makes |successorBC| the active browsing context of |predecessorTraversable|).
```
rakina, line 456: Shouldn't this try to replace the current session history in place instead of append if  |historyHandling| is "replace"?

nyaxt (approved):
LGTM myside

prerendering.bs
domenic: If it's replace, we deleted the currently-active one, so appending at this point makes sense I think.

rakina: I was thinking of the case when the currently-active one is not the last entry in the session history list (e.g. after a back navigation?) but maybe that's OK too?

domenic: Oh, I think you're right, let me re-do it with that case in mind. Thank you!

PR #157, opened 2022-04-12
Allow removing and updating speculation rule sets.

jeremyroman:
Removing seems necessary. Allowing updates without removing and inserting around that seemed like we might as well support that too, similarly to how `<style>` behaves.

### DIFF
--- a/speculation-rules.bs
+++ b/speculation-rules.bs
@@ -228,7 +228,7 @@ A [=document=] has a <dfn for=document export>list of speculation rule sets</dfn
 Periodically, for any [=document=] |document|, the user agent may [=queue a global task=] on the <a data-link-type="interface">DOM manipulation task source</a> with |document|'s [=relevant global object=] to [=consider speculation=] for |document|.
 
 <p class="note">
-  The user agent will likely do when resources are idle and available, or otherwise the circumstances of its previous decision whether to start a speculation may have changed.
+  The user agent will likely do when resources are idle and available, or otherwise the circumstances of its previous decision whether to start a speculation could have changed.
 </p>
 
 <div algorithm="consider speculation">
--- a/speculation-rules.bs
+++ b/speculation-rules.bs
@@ -121,9 +121,11 @@ The following algorithms are updated accordingly:
   The following steps are added as the <{script}> element's [=HTML element removing steps=], given |removedNode| and <var ignore>oldParent</var>:
 
   1. If [=the script's result=] is a [=speculation rule set=], then:
-      1. [=list/Remove=] it from |removedNode|'s [=Node/node document=]'s [=document/list of speculation rule sets=].
+      1. Let |document| be |removedNode|'s [=Node/node document=].
+      1. [=list/Remove=] it from |document|'s [=document/list of speculation rule sets=].
       1. Set |removedNode|'s [=already started=] flag to false.
       1. Set [=the script's result=] to null.
+      1. [=Queue a global task=] on the <a data-link-type="interface">DOM manipulation task source</a> with |document|'s [=relevant global object=] to [=consider speculation=] for |document|.
 
       <div class="note">This means that the rule set can be reparsed if the script is reinserted.</div>
 </div>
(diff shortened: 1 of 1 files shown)


domenic (approved):
What's here looks good, but I'm wondering if we should revise this now?

> Periodically, for any [document](https://dom.spec.whatwg.org/#concept-document) document, the user agent may [queue a global task](https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-global-task) on the [DOM manipulation task source](https://html.spec.whatwg.org/multipage/webappapis.html#dom-manipulation-task-source) with document’s [relevant global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global) to [consider speculation](https://wicg.github.io/nav-speculation/speculation-rules.html#consider-speculation) for document.
>
> The user agent will likely do this after the insertion of new speculation rules, or when resources are idle and available.

Either by suggesting specific points at which to consider speculation, or by adding a should after the removing/children changed steps to consider speculation at that time?

jeremyroman:
Good call. Given the discussion on #158, I think the UA should be forced to consider speculation on each of those changes, with a future PR to actually make "consider speculation" do cancellations.

I did so as a task, but I'm curious about what you think of the pros/cons of that vs a microtask here.

A separately queued task makes sure we avoid thrashing at least within the same task, including microtasks (e.g., an async function which removes a rule and then another which awaits it and adds a rule back). However while it might execute close to immediately, there might be a substantial amount of work queued in front of it.

Using a microtask would mean it will happen at a time that's slightly more predictable (since fewer wholly unrelated things will happen between the action and the microtask checkpoint), but this microtask might happen before other microtasks which modify the rules in ways that we would've liked to have known about.

It feels like the ideal midpoint is after the microtask queue becomes empty but before the end of the microtask checkpoint, like how "Cleanup Indexed Database transactions" works. But hooking into that doesn't seem to be common in the HTML spec, with most things using either a task or a microtask and leaving it at that.

domenic:
Probably we should introduce a new task source? Then we can say that the UA can prioritize it as highly as it wants, e.g. before any other tasks. (Whereas if you reuse the DOM manipulation task source, at the very least other things queued on the DOM manipulation task source will have to run first.)

I think that's basically equivalent to your ideal midpoint. IndexedDB _requires_ going before any other tasks, so it uses a special carve-out, but if we just want to _allow_ going before any other tasks, a new dedicated task source seems like the way to go.

jeremyroman:
I'm still worried about predictability, since that doesn't require the UA to do so immediately, and if you're trying to use this to cancel and try again later you won't be able to tell when that has happened.

Looking at similar places in the spec, I guess this isn't so dissimilar to mutation observers, or how images work ([reacting to environment changes](https://html.spec.whatwg.org/multipage/images.html#reacting-to-environment-changes:await-a-stable-state), [update the image data](https://html.spec.whatwg.org/multipage/images.html#updating-the-image-data)). Both use microtasks. Maybe that's fine here too -- we give up the possibility of avoiding cancellations between promise thens, but we make when that happens more predictable, and make stuff like this possible and consistent:

```js
async function resetRuleSet() {
  let ruleSet = document.querySelector('script#example-rule-set');
  let parent = ruleSet.parentNode;
  ruleSet.remove();
  await Promise.resolve();
  parent.appendChild(ruleSet);
}
```

If you don't object I'm going to see what a rewrite that just uses a microtask looks like.

domenic:
I feel like we are trying to shoehorn an imperative API into a declarative one. IMO it should be OK for the declarative one to not have strong guarantees of this sort. We should investigate an imperative one if we want strong guarantees... https://github.com/WICG/nav-speculation/issues/98

jeremyroman:
Even declarative APIs should be at least somewhat predictable about when those declarations might have effect and when they do not, surely? For example, CSS is declarative but has very clear rules about how those declarations apply.

domenic:
We've already said this is API is clearly not predictable; sometimes the browser will prerender, and sometimes it won't. Saying that the removal is definitely predictable but the insertion is not is strange.

Anyway, I don't think the normal variation in letting other tasks run before you counts as making something "unpredictable"; e.g. the fact that `history.back()` queues a task to traverse you back, or `location.href = foo` queues a task to navigate you, doesn't make them unpredictable.

jeremyroman:
Hmm, the reason I don't see it as strange is that I see the rules as saying "while these rules exist, the UA might preload the stuff you mention". So without mutation/removal of any kind, we're already in a world where stuff that you suggest might or might not be prefetched, but stuff that you don't definitely won't be. If we can relatively easily provide at least that assurance to authors, it seems useful.

By way of imperfect analogy, if you insert an `` into a document, when it loads is not entirely predictable. When you remove it, on the other hand, it goes away at a predictable moment.

jeremyroman:
For comparison, uploaded a commit that queues a microtask instead and is more explicit that the prefetch start can occur in parallel. Can revert it if necessary.

PR #166, opened 2022-06-24
Specify that prefetches are aborted when no speculation candidate continues to authorize them.

jeremyroman:
This defines a concept of a "prefetch controller" which can be threaded into the prefetch code to allow the underlying fetches to be aborted and which will prevent the prefetch from being served from the prefetch buffer. This combination means that a prefetch which isn't otherwise cacheable (e.g., because it has Cache-Control: no-store) won't be used due to the special rules that apply to prefetch.

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -434,10 +434,10 @@ These algorithms are based on [=process a navigate fetch=].
     1. Let (|response|, |locationURL|) be the result of [=performing a common navigational fetch=] given |request|, "`other`", |browsingContext|, null, an empty algorithm, |shouldBlockNavigationRequest|, and |shouldBlockNavigationResponse|, and |prefetchRecord|'s [=prefetch record/fetch controller=].
     1. If |locationURL| is failure or a [=URL=] whose [=url/scheme=] is not an [=HTTP(S) scheme=], then set |response| to a [=network error=].
     1. TODO: navigate-to, frame-src, XFO enforcement should probably be left to navigation, but what about Content-Disposition?
-    1. If |response| is a [=network error=], then [=prefetch record/cancel and discard=] |prefetchRecord| and return.
+    1. If |response| is a [=network error=], then [=prefetch record/cancel and discard=] |prefetchRecord| given |document|, and return.
     1. [=Assert=]: |response| is [=exchange record/response=] the last element of |prefetchRecord|'s [=prefetch record/redirect chain=].
-    1. If |response| does not [=support prefetch=], then [=prefetch record/cancel and discard=] |prefetchRecord| and return.
-    1. [=prefetch record/Complete=] |prefetchRecord|.
+    1. If |response| does not [=support prefetch=], then [=prefetch record/cancel and discard=] |prefetchRecord| given |document|, and return.
+    1. [=prefetch record/Complete=] |prefetchRecord| given |document|.
 </div>
 
 The <dfn>list of sufficiently strict speculative navigation referrer policies</dfn> is a list containing the following: "", "`strict-origin-when-cross-origin`", "`strict-origin`", "`same-origin`", "`no-referrer`".
@@ -447,7 +447,7 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
 
     1. [=Assert=]: |prefetchRecord|'s [=prefetch record/URL=]'s [=url/scheme=] is an [=HTTP(S) scheme=].
     1. [=list/Append=] |prefetchRecord| to |document|'s [=Document/prefetch records=].
-    1. If |prefetchRecord|'s [=prefetch record/referrer policy=] is not in the [=list of sufficiently strict speculative navigation referrer policies=], then [=prefetch record/cancel and discard=] |prefetchRecord| and return.
+    1. If |prefetchRecord|'s [=prefetch record/referrer policy=] is not in the [=list of sufficiently strict speculative navigation referrer policies=], then [=prefetch record/cancel and discard=] |prefetchRecord| given |document|, and return.
     1. Let |browsingContext| be |document|'s [=Document/browsing context=].
     1. Set |prefetchRecord|'s [=prefetch record/sandboxing flag set=] to the result of [=determining the creation sandboxing flags=] given |browsingContext| and |browsingContext|'s [=browsing context/container=].
     1. Let |isolationOrigin| be a new [=opaque origin=].
@@ -518,13 +518,13 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
     1. Let (|response|, |locationURL|) be the result of [=performing a common navigational fetch=] given |request|, "`other`", |browsingContext|, |isolatedEnvironment|, an empty algorithm, |shouldBlockNavigationRequest|, |shouldBlockNavigationResponse|, and |prefetchRecord|'s [=prefetch record/fetch controller=].
     1. If |locationURL| is failure or a [=URL=] whose [=url/scheme=] is not an [=HTTP(S) scheme=], then set |response| to a [=network error=].
     1. TODO: navigate-to, frame-src, XFO enforcement should probably be left to navigation, but what about Content-Disposition?
-    1. If |response| is a [=network error=], then [=prefetch record/cancel and discard=] |prefetchRecord| and return.
-    1. If |originsWithConflictingCredentials| is not empty, then [=prefetch record/cancel and discard=] |prefetchRecord| and return.
+    1. If |response| is a [=network error=], then [=prefetch record/cancel and discard=] |prefetchRecord| given |document|, and return.
+    1. If |originsWithConflictingCredentials| is not empty, then [=prefetch record/cancel and discard=] |prefetchRecord| given |document|, and return.
 
         <div class="note">This means that if any origin along the redirect chain had credentials, the prefetch is discarded. This reduces the chance of the user observing a logged-out page when they are logged in.</div>
     1. [=Assert=]: |response| is [=exchange record/response=] the last element of |prefetchRecord|'s [=prefetch record/redirect chain=].
-    1. If |response| does not [=support prefetch=], then [=prefetch record/cancel and discard=] |prefetchRecord| and return.
-    1. [=prefetch record/Complete=] |prefetchRecord|.
+    1. If |response| does not [=support prefetch=], then [=prefetch record/cancel and discard=] |prefetchRecord| given |document|, and return.
+    1. [=prefetch record/Complete=] |prefetchRecord| given |document|.
 
     <div class="issue">This ends up setting the `` `Cache-Control` `` and `` `Pragma` `` request headers, which is contrary to what Chromium does today when skipping cache here. One approach would be to add a flag similar to [=request/prevent no-cache cache-control header modification flag=]. It would also be possible to have a cache that can be copied into the ordinary cache, like Chromium does for cookies.</div>
 
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -136,7 +136,7 @@ A <dfn export>prefetch record</dfn> is a [=struct=] with the following [=struct/
 * <dfn export for="prefetch record">URL</dfn>, a [=URL=]
 * <dfn export for="prefetch record">referrer policy</dfn>, a [=referrer policy=]
 * <dfn export for="prefetch record">anonymization policy</dfn>, a [=prefetch IP anonymization policy=]
-* <dfn export for="prefetch record">initiator</dfn>, a {{DOMString}}
+* <dfn export for="prefetch record">label</dfn>, a [=string=]
 
   <div class="note">This is intended for use by a specification or [=implementation-defined=] feature to identify which prefetches it created. It might also associate other data with this struct.</div>
 * <dfn export for="prefetch record">state</dfn>, which is `"ongoing"` (the default), `"completed"`, or `"canceled"`
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -152,7 +152,7 @@ A <dfn export>prefetch record</dfn> is a [=struct=] with the following [=struct/
 
 A [=prefetch record=]'s <dfn export for="prefetch record">response</dfn> is the [=exchange record/response=] of the last element of its [=prefetch record/redirect chain=], or null if that list [=list/is empty=].
 
-The user agent may [=prefetch record/cancel and discard=] records from the [=Document/prefetch records=] even if they are not expired, e.g., due to resource constraints. Since completed records with expiry times in the past are never returned, they can be removed with no observable consequences.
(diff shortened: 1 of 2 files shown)


jeremyroman:
@domenic Does this make sense to you? This isn't implemented yet for prefetch (though it is for prerender) but seems a reasonably rational way to do it.

speculation-rules.bs
```diff
@@ -253,15 +268,22 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
...
+      1. &#x231B; [=list/Remove=] |startedPrefetch| from |document|'s [=Document/started prefetches=].
   1. End the [=synchronous section=], continuing the remaining steps [=in parallel=].
   1. [=list/For each=] |prefetchCandidate| of |prefetchCandidates|:
-    1. The user agent may [=prefetch=] given |document|, |prefetchCandidate|'s [=prefetch candidate/URL=], |prefetchCandidate|'s [=prefetch candidate/referrer policy=] and |prefetchCandidate|'s [=prefetch candidate/anonymization policy=].
+    1. The user agent may run the following steps:
+      1. Let |prefetchController| be a new [=prefetch controller=] whose [=prefetch controller/document=] is |document|.
+      1. [=list/Append=] a new [=started prefetch=] whose [=started prefetch/candidate=] is |prefetchCandidate| and [=started prefetch/controller=] is |prefetchController| to |document|'s [=Document/started prefetches=].
+      1. [=Prefetch=] given |document|, |prefetchCandidate|'s [=prefetch candidate/URL=], |prefetchCandidate|'s [=prefetch candidate/referrer policy=], |prefetchCandidate|'s [=prefetch candidate/anonymization policy=], and |prefetchController|.
```
domenic, line 280: It seems like it'd be cleaner if "prefetch" automatically updated the document's "started prefetches"... I guess that might make the document factoring and layering less good, though? Not sure...

prefetch.bs
```diff
@@ -225,19 +228,40 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
   1. Return false.
 </div>
 
+A <dfn export>prefetch controller</dfn> is a [=struct=] which allows managing a prefetch operation. It has the following [=items=]:
+* <dfn for="prefetch controller">state</dfn> (default `"ongoing"`), which is `"ongoing"` or `"aborted"`
+
+  <div class="note">The state is `"ongoing"` even after prefetch is complete, as long as it was not aborted by the author.</div>
```
domenic, line 234: I think "aborted by the author" is a bit confusing? E.g. can't the UA abort as well? Or even the user, perhaps, by unloading the page?

prefetch.bs
```diff
@@ -225,19 +228,40 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
...
+* <dfn for="prefetch controller">fetch controller</dfn> (default null), which is a [=fetch controller=] or null
+* <dfn for="prefetch controller">prefetch record</dfn> (default null), which is a [=prefetch record=] or nulll
+* <dfn export for="prefetch controller">document</dfn>, which is a {{Document}}
+
+  <div class="note">This is the {{Document}} whose [=prefetch buffer=] contains this's [=prefetch controller/prefetch record=], if any.
```
domenic, line 239:
```suggestion
  <div class="note">This is the {{Document}} whose [=prefetch buffer=] contains the [=prefetch controller/prefetch record=], if any.
```

("this" seems too JavaScript/Web IDL-ey to me)

prefetch.bs
```diff
@@ -225,19 +228,40 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
...
+
+<div algorithm>
+  To <dfn export for="prefetch controller">abort</dfn> a [=prefetch controller=] |prefetchController|:
```
domenic, line 242: Maybe name it "abort and discard" to make the behavior described in the note clearer.

prefetch.bs
```diff
@@ -225,19 +228,40 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
   1. Return false.
 </div>
 
+A <dfn export>prefetch controller</dfn> is a [=struct=] which allows managing a prefetch operation. It has the following [=items=]:
+* <dfn for="prefetch controller">state</dfn> (default `"ongoing"`), which is `"ongoing"` or `"aborted"`
```
domenic, line 232: This is only ever set, never consulted. So I think it can be removed?

prefetch.bs
domenic, line 239: The existence of both Document's "prefetch buffer" and Document's "started prefetches" is pretty confusing right now. Similarly having all four structures (prefetch controller, started prefetch, prefetch record, prefetch candidate) is a bit tricky. Especially since this is all spread across two specs.

Here is my attempt at summarizing the situation and suggesting improvements...

- "prefetch candidate": result of parsing speculation rules. Sometimes stored persistently inside a "started prefetch", but it doesn't have identity, we're just keeping the URL and anonymization policy (not the referrer policy?) around for later comparison.
   - Suggestion: flatten URL and anonymization policy into "started prefetch", so that "prefetch candidate" is always the result of parsing and never has persistent existence.
   - Suggestion: rename to "prefetch speculation candidate" to reflect the strong connection to speculation rules? Unsure.
- "prefetch record": represents an ongoing prefetch, including eventually the response. Lives in "prefetch buffer".
- "started prefetch": represents an ongoing or aborted prefetch, including enough information to identify it, and to control (abort) it. Lives in "started prefetches"
  - Suggestion: merge with prefetch record?
- "prefetch controller": controls the prefetch. Only stored in "started prefetch". However it also has a reference to a "prefetch record".
  - Suggestion: remove the reference to "prefetch record" so that this is more of a leaf structure. Then, "abort" needs to take a prefetch record/started prefetch.
  - Possible suggestion: merge into "prefetch record" since it's not clear the independent existence gives us much...


Suggested end result, if possible:

- "prefetch speculation candidate" contains (URL, anonymization policy, referrer policy) and represents the result of parsing speculation rules.
- "Prefetch" takes a "prefetch speculation candidate" and turns it into a "prefetch record", which contains (URL, referrer policy, sandbox flags present, redirect chain, expiry time, isolated partition key, document, state?, fetch controller). "Prefetch" puts the resulting prefetch record into the document's "prefetch records"
- "Abort and discard" operates on "prefetch record". It aborts the fetch and also removes the record from the document's "prefetch records".

WDYT?

domenic (commented):
A few minor comments... But there's one major one about consolidating the various structs and lists, which will probably make those minor comments obsolete.

prefetch.bs
jeremyroman: Did more or less this refactoring. Excellent suggestion, thank you.

prefetch.bs
jeremyroman: Used "cancel and discard" for consistency with fetch, which uses "abort" to mean not by the user/author, "terminate" to mean by the UA, and "cancel" to mean either.

prefetch.bs
jeremyroman: Its replacement is consulted.

prefetch.bs
jeremyroman: I think this is resolved in the refactor, lmk.

prefetch.bs
jeremyroman: Obsoleted by refactor.

speculation-rules.bs
jeremyroman: It now does so (well, each prefetch algorithm does so -- even though it's duplicated, it means that it would be possible to directly invoke either of them in the future).

jeremyroman:
PTAL

prefetch.bs
```diff
@@ -129,48 +130,73 @@ A <dfn>redirect chain</dfn> is a [=list=] of [=exchange records=].
...
 
-A <dfn>prefetch record</dfn> is a [=struct=] with the following [=struct/items=]:
+A <dfn export>prefetch record</dfn> is a [=struct=] with the following [=struct/items=]:
 * <dfn export for="prefetch record">URL</dfn>, a [=URL=]
 * <dfn export for="prefetch record">referrer policy</dfn>, a [=referrer policy=]
-* <dfn export for="prefetch record">sandbox flags present</dfn>, a boolean
-* <dfn export for="prefetch record">redirect chain</dfn>, a [=redirect chain=]
-* <dfn export for="prefetch record">expiry time</dfn>, a {{DOMHighResTimeStamp}}
+* <dfn export for="prefetch record">anonymization policy</dfn>, a [=prefetch IP anonymization policy=]
+* <dfn export for="prefetch record">initiator</dfn>, a {{DOMString}}
```
domenic, line 139: Nit, I prefer to use `[=string=]` when not at an IDL boundary. They are isomorphic so it doesn't matter, but it feels cleaner.

domenic, line 139: If this is not a https://fetch.spec.whatwg.org/#concept-request-initiator then it is a bad name.

prefetch.bs
```diff
@@ -129,48 +130,73 @@ A <dfn>redirect chain</dfn> is a [=list=] of [=exchange records=].
...
 
+<div class="note">This tracks prefetches from when they are started to when they are ultimately used or discarded. Consequently some of these fields are immutable, some pertain to the ongoing activity (like [=prefetch record/fetch controller=]), and some (like [=prefetch record/expiry time=]) are populated when the prefetch completes.</div>
+
 A [=prefetch record=]'s <dfn export for="prefetch record">response</dfn> is the [=exchange record/response=] of the last element of its [=prefetch record/redirect chain=], or null if that list [=list/is empty=].
 
-The user agent may remove elements from the [=prefetch buffer=] even if they are not expired, e.g., due to resource constraints. Since records with expiry times in the past are never returned, they can be removed with no observable consequences.
+The user agent may [=prefetch record/cancel and discard=] records from the [=Document/prefetch records=] even if they are not expired, e.g., due to resource constraints. Since completed records with expiry times in the past are never returned, they can be removed with no observable consequences.
```
domenic, line 155: "never returned" is a bit vague. Returned from where?

prefetch.bs
```diff
@@ -129,48 +130,73 @@ A <dfn>redirect chain</dfn> is a [=list=] of [=exchange records=].
...
 * <dfn export for="prefetch record">URL</dfn>, a [=URL=]
 * <dfn export for="prefetch record">referrer policy</dfn>, a [=referrer policy=]
-* <dfn export for="prefetch record">sandbox flags present</dfn>, a boolean
-* <dfn export for="prefetch record">redirect chain</dfn>, a [=redirect chain=]
-* <dfn export for="prefetch record">expiry time</dfn>, a {{DOMHighResTimeStamp}}
+* <dfn export for="prefetch record">anonymization policy</dfn>, a [=prefetch IP anonymization policy=]
+* <dfn export for="prefetch record">initiator</dfn>, a {{DOMString}}
+
+  <div class="note">This is intended for use by a specification or [=implementation-defined=] feature to identify which prefetches it created. It might also associate other data with this struct.</div>
+* <dfn export for="prefetch record">state</dfn>, which is `"ongoing"` (the default), `"completed"`, or `"canceled"`
```
domenic, line 142: Nit here and elsewhere, we use backticks inside the quote marks for spec string literals.

prefetch.bs
```diff
@@ -475,42 +504,42 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
...
-    1. If |response| is a [=network error=], then return.
-    1. If |originsWithConflictingCredentials| is not empty, then return.
+    1. If |response| is a [=network error=], then [=prefetch record/cancel and discard=] |prefetchRecord| given |document|, and return.
+    1. If |originsWithConflictingCredentials| is not empty, then [=prefetch record/cancel and discard=] |prefetchRecord| given |document|, and return.
 
         <div class="note">This means that if any origin along the redirect chain had credentials, the prefetch is discarded. This reduces the chance of the user observing a logged-out page when they are logged in.</div>
-    1. [=Assert=]: |response| is [=exchange record/response=] of |redirectChain|'s last element.
-    1. If |response| does not [=support prefetch=], then return.
-    1. [=Store a prefetch record=] given |document|, |url|, |referrerPolicy|, |sandboxFlags|, |redirectChain| and |isolatedPartitionKey|.
+    1. [=Assert=]: |response| is [=exchange record/response=] the last element of |prefetchRecord|'s [=prefetch record/redirect chain=].
```
domenic, line 525: I think you want to delete `[=exchange record/response=]` here?

prefetch.bs
```diff
@@ -129,48 +130,73 @@ A <dfn>redirect chain</dfn> is a [=list=] of [=exchange records=].
...
+* <dfn export for="prefetch record">anonymization policy</dfn>, a [=prefetch IP anonymization policy=]
+* <dfn export for="prefetch record">initiator</dfn>, a {{DOMString}}
+
+  <div class="note">This is intended for use by a specification or [=implementation-defined=] feature to identify which prefetches it created. It might also associate other data with this struct.</div>
+* <dfn export for="prefetch record">state</dfn>, which is `"ongoing"` (the default), `"completed"`, or `"canceled"`
+
+  <div class="note">`"canceled"` indicates that the prefetch was aborted by the author or user, or terminated by the user agent.</div>
+* <dfn export for="prefetch record">fetch controller</dfn>, a [=fetch controller=] (a new [=fetch controller=] by default)
+* <dfn export for="prefetch record">sandboxing flag set</dfn>, a [=sandboxing flag set=]
+* <dfn export for="prefetch record">redirect chain</dfn>, a [=redirect chain=] (empty by default)
```
domenic, line 147: I think you usually append raw responses to this, not exchange records. Not sure what the right fix is...

domenic (commented):
Looks really nice. The only potentially-big thing is the exchange record/redirect chain issue.

prefetch.bs

jeremyroman: It serves a similar purpose, and even Fetch already has two:

https://fetch.spec.whatwg.org/#concept-request-initiator
https://fetch.spec.whatwg.org/#request-initiator-type

I've renamed it "label" (every conceivable synonym is used somewhere in HTML or a related spec), but open to other names.

prefetch.bs

prefetch.bs
jeremyroman: Linked to "find a matching prefetch record".

prefetch.bs
jeremyroman: Ah, sorry for the gibberish. It should be the response of the last such element.

prefetch.bs
jeremyroman: I don't think so; can you point out where that happens? Initially exchange records without responses are appended using [=redirect chain/append=] and the responses are subsequently filled in with [=redirect chain/update the response=].

domenic: I think I was reading assuming that [=redirect chain/append=] was [=list/append=].

prefetch.bs
domenic: Wow, I never realized Fetch made that distinction. I don't know if I like it... but yeah, consistency is nice.

domenic (approved):
LGTM!

domenic:
We should really do something similar for prerender, but I'm unsure how much infrastructure to port... thoughts welcome.

Issue #168, opened 2022-08-09
Prerendering and non-current target windows


domenic:
In the Chromium implementation, we have accidentally created an architecture where it is hard to prerender a page without knowing what target window (or tab) it is going to be in. Currently we are always assuming that the prerender will be navigated to in the current window. This means that in scenarios like

```html
<script type=speculationrules>
{
  "prerender": [{
    "source": "list",
    "urls": ["page.html"]
  }]
}
</script>
<a target="_blank" href="page.html">click me</a>
```

the prerender will be discarded instead of activated. Instead, we would like to allow such scenarios to be prerendered.

Ideally, we would do the implementation work to allow prerendering a page without knowing its target window ahead of time. That is important for supporting cases like the user middle-clicking on a prerendered link, where we just don't know until the user clicks where the page will be opened. However, this is estimated to be an extremely large (~year-plus) project. So we are interested in coming up with something usable in the meantime, for sites that want to benefit from prerendering but have a lot of links they know will often or always be opened in a new tab.

There are basically two options I can think of here:

- Extend speculation rules to allow specifying the target window. E.g. `"urls": [{ "url": "page.html", "target": "_blank" }]`.

  - We could instead locate the target specification outside the URL list, assuming it applies to multiple URLs at once: i.e. `{ "source": "list", "target": "_blank", "urls": ["page.html"] }`. This might play well with the proposed [scores](https://github.com/WICG/nav-speculation/blob/main/triggers.md#rules), e.g. if a page suspects or has telemetry that the user will open in the same tab most of the time but a new tab sometimes, they could do `[{ "source": "list", "target": "_self", "urls": ["page.html"], score: "0.8" }, { "source": "list", "target": "_blank", "urls": ["page.html"], "score": 0.2 }]`.

  - We make this into a hint, e.g. with the name `target_hint`, instead of a constraint. Then in the future if Chromium no longer has the same architectural constraint, we could use the prerender regardless of the `target_hint` value.

  - We could deprecate and remove this over time (especially the hint variant) if Chromium manages to overhaul the architecture enough to support being agnostic.

[...]

Issue #167, opened 2022-08-03
Allow setting referrer policy in speculation rules


domenic:
We currently do not allow using a prerender or prefetch that has a mismatched referrer policy.

And, we always prerender or prefetch using the page's overall default referrer policy.

This means that if a site has something like this:

```html
<script type=speculationrules>
{
  "prerender": [{
    "source": "list",
    "urls": ["page.html"]
  }]
}
</script>
<a rel="noreferrer" href="page.html">click me</a>
```

then the prerender will be discarded instead of activated. This is correct behavior, but is unfortunate.

We could solve this with document rules. Or, we could allow setting the expected referrer policy in the list rules, e.g.

```json
{
  "prerender": [{
    "source": "list",
    "urls": ["page.html"],
    "referrer_policy": "no-referrer"
  }]
}
```

domenic:
An alternative was discussed in https://github.com/WICG/nav-speculation/issues/18, which is removing the requirement of referrer policy matching. There, we decided the reason we kept it was mostly to help in some cross-site cases related to cookies. I wonder if we should consider revisiting that decision, at least for same-site cases?

In particular, @domfarolino's survey in https://github.com/WICG/nav-speculation/issues/18#issuecomment-759647076 of how nothing else keys on referrer policy is pretty convincing...

domenic:
> An alternative was discussed in #18, which is removing the requirement of referrer policy matching.

This is not a full alternative, however. Consider a site which _does_ want to vary its content based on referrer. That site sends `Vary: referer`. Then, we really should require referrer matching.

For such sites, we need a way to indicate the referrer policy at prerendering time anyway. So, I think I am back to adding `"referrer_policy"` in the list rules.

jeremyroman:
Agreed we want syntax for it at least for list rules; in my mind that was always going to happen, just a question of priority. Failing that we should respect the attribute on the link (for document rules) and the document's default referrer policy.

I need to give it more thought, but I think I'm coming around on not keying on referrer policy at least if `Vary: Referer` isn't present. The arguments about how caches work are at least somewhat convincing, and we'd already previously talked about enforcing a minimum referrer policy.

toyoshim:
@domenic @jeremyroman hi, how is the status on this change? is there a specific blocker. or just isn't a high-priority item?
For a supportive input, I'm seeing a case that a major site hits this problem when I enforce the site to use prerendering.

kjmcnee:
[...]

PR #172, opened 2022-08-16
Explainer reorganization and rewrites

domenic:
The overall goal is to make it much clearer what we're specifying and shipping. Right now the explainers resemble historical strata, built on top of each other and containing old ideas that don't all match what we actually ended up doing.

So far I rewrote README.md and made some updates to triggers.md. This is enough to make me feel comfortable working on #167 and #168, so I might stop here, or I might keep going a bit---especially for prerendering, since it's also on my plate to specify/explainerize cross-origin same-site prerendering.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
-# Prerendering, revamped
+# Navigational speculation
 
-In order to making the experience of loading on the web faster, user agents employ prefetching and prerendering techniques. However, they have historically been [underspecified](https://w3c.github.io/resource-hints/#prerender) and [inconsistently implemented](https://caniuse.com/link-rel-prerender).
+[Draft specifications](https://wicg.github.io/nav-speculation/)
+
+In order to making the experience of navigating on the web faster, user agents employ prefetching and prerendering techniques. However, they have historically been [underspecified](https://w3c.github.io/resource-hints/#prerender) and [inconsistently implemented](https://caniuse.com/link-rel-prerender).
 
 The space is full of challenges which have not been comprehensively tackled: most notably,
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -9,150 +11,32 @@ The space is full of challenges which have not been comprehensively tackled: mos
...
-});
-```
+Our current proposals are focused around allowing cross-origin/site prefetching. We also have some early thoughts on cross-origin/site prerendering, but have not yet committed to them. The following explainers are relevant:
 
-See [the prerendering browsing context explainer](./browsing-context.md#javascript-apis) for an overview of all the different ways a site could react upon transitioning from prerendered to user-visible.
+* [Fetching modes](./fetch.md), discussing how to perform the fetch of the main content and any subresources while preloading.
+* [`Supports-Loading-Mode`](./opt-in.md), a new header which allows target pages to opt in to being loaded in an uncredentialed mode, with the understanding that they will upgrade their content later upon activation.
+* [`<meta http-equiv="supports-loading-mode">`](https://github.com/WICG/nav-speculation/blob/main/meta-processing.md), an extension to the header definition that uses a HTML preparsing pass to allow it to appear in-document.
+* [Clinet IP anonymization](./anonymous-client-ip.md), an extension to prefetching (and maybe one day prerendering?) to require using an anonymizing proxy to hide the user's IP address.
```
jeremyroman, line 29:
```suggestion
* [Client IP anonymization](./anonymous-client-ip.md), an extension to prefetching (and maybe one day prerendering?) to require using an anonymizing proxy to hide the user's IP address.
```

triggers.md
```diff
@@ -145,27 +199,37 @@ Another possible future extension, which would likely need to be restricted to s
...
+### More speculation actions
+
+As mentioned previously, we have currently only specified `"prefetch"` and `"prerender"` speculation actions.
+
+Adding `"dns-prefetch"` and `"preconnect"`, to mirror [Resource Hints](https://w3c.github.io/resource-hints/), would be an obvious extension, simply giving a more-ergonomic and capable way of triggering those actions.
+
+Another envisioned speculative action is `"prefetch-with-subresources"`, which prefetches a document and then uses the HTML preload scanner to find other subresources that are worth preloading. Chromium currently does something similar (known as "[NoState Prefetch](https://developer.chrome.com/blog/nostate-prefetch/)") for `<link rel="prerender">`. But, we're not yet sure this feature is pulling its weight, in between the lightweight prefetch and the fully-instant prerender features, so it's not yet clear whether this will be worth integrating.
```
jeremyroman, line 214:
```suggestion
Another envisioned speculative action is `"prefetch_with_subresources"`, which prefetches a document and then uses the HTML preload scanner to find other subresources that are worth preloading. Chromium currently does something similar (known as "[NoState Prefetch](https://developer.chrome.com/blog/nostate-prefetch/)") for `<link rel="prerender">`. But, we're not yet sure this feature is pulling its weight, in between the lightweight prefetch and the fully-instant prerender features, so it's not yet clear whether this will be worth integrating.
```

It was previously spelled with underscores [since it's a JSON key](https://w3ctag.github.io/design-principles/#casing-rules).

jeremyroman (approved):

domenic:
Oh no, rebase and merge is not enabled; my careful work to separate these into two commits is useless :). Oh well.

PR #173, opened 2022-08-17
Add "target_hint" to prerender speculation rules

### LINKED ISSUE #168: Prerendering and non-current target windows

domenic:
Closes #168.

### DIFF
--- a/triggers.md
+++ b/triggers.md
@@ -13,6 +13,7 @@
 - [The proposal](#the-proposal)
   - [List rules](#list-rules)
   - [Requirements](#requirements)
+  - [Window name targeting hints](#window-name-targeting-hints)
 - [Future extensions](#future-extensions)
   - [Scores](#scores)
   - [Document rules](#document-rules)
@@ -23,9 +24,11 @@
 - [Developer tooling](#developer-tooling)
 - [Feature detection](#feature-detection)
 - [Alternatives considered](#alternatives-considered)
-  - [General interop and compat concerns](#general-interop-and-compat-concerns)
-  - [Forward-compatibility problems](#forward-compatibility-problems)
-  - [Duplication](#duplication)
+  - [Extending the `<link>` element](#extending-the-link-element)
+    - [General interop and compat concerns](#general-interop-and-compat-concerns)
+    - [Forward-compatibility problems](#forward-compatibility-problems)
+    - [Duplication](#duplication)
+  - [Alternatives to `"target_hint"`](#alternatives-to-target_hint)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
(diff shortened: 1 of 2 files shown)


nhiroki (approved):
LGTM!

triggers.md
```diff
@@ -126,6 +129,45 @@ Currently the only requirement that is specified is denoted by `"anonymous-clien
...
+    "urls": ["page.html"]
+   },
+   {"source": "list",
+    "target_hint": "_blank",
+    "urls": ["page.html"]
+   }]
+}
+```
+
+However, in browsers such as Chromium that need the target hint, this will prerender the page twice, and thus use twice as many resources. So this is best avoided if possible.
```
jeremyroman, line 169: nit: I think technically "Chromium" is an open-source project, not a browser. Maybe "such as Chrome" or "in Chromium-based browsers".

jeremyroman (approved):

triggers.md
domenic: Nice catch; switching to "implementations".

PR #175, opened 2022-08-18
Remove keying on referrer policy

### LINKED ISSUE #18: What exactly are we trying to achieve with keying off of referrer policy?

domenic:
This also removes the `<link rel="prerender">` specification, since that was speculative and ended up not being implemented. And it solves the issue about prefetch/prerender candidates from speculation rules using hard-coded default referrer policies; instead they now use the document's referrer policy.

Some work probably remains about properly restricting preloading to sufficiently-strict referrer policies. Right now that is checked when going cross-origin, which might be enough?

Closes #18.

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -19,7 +19,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
     urlPrefix: browsers.html
       text: browsing context scope origin; url: browsing-context-scope-origin
-      text: container; for: browsing context; url: bc-container 
+      text: container; for: browsing context; url: bc-container
       text: determine the origin; url: determining-the-origin
     urlPrefix: browsing-the-web.html
       text: history handling behavior; url: history-handling-behavior
@@ -134,7 +134,6 @@ Each {{Document}} has <dfn export for="Document">prefetch records</dfn>, which i
 
 A <dfn export>prefetch record</dfn> is a [=struct=] with the following [=struct/items=]:
 * <dfn export for="prefetch record">URL</dfn>, a [=URL=]
-* <dfn export for="prefetch record">referrer policy</dfn>, a [=referrer policy=]
 * <dfn export for="prefetch record">anonymization policy</dfn>, a [=prefetch IP anonymization policy=]
 * <dfn export for="prefetch record">label</dfn>, a [=string=]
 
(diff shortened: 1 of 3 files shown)


prefetch.bs
```diff
@@ -381,8 +380,6 @@ These algorithms are based on [=process a navigate fetch=].
 
         :  [=request/URL=]
         :: |prefetchRecord|'s [=prefetch record/URL=]
-        :  [=request/referrer policy=]
-        :: |prefetchRecord|'s [=prefetch record/referrer policy=]
```
jeremyroman, line 385: Why remove this? I would expect that we will still want it to be able to have a value other than the document's default in the future, such as the mechanism in #167 and inferring from `<a referrerpolicy>`.

prerendering.bs
```diff
@@ -388,11 +346,11 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
 </div>
 
 <div algorithm="create a prerendering browsing context">
-  To <dfn export>create a prerendering browsing context</dfn> given a [=URL=] |startingURL|, a [=referrer policy=] |referrerPolicy|, and a {{Document}} |referrerDoc|:
```
jeremyroman, line 391: Similarly here, even if we're not going to key on it, it seems like we're going to want a policy plumbed through here.

prefetch.bs
domenic: Indeed, #167 might add something like this back. However, what we add back might not be the same, e.g. it'd probably be more like an optional referrer policy so that we get the correct fallback behavior if it's absent. So my plan was to remove for now to align with the implementation plans, and then evaluate the solution for #167, or for document rules, independently.

That said, I see there's a PR for document rules now, and I hope to work on #167 soon, so if we want to leave this un-merged and see how those shape up before merging, I'd be OK with that.

jeremyroman: It feels like it'd be pretty close to what's here now, except possibly making it optional (depending on whether we want the document's referrer policy at the point where the speculation is initiated or where the speculation is detected -- the two will seldom if ever differ). If you felt strongly about landing this as-is I think it'd be okay, but it just feels odd to remove and then re-add something very similar.

domenic: I'm getting pretty nervous about how far the explainers and specs are from what we've currently implemented, so I'd like to merge this and then rebuild it as appropriate :).

PR #181, opened 2022-09-01
Support same-site cross-origin prerendering

domenic:
This includes our first formal spec for Supports-Loading-Mode, and an update to its explainer to reflect what we're most immediately looking to ship.


prerendering.bs
```diff
@@ -528,11 +534,23 @@ Navigation redirects can also [=prerendering browsing context/activate=] [=prere
...
+
+  1. If |browsingContext| is a prerendering browsing context, and <var ignore>responseOrigin</var> is not [=same origin=] with |incumbentNavigationOrigin|, then:
+
+    1. Let |loadingModes| be the result of [=getting the supported loading modes=] for |response|.
+
+    1. If |loadingModes| does not [=list/contain=] \`<code><a for="Supports-Loading-Mode">credentialed-prerender</a></code>\`, then set |response| to a [=network error=].
+
+       <p class="note">This will cause the prerender to fail in the following steps of the algorithm.</p>
+
+       <p class="XXX">In the future we could possibly also allow the `uncredentialed-prerender` token to work here. However, since that is primarily intended for the cross site case, which isn't specified or implemented yet, we don't want to encourage its use in the wild yet, so implementations must reject the `uncredentialed-prerender` token for now.</p>
```
nhiroki, line 553: Just to confirm: Is `class="XXX"` the correct identifier?

nhiroki (approved):
LGTM!

opt-in.md
```diff
@@ -1,96 +1,115 @@
...
-  <dd>The resource is suitable for any user who meets the cache conditions. Either it is not personalized, or it includes script to modify the document to reflect credentialed state (e.g., to show the user's logged-in state). However, access to credentials will be available when the document loads and script executes.</dd>
+### Cross-origin same-site prerendering
 
-  <dt><code>uncredentialed-prerender</code></dt>
-  <dd>Implies <code>uncredentialed-prefetch</code>. The resource can be loaded without access to its credentials and storage. Access to certain other APIs may be limited in this state. Either it is not personalized based on credentials, or it includes script to modify the document when the loading state changes to permit it.</dd>
-</dl>
+For prerendering, there is an additional complication when dealing with cross-origin destinations. Because the web's privacy boundary is site, we can send credentials and give storage access to such pages before activation, with no privacy issues. However, the web's _security_ boundary is the origin, not the site. So we need to ensure that pages cannot use prerendering to attack other origins—even same-site ones.
```
jeremyroman, line 55: "with no privacy issues" feels like a stronger statement than we need. I think we can just omit that phrase because the intent is implied by the rest of the sentence

opt-in.md
```diff
@@ -1,96 +1,115 @@
...
-This means that the supported loading modes, if not declared in a response header, can be statically computed with use of an HTML parser without rendering or script execution.
-See [the meta processing model](meta-processing.md) for details.
+As a concrete example of what an "attack" might look like, consider an organization which provides both web hosting and document editing services. Furthermore, assume that whenever you access a document on the document editing service, it updates the user's list of recently-viewed documents. And finally, assume that the document editing service is not prerendering-aware, i.e. it has no code that checks [`document.prerendering`](./prerendering-state.md) to customize its behavior. Then, if we allowed `https://sites.example.com/bobs-star-wars-fan-site` to prerender `https://docs.example.com/public-list-of-best-star-wars-movies` with no opt-in, then Bob's Star Wars Fan Site would be able to update the user's list of recently-viewed documents, pushing the List of Best Star Wars Movies document to the top.
```
jeremyroman, line 57: Hmmmmm, this depends on whether this change is made as part of the script execution or as part of serving the response. In the latter case using something like an `<iframe>` or a non-CORS `fetch` would do the trick, as would a prefetch, unless the server was checking the request headers.

Not a problem per se, but kinda muddies the waters about why this same argument doesn't apply there.

prerendering-same-site.md
```diff
@@ -90,7 +92,7 @@ The [speculation rules API](./triggers.md) can be used to trigger a prerender to
 <a href="https://a.test/foo">Click me!</a>
 ```
 
-The `<script type="speculationrules">` block here hints to prerender `https://a.test/foo`. As in the previous section, such prerendering includes sending the `Sec-Purpose` header.
+The `<script type="speculationrules">` block here hints to prerender `https://a.test/foo`. As in the previous section, such prerendering includes sending the `Sec-Purpose` header. If the prerendered URL (or ultimate response URL, after redirects) is cross-site, then prerendering will fail. If it is same-site but cross-origin, then the destination needs the opt-in [`Supports-Loading-Mode: credentialed-prerender` header](./opt-in.md#cross-origin-same-site-prerendering).
```
jeremyroman, line 95: For the first sentence, I would expect the cross-site URL restriction to apply to not just the first and last URLs in the redirect chain, but all of them. (Of course, it's probably uncommon to redirect cross-site and back again, but I live for edge cases.)

prerendering-same-site.md
```diff
@@ -254,21 +264,37 @@ A prerendered `Document` can apply CSP to itself as normal. Being in a prerender
 
 Prerendered content will be affected by [`prefetch-src`](https://w3c.github.io/webappsec-csp/#directive-prefetch-src) on the referring page, which provides a way of preventing prefetching in addition to the [triggers](./triggers.md).
 
+## More details on cross-origin same-site
```
jeremyroman, line 267: maybe link to here from the section I was a little unsure about above?

jeremyroman (approved):

prerendering.bs
domenic: I think it is OK (it basically means "TODO", and generates a red box like you can see in https://html.spec.whatwg.org/#navigating-across-documents:shared-declarative-refresh-steps ), but on second glance it might be nicer as a `class="note"`.

opt-in.md
domenic: Yeah, I was unsure how much detail to go into here. The other document mentions

> Side effects from simply issuing credentialed requests to the target URL, e.g. triggering non-idempotent GETs, are not a concern. These are already possible today via `<iframe>`s, ``s, etc.

which is relevant to what you're saying.

Maybe I'll tweak the wording a bit to try to make it clearer I am envisioning a client-side update...

prerendering-same-site.md
domenic: Yeah I can see how this is confusing. Let me delete the parenthetical, so we just talk about "prerendered URL" and ignore redirects in this section. Discussion of them can then be done at leisure in the dedicated cross-origin same-site section below.

PR #182, opened 2022-09-06
Describe how client hints should work in prerendering context

Clqsin45:
Context:
Since Accept-CH cache is a global instance, prerender pages should not be able to modify it until users activate these pages.

The proposal is that prerender pages maintain their own caches, and use them to:

- store the opt-in sent with prerender repopnse,  
- update the following prerender fetch requests by appending opt-in headers  

Changes to Accept-CH cache will be exected upon prerender activation.

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -696,8 +696,10 @@ To <dfn>add a new prerender Accept-CH cache entry</dfn> given a [=prerendering b
 <div algorithm="append client hints to request patch">
   Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>append client hints to request</a> algorithm, by prepending the following steps before iterating over the |hintSet|:
 
-  1. Let |browsingContext| be |settingsObject|’s [=environment settings object/global object=]'s [=Window/browsing context=].
-    If |browsingContext| is a [=prerendering browsing context=], retrieve |prerender client hint set| in [=prerendering browsing context/prerender-scoped Accept-CH cache=], and then update the |hintSet| with |prerender client hint set|.
+  1. Let |browsingContext| be <var ignore>settingsObject</var>’s [=environment settings object/global object=]'s [=Window/browsing context=].
+    If |browsingContext| is a [=prerendering browsing context=], then:
+      1. Let |origin| be <var ignore>request</var>'s [=request/url=]'s [=url/origin=].
+      1. Set |hintSet| to the [=set/union=] of |hintSet| and |browsingContext|'s [=prerendering browsing context/prerender-scoped Accept-CH cache=][|origin|].
   
 </div>
 
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -696,7 +696,7 @@ To <dfn>add a new prerender Accept-CH cache entry</dfn> given a [=prerendering b
 <div algorithm="append client hints to request patch">
   Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>append client hints to request</a> algorithm, by prepending the following steps before iterating over the |hintSet|:
 
-  1. Let |browsingContext| be |settingsObject|’s [=global object=]'s [=browsing context=].
+  1. Let |browsingContext| be |settingsObject|’s [=environment settings object/global object=]'s [=Window/browsing context=].
     If |browsingContext| is a [=prerendering browsing context=], retrieve |prerender client hint set| in [=prerendering browsing context/prerender-scoped Accept-CH cache=], and then update the |hintSet| with |prerender client hint set|.
   
 </div>
@@ -705,8 +705,8 @@ To <dfn>add a new prerender Accept-CH cache entry</dfn> given a [=prerendering b
   Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>update the Client Hints set</a> algorithm, by updating the last step.
 
   1. If |browsingContext| is a [=prerendering browsing context=],
-    [=add a new prerender Accept-CH cache entry=] with |response|'s [=url/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
-    Otherwise, [=Add a new Accept-CH cache entry=] with |response|'s [=url/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
+    [=add a new prerender Accept-CH cache entry=] with |response|'s [=response/url=]'s [=url/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
+  1. Else [=add a new Accept-CH cache entry=] with |response|'s [=response/url=]'s [=url/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
 
 </div>
 
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -206,6 +206,10 @@ spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
     for: structured header
       text: list; url: name-lists
       text: token; url: name-tokens
+spec: client-hints-infrastructure; urlPrefix: https://wicg.github.io/client-hints-infrastructure
+  type: dfn
+    text: Accept-CH cache; url: accept-ch-cache
+    text: append client hints to request; url: abstract-opdef-append-client-hints-to-request
 </pre>
 <pre class="biblio">
 {
@@ -657,6 +661,22 @@ Extend the {{PerformanceNavigationTiming}} interface as follows:
 The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> getter steps are:
 
 1. Return a {{DOMHighResTimeStamp}} with a time value equal to the [=current document=]'s [=Document/activation start time=].
+
+<h3 id="client hint cache">Interaction with Client Hint Cache</h3>
+
+We need to ensure that the [=Accept-CH cache=], which is owned by user agent as a global storage, is not modified while prerendering, but is properly updated after activation. The following modifications ensure this.
+
+Each [=prerendering browsing context=] has a <dfn for="prerendering browsing context">prerender-scoped Accept-CH cache</dfn>, which is an [=ordered map=] of [=origin=] to [=client hints sets=].
+
+This stores which client hints each origin has opted into receiving, until it can be copied to the global [=Accept-CH cache=] when [=prerendering browsing context/finalize activation|activation is finalized=].
+
+<div algorithm="append client hints to request patch">
+  Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>append client hints to request</a> algorithm, by prepending the following steps before iterating over the |hintSet|:
+
+  1. Let |browsingContext| be |settingsObject|’s [=global object=]'s [=browsing context=].
+    If |browsingContext| is a [=prerendering browsing context=], retrieve |prerender client hint set| in [=prerendering browsing context/prerender-scoped Accept-CH cache=], and then update the |hintSet| with |prerender client hint set|.
+  
+</div>
 
 <h2 id="supports-loading-mode">The \`<dfn http-header><code>Supports-Loading-Mode</code></dfn>\` HTTP response header</h2>
 
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -460,6 +460,10 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
 
     1. If |doc|'s [=Document/origin=] is the same as |origin|, then set |doc|'s [=Document/activation start time=] to the [=relative high resolution time=] for |unsafe time| and |doc|.
 
+    1. [=map/For each=] |origin| → |hintSet| in |bc|'s [=prerendering browsing context/prerender-scoped Accept-CH cache=]:
+      
+      1. [=Add a new Accept-CH cache entry=] to the [=Accept-CH cache=] with the |origin| and |hintSet|.
+
     1. [=Fire an event=] named {{Document/prerenderingchange}} at |doc|.
 
     1. [=list/For each=] |steps| in |doc|'s [=Document/post-prerendering activation steps list=]:
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -673,7 +673,7 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
 
 1. Return a {{DOMHighResTimeStamp}} with a time value equal to the [=current document=]'s [=Document/activation start time=].
 
-<h3 id="client hint cache">Interaction with Client Hint Cache</h3>
+<h3 id="client-hint-cache">Interaction with Client Hint Cache</h3>
 
 We need to ensure that the [=Accept-CH cache=], which is owned by user agent as a global storage, is not modified while prerendering, but is properly updated after activation. The following modifications ensure this.
 
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -209,7 +209,14 @@ spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
 spec: client-hints-infrastructure; urlPrefix: https://wicg.github.io/client-hints-infrastructure
   type: dfn
     text: Accept-CH cache; url: accept-ch-cache
+    text: add a new Accept-CH cache entry; url: add-a-new-accept-ch-cache-entry
     text: append client hints to request; url: abstract-opdef-append-client-hints-to-request
+    text: client hints token; url: client-hints-token-definition
+    text: client hints set; url: client-hints-set
+    text: update the Client Hints set; url: abstract-opdef-update-the-client-hints-set
+    for: environment settings object
+      text: client hints set; url: environment-settings-object-client-hints-set
+
 </pre>
 <pre class="biblio">
 {
@@ -670,12 +677,33 @@ Each [=prerendering browsing context=] has a <dfn for="prerendering browsing con
 
 This stores which client hints each origin has opted into receiving, until it can be copied to the global [=Accept-CH cache=] when [=prerendering browsing context/finalize activation|activation is finalized=].
 
+<div algorithm>
+
+To <dfn>add a new prerender Accept-CH cache entry</dfn> given a [=prerendering browsing context=] |bc|, an [=origin=] |origin|, and a [=client hints set=] |hintSet|:
+
+  1. Let |cache| be |bc|'s [=prerendering browsing context/prerender-scoped Accept-CH cache=].
+
+  1. [=map/set=] |cache|[|origin|] to |hintSet|.
+
+
+<p class="note">This is similar to [=add a new Accept-CH cache entry=].</p>
+</div>
+
 <div algorithm="append client hints to request patch">
   Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>append client hints to request</a> algorithm, by prepending the following steps before iterating over the |hintSet|:
 
   1. Let |browsingContext| be |settingsObject|’s [=global object=]'s [=browsing context=].
     If |browsingContext| is a [=prerendering browsing context=], retrieve |prerender client hint set| in [=prerendering browsing context/prerender-scoped Accept-CH cache=], and then update the |hintSet| with |prerender client hint set|.
   
+</div>
+
+<div algorithm="update the Client Hints set patch">
+  Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>update the Client Hints set</a> algorithm, by updating the last step.
+
+  1. If |browsingContext| is a [=prerendering browsing context=],
+    [=add a new prerender Accept-CH cache entry=] with |response|'s [=url/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
+    Otherwise, [=Add a new Accept-CH cache entry=] with |response|'s [=url/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
+
 </div>
 
 <h2 id="supports-loading-mode">The \`<dfn http-header><code>Supports-Loading-Mode</code></dfn>\` HTTP response header</h2>
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -706,9 +706,10 @@ To <dfn>add a new prerender Accept-CH cache entry</dfn> given a [=prerendering b
 <div algorithm="update the Client Hints set patch">
   Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>update the Client Hints set</a> algorithm, by updating the last step.
 
-  1. If |browsingContext| is a [=prerendering browsing context=],
-    [=add a new prerender Accept-CH cache entry=] with |response|'s [=response/url=]'s [=url/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
-  1. Else [=add a new Accept-CH cache entry=] with |response|'s [=response/url=]'s [=url/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
+  1. Let |origin| be <var ignore>response</var>'s [=response/url=]'s [=url/origin=].
+  1. Let |hintSet| be <var ignore>settingsObject</var>'s [=environment settings object/client hints set=].
+  1. If |browsingContext| is a [=prerendering browsing context=], [=add a new prerender Accept-CH cache entry=] with |browsingContext|, |origin| and |hintSet|.
+  1. Otherwise, [=add a new Accept-CH cache entry=] with |origin| and |hintSet|.
 
 </div>
 


prerendering.bs
```diff
@@ -658,6 +673,33 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
 
 1. Return a {{DOMHighResTimeStamp}} with a time value equal to the [=current document=]'s [=Document/activation start time=].
 
+<h3 id="client hint cache">Interaction with Client Hint Cache</h3>
+
+[=Accept-CH cache=], which is owned by user agent as a global storage, should not be modified by [=prerendering browsing context=].
```
jeremyroman, line 678: I don't understand what this means as normative text. Is it merely informative text to explain the following?

prerendering.bs
```diff
@@ -658,6 +673,33 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+
+Instead, each [=prerendering browsing context=] maintains its <dfn>prerender scoped Accept-CH cache</dfn>, to persist the origin's opt-in to receive the indicated header fields.
```
jeremyroman, line 680: Maybe this should be a `<dfn for="prerendering browsing context">`? Also, what type of value is this?

prerendering.bs
```diff
@@ -658,6 +673,33 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+<h3 id="client hint cache">Interaction with Client Hint Cache</h3>
+
+[=Accept-CH cache=], which is owned by user agent as a global storage, should not be modified by [=prerendering browsing context=].
+
+Instead, each [=prerendering browsing context=] maintains its <dfn>prerender scoped Accept-CH cache</dfn>, to persist the origin's opt-in to receive the indicated header fields. 
+
+Similar to <a spec=CLIENT-HINTS-INFRASTRUCTURE> add a new Accept-CH cache entry</a>,
+to <dfn>add a new prerender Accept-CH cache entry</dfn> to the [=prerender scoped Accept-CH cache=],
+given an |origin| and a [=client hints set=] |hintSet|,
+[=map/set=] [=prerender scoped Accept-CH cache=][|origin|] to |hintSet|.
```
jeremyroman, line 685: Even though it's trivial this might be clearer to do in a `<div algorithm>`. I'm also confused as to why an algorithm var is linked to another dfn.

prerendering.bs
```diff
@@ -658,6 +673,33 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+  Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>append client hints to request</a> algorithm, by prepending the following steps before iterating over the |hintSet|:
+
+  1. If [=browsing context=] is a [=prerendering browsing context=], retrieve |prerender client hint set| in [=prerender scoped Accept-CH cache=], and then update the |hintSet| with |prerender client hint set|.
+  
+</div>
+
+<div algorithm="update the Client Hints set patch">
+  Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>update the Client Hints set</a> algorithm, by updating the last step.
+
+  1. If [=browsing context=] is a [=prerendering browsing context=],
```
jeremyroman, line 697: This should be the `|browsingContext|` var, not the `[=browsing context=]` dfn, right?

jeremyroman (commented):
@domenic is probably a better reviewer for prerendering BC text than me, but noticed this while passing by

prerendering.bs
```diff
@@ -449,6 +460,10 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
 
     1. If |doc|'s [=Document/origin=] is the same as |origin|, then set |doc|'s [=Document/activation start time=] to the [=relative high resolution time=] for |unsafe time| and |doc|.
 
+    1. [=map/For each=] |entry| in [=prerender scoped Accept-CH cache=]:
```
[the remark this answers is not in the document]
Clqsin45: Thanks! I see, it should be the prerender browsing context (bc)'s property; revised.

(Thanks for the link!)

prerendering.bs
```diff
@@ -604,7 +619,7 @@ Patch the [=process a navigate fetch=] algorithm to allow delaying or discarding
 
     1. Assert: |historyHandling| is not "`entry update`", since prerendering browsing contexts have trivial session histories and thus will never end up [=traverse the history|traversing=] back to an entry with null [=session history entry/document=].
 
-    1. If |historyHandling| is "`default`", then set |historyHandling| to "`replace`".
+    1. If |historyHandling| is "`default`", the=entryset |historyHandling| to "`replace`".
```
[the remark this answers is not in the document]
Clqsin45: oops.. reverted

prerendering.bs
Clqsin45: Thanks! I updated its type. 

(BTW: Sorry for the stupid question. I'd like to understand when do we usually use "for=xxx". I guess we use "for" when a property belongs to another? It might be similar to a class member from programming's perspective. Is my understanding correct?

prerendering.bs
Clqsin45: indeed . 👍

prerendering.bs
Clqsin45: I meant to say that prerender browsing context should not modify Accept-CH cache, so it makes changes to its own cache before being activated.

Should I include this information in normatice text? I removed it for now.

prerendering.bs
Clqsin45: Thanks! I moved it into a <div algorithm>.

Do you mean that I should not use `Similar to <a spec=CLIENT-HINTS-INFRASTRUCTURE> add a new Accept-CH cache entry</a>,`?
Sorry, I haven't figured out whether I should use <a spec=""> or not.. Could you help point me to the instructions?

Clqsin45:
Thanks for your comments!

prerendering.bs
```diff
@@ -658,6 +673,37 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
 
 1. Return a {{DOMHighResTimeStamp}} with a time value equal to the [=current document=]'s [=Document/activation start time=].
 
+<h3 id="client hint cache">Interaction with Client Hint Cache</h3>
+
+Each [=prerendering browsing context=] maintains its <dfn for="prerendering browsing context">prerender scoped Accept-CH cache</dfn>, to persist the origin's opt-in to receive the indicated header fields.
+
+[=prerendering browsing context/Prerender scoped Accept-CH cache=] is an [=ordered map=] of [=origin=] to [=client hints set=].
```
domenic, line 680:
```suggestion
The [=prerendering browsing context/prerender scoped Accept-CH cache=] is an [=ordered map=] of [=origins=] to [=client hints sets=].
```

prerendering.bs
```diff
@@ -658,6 +673,37 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
 
 1. Return a {{DOMHighResTimeStamp}} with a time value equal to the [=current document=]'s [=Document/activation start time=].
 
+<h3 id="client hint cache">Interaction with Client Hint Cache</h3>
+
+Each [=prerendering browsing context=] maintains its <dfn for="prerendering browsing context">prerender scoped Accept-CH cache</dfn>, to persist the origin's opt-in to receive the indicated header fields.
```
domenic, line 678: Nit: "prerender-scoped" is a bit more grammatical than "prerender scoped"

prerendering.bs
```diff
@@ -658,6 +673,37 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+
+[=prerendering browsing context/Prerender scoped Accept-CH cache=] is an [=ordered map=] of [=origin=] to [=client hints set=].
+
+<div algorithm="add a new prerender Accept-CH cache entry">
+
+Similar to [=add a new Accept-CH cache entry=], to <dfn>add a new prerender Accept-CH cache entry</dfn> to the [=prerendering browsing context/prerender scoped Accept-CH cache=],
+given an |origin| and a [=client hints set=] |hintSet|,
+[=map/set=] [=prerendering browsing context/prerender scoped Accept-CH cache=][|origin|] to |hintSet|.
```
domenic, line 686: The prerender scoped Accept-CH cache belongs to a given prerendering browsing context. So it needs to be referenced as `|bc|'s [=...=]`. Which probably means this algorithm needs to take a prerendering browsing context.

prerendering.bs
```diff
@@ -658,6 +673,37 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+
+</div>
+
+<div algorithm="append client hints to request patch">
+  Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>append client hints to request</a> algorithm, by prepending the following steps before iterating over the |hintSet|:
+
+  1. Let |browsingContext| be |settingsObject|’s [=global object=]'s [=browsing context=].
```
domenic, line 693:
```suggestion
  1. Let |browsingContext| be |settingsObject|’s [=environment settings object/global object=]'s [=Window/browsing context=].
```

domenic: Hmm, this reveals a potential issue. What if settings object is for a worker, instead of a window?

I think we avoid this issue in the current prerendering spec by delaying all worker loads. So I think we can just bail out if `|settingsObject|'s [=environment settings object/global object=]` is not a `{{Window}}` object.

prerendering.bs
```diff
@@ -658,6 +673,37 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+    If |browsingContext| is a [=prerendering browsing context=], retrieve |prerender client hint set| in [=prerendering browsing context/prerender scoped Accept-CH cache=], and then update the |hintSet| with |prerender client hint set|.
```
domenic, line 694: I think you can replace the "retrieve ... and then update" step with something like

```
Set |hintSet| to the [=set/union=] of |hintSet| and |browsingContext|'s [=prerendering browsing context/prerender scoped Accept-CH cache=][|origin|].
```

However you'll need to get `|origin|` from somewhere... maybe from `|request|`?

prerendering.bs
```diff
@@ -658,6 +673,37 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+<div algorithm="add a new prerender Accept-CH cache entry">
+
+Similar to [=add a new Accept-CH cache entry=], to <dfn>add a new prerender Accept-CH cache entry</dfn> to the [=prerendering browsing context/prerender scoped Accept-CH cache=],
+given an |origin| and a [=client hints set=] |hintSet|,
+[=map/set=] [=prerendering browsing context/prerender scoped Accept-CH cache=][|origin|] to |hintSet|.
+
+</div>
+
+<div algorithm="append client hints to request patch">
+  Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>append client hints to request</a> algorithm, by prepending the following steps before iterating over the |hintSet|:
```
domenic, line 691: Good work adding this before the iteration step. That ensures all the complicated logic about what to skip and modify still works.

prerendering.bs
```diff
@@ -658,6 +673,37 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+  1. Let |browsingContext| be |settingsObject|’s [=global object=]'s [=browsing context=].
+    If |browsingContext| is a [=prerendering browsing context=], retrieve |prerender client hint set| in [=prerendering browsing context/prerender scoped Accept-CH cache=], and then update the |hintSet| with |prerender client hint set|.
+  
+</div>
+
+<div algorithm="update the Client Hints set patch">
+  Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>update the Client Hints set</a> algorithm, by updating the last step.
+
+  1. If |browsingContext| is a [=prerendering browsing context=],
+    [=add a new prerender Accept-CH cache entry=] with |response|'s [=url/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
```
domenic, line 702:
```suggestion
    [=add a new prerender Accept-CH cache entry=] with |response|'s [=response/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
```

prerendering.bs
```diff
@@ -658,6 +673,37 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+    Otherwise, [=Add a new Accept-CH cache entry=] with |response|'s [=url/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
```
domenic, line 703:
```suggestion
    Otherwise, [=Add a new Accept-CH cache entry=] with |response|'s [=response/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
```

prerendering.bs
domenic: I think it's useful to state this as a [statement of fact](https://ln.hixie.ch/?start=1140242962&count=1), but indeed, stating it as a normative requirement using "must" and "should not" is not correct. So something like

> We need to ensure that the Accept-CH cache, which his owned..., is not modified while prerendering, but is properly updated after activation. The following modifications ensure this:

prerendering.bs
domenic: Yes, your understanding is correct!

One detail is that it's not always mandatory in spec-writing, since the for="" value is only used for disambiguation. But I think it's always good practice to include it, even if there is no other type of object that has a property named "prerender scoped Accept-CH cache". It makes it clearer to people reading the spec source code, and protects you from any future collision.

domenic (commented):
This looks great. The only substantial issue is around being more precise on where and how you retrieve the prerender scoped Accept-CH cache, and client hint sets from it.

prerendering.bs
Clqsin45: use |request|'s for now.

The setting object also has an orign field. Which one would be better?
https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-origin?

prerendering.bs
Clqsin45: renamed it and referrences below, Thanks!

prerendering.bs
Clqsin45: Revised. Thanks! 
(Thanks for the link! Now I understnad why I should not use "should" here :)) )

prerendering.bs
Clqsin45: Thank you, Domenic! 

I see. Yeah, when I read the specification, I find it is very helpful! I'll be careful when I write specificaiotn :D

prerendering.bs
Clqsin45: Indeed. 👍  Fixed.

Clqsin45:
Thank you for reviewing :D

prerendering.bs
```diff
@@ -658,6 +673,41 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
 
 1. Return a {{DOMHighResTimeStamp}} with a time value equal to the [=current document=]'s [=Document/activation start time=].
 
+<h3 id="client hint cache">Interaction with Client Hint Cache</h3>
+
+We need to ensure that the [=Accept-CH cache=], which is owned by user agent as a global storage, is not modified while prerendering, but is properly updated after activation. The following modifications ensure this:
```
jeremyroman, line 678: nit: I'd probably just use a period (`.`) instead of a colon (`:`) here, since it's really introducing the entire section and not a bulleted list.

prerendering.bs
```diff
@@ -658,6 +673,41 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+
+Each [=prerendering browsing context=] maintains its <dfn for="prerendering browsing context">prerender-scoped Accept-CH cache</dfn>, to persist the origin's opt-in to receive the indicated header fields.
+
+The [=prerendering browsing context/prerender-scoped Accept-CH cache=] is an [=ordered map=] of [=origins=] to [=client hints sets=].
```
jeremyroman, line 682: nit: of...to is slightly awkward, maybe `an [=ordered map=] from [=origin=] to [=client hint set=]`, or possibly the very verbose `an [=ordered map=] with [=map/keys=] which are [=origins=] and [=map/values=] which are [=client hint sets=].`

prerendering.bs
```diff
@@ -658,6 +673,41 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+
+<div algorithm="add a new prerender Accept-CH cache entry">
```
jeremyroman, line 684: nit: if there's only one dfn in the div, the name of the algorithm is automatically inferred, so you can just write `<div algorithm>`

prerendering.bs
```diff
@@ -658,6 +673,41 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+
+Similar to [=add a new Accept-CH cache entry=], to <dfn>add a new prerender Accept-CH cache entry</dfn> to the [=prerendering browsing context/prerender-scoped Accept-CH cache=],
+given a [=prerendering browsing context=] |bc|, an |origin| and a [=client hints set=] |hintSet|,
```
jeremyroman, line 687: nit: type of `|origin|` missing

prerendering.bs
```diff
@@ -658,6 +673,41 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+
+We need to ensure that the [=Accept-CH cache=], which is owned by user agent as a global storage, is not modified while prerendering, but is properly updated after activation. The following modifications ensure this:
+
+Each [=prerendering browsing context=] maintains its <dfn for="prerendering browsing context">prerender-scoped Accept-CH cache</dfn>, to persist the origin's opt-in to receive the indicated header fields.
+
+The [=prerendering browsing context/prerender-scoped Accept-CH cache=] is an [=ordered map=] of [=origins=] to [=client hints sets=].
+
+<div algorithm="add a new prerender Accept-CH cache entry">
+
+Similar to [=add a new Accept-CH cache entry=], to <dfn>add a new prerender Accept-CH cache entry</dfn> to the [=prerendering browsing context/prerender-scoped Accept-CH cache=],
```
jeremyroman, line 686: nit: I would split out the assertion about similarity into a separate informative note. I personally mildly prefer separating the steps from the line that defines the parameters (even if there's only one step), but this might be a matter of taste.

nit: another matter of taste, but trying to figure out whether the indexing syntax binds more tightly than the `'s` took me a second; splitting it into two steps seems to me to make it easier -- but it's probably fine not to, if you'd prefer

```
To <dfn>add a new prerender Accept-CH cache entry</dfn> given a [=prerendering browsing context=] |bc|, an [=origin=] |origin|, and a [=client hints set=] |hintSet|:

1. Let |cache| be |bc|'s [=prerendering browsing context/prerender-scoped Accept-CH cache=].
1. Set |cache|[|origin|] to |hintSet|.
```

prerendering.bs
```diff
@@ -658,6 +673,41 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+[=map/set=] |bc|'s [=prerendering browsing context/prerender-scoped Accept-CH cache=][|origin|] to |hintSet|.
+
+</div>
+
+<div algorithm="append client hints to request patch">
+  Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>append client hints to request</a> algorithm, by prepending the following steps before iterating over the |hintSet|:
+
+  1. Let |browsingContext| be |settingsObject|’s [=environment settings object/global object=]'s [=Window/browsing context=].
+    If |browsingContext| is a [=prerendering browsing context=], then:
+      1. Let |orign| be |request|'s [=request/url=]'s [=url/origin=].
```
jeremyroman, line 697: sp: `|orign|` -> `|origin|`

prerendering.bs
```diff
@@ -658,6 +673,41 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+      1. Set |hintSet| to the [=set/union=] of |hintSet| and |browsingContext|'s [=prerendering browsing context/prerender-scoped Accept-CH cache=][|origin|].
+  
+</div>
+
+<div algorithm="update the Client Hints set patch">
+  Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>update the Client Hints set</a> algorithm, by updating the last step.
+
+  1. If |browsingContext| is a [=prerendering browsing context=],
+    [=add a new prerender Accept-CH cache entry=] with |response|'s [=response/url=]'s [=url/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
+  1. Else [=add a new Accept-CH cache entry=] with |response|'s [=response/url=]'s [=url/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
```
jeremyroman, line 707: nit: unlike programming languages, specs seem to usually say "Otherwise," instead of "Else"

prerendering.bs
```diff
@@ -658,6 +673,41 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
 
 1. Return a {{DOMHighResTimeStamp}} with a time value equal to the [=current document=]'s [=Document/activation start time=].
 
+<h3 id="client hint cache">Interaction with Client Hint Cache</h3>
+
+We need to ensure that the [=Accept-CH cache=], which is owned by user agent as a global storage, is not modified while prerendering, but is properly updated after activation. The following modifications ensure this:
+
+Each [=prerendering browsing context=] maintains its <dfn for="prerendering browsing context">prerender-scoped Accept-CH cache</dfn>, to persist the origin's opt-in to receive the indicated header fields.
```
jeremyroman, line 680: nit: I'd suggest slightly separating the normative and informative parts of this more sharply, and putting the normative part first. Maybe something like:

```
Each [=prerendering browsing context=] has a <dfn for="prerendering browsing context">prerender-scoped Accept-CH cache</dfn>, which is an [=ordered map=] from [=origin=] to [=client hint set=].

<p class="note">This stores which client hints each origin has opted into receiving, until it can be copied to the global [=Accept-CH cache=] when [=prerendering browsing context/finalize activation|activation is finalized=].</p>
```

domenic: Although I like Jeremy's rephrasing (modulo preferring plurals for the map declaration), I think it's a bit nicer to keep the explanation as part of the same non-note paragraph. We are debating really fine points of style here, but I tend to prefer notes to be more like "Note: some interesting supplementary information" instead of "Note: some basic information".

prerendering.bs
domenic: Hmm, I was the one who suggested this plural "of ... to" phrasing, and I think it matches existing spec verbiage. E.g. https://html.spec.whatwg.org/#concept-worklet-module-responses-map or https://wicg.github.io/navigation-api/#navigation-upcoming-traverse-navigations .

prerendering.bs
```diff
@@ -658,6 +673,41 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+      1. Let |orign| be |request|'s [=request/url=]'s [=url/origin=].
+      1. Set |hintSet| to the [=set/union=] of |hintSet| and |browsingContext|'s [=prerendering browsing context/prerender-scoped Accept-CH cache=][|origin|].
+  
+</div>
+
+<div algorithm="update the Client Hints set patch">
+  Modify the <a spec=CLIENT-HINTS-INFRASTRUCTURE>update the Client Hints set</a> algorithm, by updating the last step.
+
+  1. If |browsingContext| is a [=prerendering browsing context=],
+    [=add a new prerender Accept-CH cache entry=] with |response|'s [=response/url=]'s [=url/origin=] and |settingsObject|'s [=environment settings object/client hints set=].
```
domenic, line 706: This is missing passing |browsingContext|, now that "add a new prerender Accept-CH cache entry" takes three arguments

prerendering.bs
domenic: This is an example where splitting it into a note aligns with my sense of style :).

domenic (approved):
Only crucial feedback is fixing the missing |browsingContext|, so I will give this a green checkmark from my side assuming that gets fixed. Great work!

domenic:
Oh, and per the [CI build failure](https://github.com/WICG/nav-speculation/runs/8237625287?check_suite_focus=true) you need to change some `|variableReference|`s to `<var ignore>variableReference</var>`. This is an unfortunate side effect of doing these sorts of patch specs.

prerendering.bs
Clqsin45: Thanks! Done.

prerendering.bs
Clqsin45: Thanks! :D got it!

prerendering.bs
Clqsin45: fixed.

prerendering.bs
Clqsin45: I used "Else" per this document (https://garykac.github.io/procspec/). 
But maybe the keypoint is to be consistent? 

Used "Otherwise, " instead.

prerendering.bs
Clqsin45: There is the pattern of "map of typeA to typeB" in my corpus.. I cannot tell if it is widely used, but it seems understandable. 
( I find some sentences in this specifiation also use this pattern, so can I keep it for now and update them if we find "map from typeA to typeB" is preferred to  "map of typeA to typeB" ?

prerendering.bs
Clqsin45: Thanks!! Split this one and others!

prerendering.bs
Clqsin45: This part is describing the usage of prerender-scoped CH cache, based on https://www.rfc-editor.org/rfc/rfc8942(This should be a specification).
I'd like this to be a part of the "normative paragraph", and wonder how I can make it more normative..

prerendering.bs
Clqsin45: ooops.. 💯 right, I fixed it. 
Thanks!!

Clqsin45:
Thank you for reviewing and the detailed comments!

Clqsin45:
Would you help take another look?

jeremyroman (approved):

prerendering.bs
```diff
@@ -658,6 +673,48 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
 
 1. Return a {{DOMHighResTimeStamp}} with a time value equal to the [=current document=]'s [=Document/activation start time=].
 
+<h3 id="client hint cache">Interaction with Client Hint Cache</h3>
```
domenic, line 676:
```suggestion
<h3 id="client-hint-cache">Interaction with Client Hint Cache</h3>
```

prerendering.bs
```diff
@@ -658,6 +673,48 @@ The <dfn attribute for="PerformanceNavigationTiming">activationStart</dfn> gette
...
+
+We need to ensure that the [=Accept-CH cache=], which is owned by user agent as a global storage, is not modified while prerendering, but is properly updated after activation. The following modifications ensure this.
+
+Each [=prerendering browsing context=] has a <dfn for="prerendering browsing context">prerender-scoped Accept-CH cache</dfn>, which is an [=ordered map=] from [=origin=] to [=client hints set=].
+
+This stores which client hints each origin has opted into receiving, until it can be copied to the global [=Accept-CH cache=] when [=prerendering browsing context/finalize activation|activation is finalized=].
+
+The [=prerendering browsing context/prerender-scoped Accept-CH cache=] is an [=ordered map=] of [=origins=] to [=client hints sets=].
```
domenic, line 684: This seems redundant with line 680, but it uses the preferred phrasing ("of origins to client hint sets"). So let's change line 380 to this style and delete this line.

domenic (approved):
LGTM with nits.

prerendering.bs
Clqsin45: Thanks! I updated that line and removed this line.

domenic (approved):

PR #183, opened 2022-09-06
Add an explainer for No-Vary-Query

domenic:
Not yet finished and ready for final review, but I'm putting it up in draft PR form in case people want to comment anyway.

### DIFF
--- a/no-vary-query.md
+++ b/no-vary-query.md
@@ -1,0 +1,264 @@
+# The `No-Vary-Query` HTTP response header
+
+Caching is useful for making web pages load faster, and thus creating better user experiences. Prominent caches on the web platform include the HTTP cache, as well as the subjects of this repository, the prefetch and prerender caches.
+
+One of the most important cache keys for web resources is the resource's URL. However, sometimes multiple URLs can represent the same resource. This leads to caches not always being as helpful as they could be: if the browser has the resource cached under one URL, but the resource is then requested under another, the cached version will be ignored.
+
+This proposal tackles a specific subset of this general problem, for when multiple URLs for the same resource differ according to their URL's query components. Via a new HTTP header, `No-Vary-Query`, resources can declare to the browser that some or all parts of the query can be ignored for cache matching purposes. For example, if the order of the query parameters should not cause cache misses, this is indicated using
+
+```http-header
+No-Vary-Query: order
+```
+
+If the specific query parameters (e.g., ones indicating something for analytics) should not cause cache misses, this is indicated using
+
+```http-header
+No-Vary-Query: "utm_source", "utm_medium", "utm_campaign"
+```
+
+And if the page instead wants to take an allowlist-based approach, where only certain known query parameters should cause cache misses, they can use
+
+```http-header
+No-Vary-Query: *; except=("productId")
+```
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+## Table of contents
+
+- [Goals](#goals)
+- [Non-goals](#non-goals)
+- [Prior art](#prior-art)
+- [Use cases](#use-cases)
+  - [Carrying data that is or can be processed by client-side script only](#carrying-data-that-is-or-can-be-processed-by-client-side-script-only)
+  - [Avoiding unnecessary cache mismatches due to inconsistent referrers](#avoiding-unnecessary-cache-mismatches-due-to-inconsistent-referrers)
+  - [Carrying data not yet determined at the time of preloading](#carrying-data-not-yet-determined-at-the-time-of-preloading)
+  - [Customizing behavior when preloading](#customizing-behavior-when-preloading)
+- [Detailed design](#detailed-design)
+  - [The header](#the-header)
+  - [Integration with…](#integration-with)
+    - [… HTTP caches](#-http-caches)
+    - [… preloading caches](#-preloading-caches)
+    - [… the cache API](#-the-cache-api)
+    - [… other browser caches](#-other-browser-caches)
+  - [Interaction with prerendering activation](#interaction-with-prerendering-activation)
+  - [Interaction with redirects](#interaction-with-redirects)
+  - [Interaction with storage partitioning](#interaction-with-storage-partitioning)
+- [Alternatives considered](#alternatives-considered)
+- [Future extensions](#future-extensions)
+  - [More complex no-vary rules](#more-complex-no-vary-rules)
+  - [`No-Vary-Path`](#no-vary-path)
+  - [A referrer hint](#a-referrer-hint)
+- [Security and privacy considerations](#security-and-privacy-considerations)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Goals
+
+- Allow caches to avoid keying on the order of URL query parameters
+
+- Allow caches to avoid keying on specifically-named URL query parameters
+
+- Allow caches to avoid keying on all URL query parameters, or all-but-some-specifically-named URL query parameters
+
+## Non-goals
+
+- Allow more complicated keying rules than the above, e.g. "treat `key` and `KEY` as equivalent"; or, "treat `value1` and `value2` as equivalent for `key`, but treat `value3` differently"; or, "order doesn't matter for most parameters, but the position of the `key` parameter matters"; or, "allow the `width` parameter to match as long as its value is within 100 of the already-cached URL's `width` value". Some of these [seem possible as future extensions](#more-complex-no-vary-rules), but we are not designing around them.
+
+- Support non-standard query parameter structures, e.g. `?color=red;size=large` with a `;` delimiter instead of a `&`. We will use the [`application/x-www-form-urlencoded` format](https://url.spec.whatwg.org/#concept-urlencoded) supported by the URL Standard.
+
+- Allow caches to avoid keying on other parts of the URL, notably the path. Although this could be useful, we think it would best be done via [an additive separate solution](#no-vary-path).
+
+## Prior art
+
+This proposal takes some of its inspiration from the existing [`Vary`](https://httpwg.org/specs/rfc7231.html#header.vary) HTTP header, which lets servers indicate what _header names_ should be _included_ when constructing the cache key. Unlike for headers, by default responses vary on their URL (including query string), so our proposal uses the `No-Vary-` prefix to show how it's indicating which query parameters should be _excluded_ when constructing the cache key. We see these two headers, the existing `Vary` and our proposed `No-Vary-Query`, as part of a general cache-key-construction mechanism, and want all places on the web platform that respect the former to also respect the latter.
+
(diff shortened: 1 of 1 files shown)


no-vary-query.md
```diff
@@ -0,0 +1,264 @@
+# The `No-Vary-Query` HTTP response header
+
+Caching is useful for making web pages load faster, and thus creating better user experiences. Prominent caches on the web platform include the HTTP cache, as well as the subjects of this repository, the prefetch and prerender caches.
+
+One of the most important cache keys for web resources is the resource's URL. However, sometimes multiple URLs can represent the same resource. This leads to caches not always being as helpful as they could be: if the browser has the resource cached under one URL, but the resource is then requested under another, the cached version will be ignored.
+
+This proposal tackles a specific subset of this general problem, for when multiple URLs for the same resource differ according to their URL's query components. Via a new HTTP header, `No-Vary-Query`, resources can declare to the browser that some or all parts of the query can be ignored for cache matching purposes. For example, if the order of the query parameters should not cause cache misses, this is indicated using
```
jeremyroman, line 7: nit: "their URL's" technically expands to "multiple URLs' URL's", and I'd suggest emphasizing that the URLs differ only in query, i.e, "...for when a resource has multiple URLs which differ only in certain query components".

nit: elide "to the browser", since any HTTP cache implementation could also process this in theory; while transparent caching proxies are uncommon these days, caching CDNs are common and offer similar functionality.

no-vary-query.md
```diff
@@ -0,0 +1,264 @@
...
+
+- Support non-standard query parameter structures, e.g. `?color=red;size=large` with a `;` delimiter instead of a `&`. We will use the [`application/x-www-form-urlencoded` format](https://url.spec.whatwg.org/#concept-urlencoded) supported by the URL Standard.
+
+- Allow caches to avoid keying on other parts of the URL, notably the path. Although this could be useful, we think it would best be done via [an additive separate solution](#no-vary-path).
+
+## Prior art
+
+This proposal takes some of its inspiration from the existing [`Vary`](https://httpwg.org/specs/rfc7231.html#header.vary) HTTP header, which lets servers indicate what _header names_ should be _included_ when constructing the cache key. Unlike for headers, by default responses vary on their URL (including query string), so our proposal uses the `No-Vary-` prefix to show how it's indicating which query parameters should be _excluded_ when constructing the cache key. We see these two headers, the existing `Vary` and our proposed `No-Vary-Query`, as part of a general cache-key-construction mechanism, and want all places on the web platform that respect the former to also respect the latter.
```
jeremyroman, line 74: The "all places on the web platform..." statement is quite strong. If we want to make it, it would be good to also note that any software which doesn't implement `No-Vary-Query` wouldn't behave incorrectly, it would just have fewer cache hits than it might otherwise.

no-vary-query.md
```diff
@@ -0,0 +1,264 @@
...
+```
+
+When the page loads, we don't know yet whether the user will click on the hero image or the headline link. So it's useful to prerender the page with no query parameter, and then later [fill in that information](#interaction-with-prerendering-activation) once the user clicks on a link.
+
+_Note: analytics, where you carry generic data such as the marketing campaign or referring source, is different than [navigational tracking](https://privacycg.github.io/nav-tracking-mitigations/#terminology), which passes along user identifiers. Nothing about this proposal is intended to aid navigational tracking, and we expect it to be fully compatible with future and [currently-deployed mitigations](https://privacycg.github.io/nav-tracking-mitigations/#deployed-mitigations) for navigational tracking._
+
+### Customizing behavior when preloading
+
+TODO
+
```
jeremyroman, line 128: This might be rather niche but we could also consider cases where it impacts server-side processing in a way that doesn't affect the resulting resource, e.g., to request load balancing to a particular instance/pool or enable server-side debug logging for a particular request.

While these could be indicated with a request header instead, it's not always convenient either client-side (to add request headers to some types of request, such as link navigations, form submissions and subresource requests) or server-side (server software might be written in such a way as to make policy decisions before request headers are received).

no-vary-query.md
```diff
@@ -0,0 +1,264 @@
...
+## Detailed design
+
+### The header
+
+The `No-Vary-Query` header is a [HTTP structured field](https://www.rfc-editor.org/rfc/rfc8941.html) whose value must be a list. The list can contain either predefined [tokens](https://www.rfc-editor.org/rfc/rfc8941.html#name-tokens), which have special behavior, or [strings](https://www.rfc-editor.org/rfc/rfc8941.html#name-strings), which indicate query parameters to not-vary on. The special tokens are:
```
jeremyroman, line 133: nit: "to not-vary" scans less well than "not to vary" to me

no-vary-query.md
```diff
@@ -0,0 +1,264 @@
...
+
+To resolve this, we specify that prerendering [activation](https://wicg.github.io/nav-speculation/prerendering.html#prerendering-browsing-context-activate) involves changing the page's URL before firing the `prerenderingchange` event. This URL change is done using the [URL and history update steps](https://html.spec.whatwg.org/#url-and-history-update-steps), i.e. the same mechanism underlying `history.replaceState()`.
+
+With this setup, prerendered pages which care about a given query parameter, but don't want mismatches to prevent a prerender, can delay processing of the query parameter until activation time. For example:
+
+```js
+document.addEventListener("prerenderingchange", () => {
+  const via = (new URLSearchParams(location.search)).get("via");
```
jeremyroman, line 187: super-nit: the parentheses around the `new` expression aren't necessary, since that binds more tightly than the property-invoke expression :)

no-vary-query.md
```diff
@@ -0,0 +1,264 @@
...
+
+TODO
+
+### Interaction with storage partitioning
+
+This proposal takes place fully within the bounds of the [client side storage partitioning](https://github.com/privacycg/storage-partitioning/) work. It is only about calculating the cache key within a given partition. It does not affect the computation of the partition key itself.
+
+## Alternatives considered
+
+TODO
```
jeremyroman, line 202: Some of these do exist in the internal outline doc.

no-vary-query.md
```diff
@@ -0,0 +1,264 @@
...
+
+- We could allow restrictions on the value space, using syntax like `No-Vary-Query: "color";regexp="(?:blue|azure)"`.
+
+- We could allow treating multiple keys as the same key, using syntax like `No-Vary-Query: ("color", "colour")`.
+
+We think capabilities such as these serve less urgent [use cases](#use-cases) than what we have so far, and so don't plan on supporting them now. But it's good to know there's room for them in the future.
+
+### `No-Vary-Path`
+
+Many of the arguments for avoiding cache misses on differing query parameters, also apply to avoiding cache misses on differing paths. This is most prominent in the case of client-rendered single-page applications, where e.g. `/products/123` and `/products/456` might both have the same HTTP response, containing the skeleton product page, with their differing content being done via client-side rendering.
```
jeremyroman, line 220: Even though the web doesn't provide a total security boundary, it's some shared hosting providers (e.g., university departments) often separate different authors by path alone (more often than by query alone). That's not to say this couldn't be done eventually, but it merits a little additional thought on the risk/return tradeoff.

no-vary-query.md
```diff
@@ -0,0 +1,264 @@
...
+- [Module maps](https://html.spec.whatwg.org/#module-map)
+- [Shared workers](https://html.spec.whatwg.org/#shared-workers-and-the-sharedworker-interface:concept-sharedworkerglobalscope-constructor-url)
+- [Resources preloaded via `<link rel="preload">`](https://html.spec.whatwg.org/#map-of-preloaded-resources)
+
+The proposal does _not_ update cache key computation for these caches. This is because:
+
+- None of them respect the `Vary` header today
+- All of them are keyed by the _request_ URL, and not by the _response_ URL, whereas `No-Vary-Query` is a property provided by the response.
+
+### Interaction with prerendering activation
```
jeremyroman, line 177: It seems like it's probably also worth covering the general case even though it's trivial, i.e., that `window.location` (and everything else that looks at history) reflects the URL that was navigated to (i.e., the URL used to query the cache), not the request/response URL for the fetch that populated the cache.

This applies both to prefetching and the general cache application.

no-vary-query.md
domenic: Yeah, I just hate that, so I keep them :). https://github.com/eslint/eslint/issues/12428

no-vary-query.md
domenic: Yep, just ran out of time yesterday; will be porting them over today.

no-vary-query.md
domenic: I'll expand on the consequences of the URL and history update steps, but I'm not sure I understand your last point.

domenic: Hmm, I think I see what you mean now. Let me try.

jeremyroman: Yeah, you figured out what I meant. :)

no-vary-query.md
domenic: I like this expansion of the framing, and think it'll be pretty compelling! I've tried to incorporate it; PTAL.

no-vary-query.md
domenic: At first I thought this was too weak of a consideration to include, but then I remembered service workers did indeed add a path restriction. So I've incorporated this. Thanks!

annevk:
I don't have a strong opinion but I don't think we have exposed "Query" anywhere and when we added `URLSearchParams` we decided to stick with the status quo. HTTP headers are new ground, but unless there's precedent I think we ought to stick with "Search" or "URL-Search".

no-vary-query.md
```diff
@@ -0,0 +1,337 @@
...
+
+The proposed solution could be given a narrower scope, such a header respected only by the preloading caches. We're open to this if necessary, but we think there's enough evidence of this being a larger problem that it's worth tackling other caches at the same time.
+
+The response header could be flipped, i.e. `Vary-Query`, so as to better match `Vary`. We chose the `No-Vary-Query` framing because it means that the "empty" default state matches existing HTTP caching semantics (i.e., that the response varies by all query parameters). A `Vary-Query` version would probably need to have different behavior when the header is absent (vary on everything) vs. when the header is present but empty (vary on the empty list, i.e. nothing), which seems awkward.
+
+The name could be `No-Vary-Search`, to better match web platform APIs. Actually, maybe we should just make this change?
+
+## Future extensions
```
jeremyroman, line 263: I do wonder based on the reception of this at TPAC whether we should be clearer that these are evidence that if we want these features it would be possible to extend to encompass them, as opposed to actual future extensions we anticipate implementing in the foreseeable future (at least for the first two). Especially No-Vary-Path I think has interesting implementation questions, and I'm also uneasy about regular expressions here.

If you agree, maybe we could rename this "Extensibility" and add an intro paragraph to clarify?

jeremyroman:
I could live with `No-Vary-Search` or even `No-Vary-Search-Params` (arguably clearer, given "search" alone seems a little vague, but probably too long) if we prefer to not start a precedent of using the word "query". (As disappointed as I would be to lose the query/vary rhyme, depending on your pronunciation of "query".)

domenic:
OK, I think this is ready to merge; @jeremyroman want to give it a final look?

no-vary-search.md
```diff
@@ -0,0 +1,340 @@
...
+
+When the page loads, we don't know yet whether the user will click on the hero image or the headline link. So it's useful to prerender the page with no query parameter, and then later [communicate that information](#prerendering-activation) once the user clicks on a link.
+
+## Detailed design
+
+### The header
+
+The `No-Vary-Search` header is a [HTTP structured field](https://www.rfc-editor.org/rfc/rfc8941.html) whose value must be a list. The list can contain either predefined [tokens](https://www.rfc-editor.org/rfc/rfc8941.html#name-tokens), which have special behavior, or [strings](https://www.rfc-editor.org/rfc/rfc8941.html#name-strings), which indicate query parameters not to vary on. The special tokens are:
+
+- `order`: indicates the order of the query parameters should be ignored when constructing the cache key.
```
jeremyroman, line 139: Do we want to clarify that we mean only relative to distinct keys? (At least, that is what https://url.spec.whatwg.org/#dom-urlsearchparams-sort does and seems likely to be the most useful.)

no-vary-search.md
```diff
@@ -0,0 +1,340 @@
...
+
+Since everything about the HTTP cache is best-effort, using `No-Vary-Search` will not guarantee additional cache hits. As such, browsers might implement HTTP cache integration after this feature has already proven its worth on, e.g., preloading caches.
+
+### Navigated-to pages
+
+When navigating to a URL which can be matched with a cached resource via `No-Vary-Search`, we still treat the resource _as if_ it were fetched from the target URL, and not from the originally-cached URL. For example:
+
+- Service workers and resource timing APIs see fetches go by targeting the original request URL.
+- Once the document is constructed and displayed:
+  - `location.href` is the navigated-to-URL.
```
jeremyroman, line 190: maybe clarify this is after redirects, at least once?

jeremyroman (approved):

no-vary-search.md
domenic: I'll just slip in "response URL"... I think getting too far into the details of redirects at this point risks drawing attention from the main point, which is to contrast two conceivable possibilities for the document's URL.

PR #184, opened 2022-09-08
Document rules spec

jeremyroman:
This specifies the structure, parsing, and processing of document rules, including use of CSS selectors and URL patterns, and a basic and/or/not boolean algebra.

Referrer policy (including `rel=noreferrer`) was previously included (see earlier commits) but has since been removed consistent with the removal of referrer policy from the rest of this specification. It should be restored when referrer policy plumbing generally is again.

While this is specified in terms of periodically scanning the entire document tree, implementations are welcome to track invalidations due to addition/removal of links, changes to key attributes (`href`, `target`, etc), changes to the `<base>` element, and any changes which would affect the CSS selectors, and only make the necessary partial changes.

Preview: https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jeremyroman/nav-speculation/document-rules-spec/speculation-rules.bs


jeremyroman:
@a4sriniv FYI (though your review is also welcome)

speculation-rules.bs
```diff
@@ -275,10 +389,23 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
...
     1. &#x231B; [=list/For each=] |rule| of |ruleSet|'s [=speculation rule set/prerender rules=]:
       1. &#x231B; [=list/For each=] |url| of |rule|'s [=speculation rule/URLs=]:
         1. &#x231B; Let |prerenderCandidate| be a new [=prerender candidate=] whose [=prerender candidate/URL=] is |url| and [=prerender candidate/target browsing context name hint=] is |rule|'s [=speculation rule/target browsing context name hint=].
         1. &#x231B; [=list/Append=] |prerenderCandidate| to |prerenderCandidates|.
+      1. &#x231B; If |rule|'s [=speculation rule/predicate=] is not null, then:
+        1. &#x231B; Let |links| be the result of [=finding matching links=] given |document| and |rule|'s [=speculation rule/predicate=].
+        1. &#x231B; [=list/For each=] |link| of |links|:
+          1. &#x231B; Let |href| be the result of running |link|'s {{HTMLHyperlinkElementUtils/href}} getter steps.
+          1. &#x231B; Let |target| be |rule|'s [=speculation rule/target browsing context name hint=].
+          1. &#x231B; If |target| is null, set it to the result of [=getting an element's target=] given |link|.
```
domenic, line 406: Hmm, this is making me wonder if we should use `<base target>` for list rules too, just like we already use `<base href>`!

speculation-rules.bs
```diff
@@ -261,6 +362,19 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
...
 </div>
 
+<div algorithm>
+  To <dfn>find matching links</dfn> given a {{Document}} |document| and [=document rule predicate=] |predicate|:
+
+  1. Let |links| be an empty [=list=].
+  1. For each [=shadow-including descendant=] |descendant| of |document|, in [=shadow-including tree order=]:
+    1. If |descendant| is not an <{a}> with an <{a/href}> attribute or <{area}> element with an <{area/href}> attribute, [=iteration/continue=].
+
+        <div class="note">This corresponds to the elements which match the <a dfn-type="selector">:link</a> and <a dfn-type="selector">:visited</a> [=pseudo-classes=], or which appear in the {{Document/links}} collection.</div>
```
domenic, line 372: Can we use `:any-link` instead?

speculation-rules.bs
```diff
@@ -300,6 +427,40 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
...
+    1. [=list/For each=] |clause| of |predicate|'s [=document rule disjunction/clauses=]:
+      1. If |clause| [=document rule predicate/matches=] |el|, return true.
+    1. Return false.
+  1. If |predicate| is a [=document rule negation=], then:
+    1. If |predicate|'s [=document rule negation/clause=] [=document rule predicate/matches=] |el|, return false.
+    1. Return true.
+  1. If |predicate| is a [=document rule URL pattern predicate=], then:
+    1. Let |href| be the result of running |el|'s {{HTMLHyperlinkElementUtils/href}} getter steps.
+    1. [=list/For each=] |pattern| of |predicate|'s [=document rule URL pattern predicate/patterns=]:
+      1. Execute |pattern|'s {{URLPattern/test(input, baseURL)}} steps given |href|. If the reuslt is true, return true.
```
domenic, line 449: Let's use the https://wicg.github.io/urlpattern/#match operation directly

Also, typo "reuslt"

speculation-rules.bs
```diff
@@ -300,6 +427,40 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
...
+    1. Return false.
+  1. If |predicate| is a [=document rule CSS selector predicate=], then:
+    1. [=list/For each=] |selector| of |predicate|'s [=document rule CSS selector predicate/selectors=]:
+      1. [=match a selector against an element|Match=] |selector| against |el| with [=scoping root=] |el|'s [=Node/node document=]. If the result is true, return true.
```
domenic, line 453:
```suggestion
      1. [=match a selector against an element|Match=] |selector| against |el| with the [=scoping root=] set to |el|'s [=Node/node document=]. If the result is true, return true.
```

Although, I wonder if it should be not supplied? Or set to el's root? Not supplied means you can't use `:scope` I think, which is perhaps simplest?

speculation-rules.bs
```diff
@@ -231,7 +276,63 @@ Inside the [=prepare the script element=] algorithm we make the following change
...
+    1. Let |rawClause| be the |input|[|predicateType|].
+    1. Let |clause| be the result of [=parsing a document rule predicate=] given |rawClause|.
+    1. If |clause| is null, then return null.
+    1. Return a [=document rule negation=] whose [=document rule negation/clause=] is |clause|.
+  1. If |predicateType| is "`href_matches`":
+    1. If |input| has any [=map/key=] other than "`href_matches`", then return null.
+    1. Let |rawPatterns| be |input|["`href_matches`"].
+    1. If |rawPatterns| is not a [=list=], then set |rawPatterns| to « |rawPatterns| ».
+    1. Let |patterns| be an empty [=list=].
+    1. [=list/For each=] |rawPattern| of |rawPatterns|:
```
domenic, line 309: So this allows `"href_matches": ["pattern1", "pattern2"]` instead of requiring the more explicit `"or": [{"href_matches": "pattern1"}, {"href_matches": "pattern2"}]`. I'm not sure that's a good tradeoff? Especially for selectors where we now have *three* ways of doing "or", the third being `"selector_matches": "selector1, selector2"`?

domenic (commented):
Generally looks great! Seems like things turned out pretty elegantly.

The only substantial comment is about whether we want to allow the array forms of `"href_matches"` and `"selector_matches"` after all. Apologies for not discussing that in the explainer thread, and feel free to spin that off into its own issue if you'd prefer to proceed with landing this sooner rather than later.

speculation-rules.bs
a4sriniv: Or alternatively just making this match `Element.matches()`, i.e. set scope to |el|?

speculation-rules.bs
jeremyroman: We could. My reason for this choice is this corresponds more directly to [how it's defined in HTML](https://html.spec.whatwg.org/multipage/semantics-other.html#selector-link) and because I think `:any-link` is a little less familiar to most readers. I have no real objection to changing though; lmk if you'd prefer that.

speculation-rules.bs
jeremyroman: That's a valid question. I don't think the answer is obvious -- for links, we should respect whatever target the link would have, but for a list rule we don't know whether a `<base target>` respecting link navigation will be used or not.

While `<base href>` defines how all URLs should be interpreted throughout the document (and so clearly must apply), `<base target>` merely provides a default value for `<a>`, `<form>`, and a handful of other cases.

As another example, navigation by `location = ...` would respect `<base href>` but not `<base target>`, while link navigation and form submission would affect both.

speculation-rules.bs
jeremyroman: Done.

speculation-rules.bs
jeremyroman: I admit to not seeing `:scope` being used much and therefore not having a good intuition for it.

My gut is that what we're doing here is fundamentally more similar to `document.querySelectorAll` (i.e., we're searching for matching elements within the document when we consider speculation). That algorithm uses [scope-match a selectors string](https://dom.spec.whatwg.org/#scope-match-a-selectors-string) which provides the document itself as the scoping root. (Aside: I can't find in the spec why the root element matches `:scope` despite the document being provided as the scoping root, but this does seem to be the case.)

`Element.matches` would be the obvious counterexample, but I think that's different, because as far as the API user is concerned, they're searching the document, not particular elements. I could be convinced otherwise, though.

So it seems to me the motivating factor should be "what, to the author, is the scope within which their selector applies"? I think that's the document whose tree we're walking.

speculation-rules.bs
```diff
@@ -231,7 +276,63 @@ Inside the [=prepare the script element=] algorithm we make the following change
...
+
+            <div class="note">This will need to be updated if {{URLPattern}} gains members of other types.</div>
+          1. Set |init|[|key|] to |value|.
+        1. Set |pattern| to the result of constructing a {{URLPattern}} using the {{URLPattern/URLPattern(input, baseURL)}} constructor steps given |init|. If those steps throw, catch the exception and return null.
+      1. Otherwise, return null.
+      1. [=Assert=]: |pattern| is a {{URLPattern}}.
+      1. [=list/Append=] |pattern| to |patterns|.
+    1. Return a [=document rule URL pattern predicate=] whose [=document rule URL pattern predicate/patterns=] is |patterns|.
+  1. If |predicateType| is "`selector_matches`":
+    1. If |input| has any [=map/key=] other than "`href_matches`", then return null.
```
a4sriniv, line 327: Should be `selector_matches` instead of `href_matches` here and below.

speculation-rules.bs
```diff
@@ -231,7 +276,63 @@ Inside the [=prepare the script element=] algorithm we make the following change
...
+  1. Return a [=speculation rule=] with [=speculation rule/URLs=] |urls|, [=speculation rule/predicate=] |predicate|, [=speculation rule/requirements=] |requirements|, and [=speculation rule/target browsing context name hint=] |targetHint|.
+</div>
+
+<div algorithm="parse a document rule predicate">
+  To <dfn>parse a document rule predicate</dfn> given a value |input| and [=URL=] |baseURL|:
+
+  1. If |input| is not a [=map=], then return null.
+  1. If |input| does not [=map/contain=] exactly one of "`and`", "`or`", "`not`", "`href_matches`" and "`selector_matches`", then return null. Otherwise, let |predicateType| be that key.
+  1. If |predicateType| is "`and`" or "`or`", then:
+    1. If |input| has any [=map/key=] other than |predicateType|, then return null.
```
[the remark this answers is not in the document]
domenic: Thanks, that makes sense. I'm not sure if this is subtle (and thus would benefit from a small example), or if it's obvious and I was being dumb.

speculation-rules.bs
```diff
@@ -231,7 +276,63 @@ Inside the [=prepare the script element=] algorithm we make the following change
...
+    1. Let |patterns| be an empty [=list=].
+    1. [=list/For each=] |rawPattern| of |rawPatterns|:
+      1. Let |serializedBaseURL| be the [=URL serializer|serialization=] of |baseURL|.
+      1. Let |pattern| be null.
+      1. If |rawPattern| is a [=string=], then:
+        1. Set |pattern| to the result of constructing a {{URLPattern}} using the {{URLPattern/URLPattern(input, baseURL)}} constructor steps given |rawPattern| and |serializedBaseURL|. If those steps throw, catch the exception and return null.
+      1. Otherwise, if |rawPattern| is a [=map=], then:
+        1. Let |init| be «[ "`baseURL`" → |serializedBaseURL| ]», representing a dictionary of type {{URLPatternInit}}.
+        1. [=map/For each=] |key| → |value| of |rawPattern|:
+          1. If |key| is not the [=identifier=] of a [=dictionary member=] of {{URLPatternInit}} or one of its [=inherited dictionaries=], |value| is not a [=string=], or the member's type is not declared to be {{USVString}}, then return null.
```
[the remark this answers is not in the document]
domenic: I was confused between "inherited dictionaries" and "derived dictionaries".

speculation-rules.bs
domenic: The `location` example convinced me; thanks!

speculation-rules.bs
domenic: Your versus example is tough... the latter version is clearly more concise, but it's still not obvious to me that it's an "or" operation. Hmm. I guess I'm OK keeping it.

domenic (approved):

speculation-rules.bs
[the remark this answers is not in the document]
jeremyroman: Added a couple notes.

speculation-rules.bs
jeremyroman: Acknowledged.

speculation-rules.bs
jeremyroman: Done. (I actually don't know how many people are familiar with `:any-link` but it's not hard to learn if the reader isn't, so it's fine.)

Issue #185, opened 2022-09-14
Add specification for Web Locks API


Clqsin45:
Requesting a web lock returns a promise, so this API seems can be deferred. 
We may want to describe the expected behavior.

Clqsin45:
(filed this issue because I am reviewing a CL that upstreams the tests for Web Locks. happy to update the specification for it 
:-))

domenic:
Happy to review such an update! It seems like it might very similar to other simple cases like https://wicg.github.io/nav-speculation/prerendering.html#webhid-patch .

Clqsin45:
Thanks for the link! I add a new section to the specification and uploaded the patch.

PR #186, opened 2022-09-15
Add specification for Web Locks API

Clqsin45:
WebLocks API's request() and query() methods return promises, and prerender can safely defer them.
This PR describes how this API should work in the prerendering context.


Clqsin45:
This PR is for #185

domenic (approved):

PR #190, opened 2022-09-27
Remove all mention of navigate-to

domenic:
It was removed from CSP!


prerendering.bs
```diff
@@ -516,7 +516,7 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
       * |browsingContext|'s [=active document=]'s [=Document/prerendering browsing contexts map=][|request|'s [=request/URL=]] [=map/exists=] and is not [=prerendering browsing context/empty=]
       * The result of calling [=should navigation request of type be blocked by content security policy?=] given |request| and <var ignore>navigationType</var> is "`Allowed`".
 
-        <p class="note">There is no need for an additional CSP check for the response, see <tt>navigate-to</tt> (before it was <a href="https://github.com/w3c/webappsec-csp/pull/564">removed</a>.
+        <p class="note">There is no need for an additional CSP check for the response.
```
domenic, line 519: I'm not sure I fully remember the purpose of this note, so maye we should just delete it...

jeremyroman: I _think_ we can probably remove it too. I had the same thought when I was fixing the build (which broke when navigate-to was removed).

jeremyroman (approved):

PR #191, opened 2022-10-21
Add Speculation-Rules header to explainer

kjmcnee:
We add a description of the proposed Speculation-Rules response header for fetching external speculation rule sets.


triggers.md
```diff
@@ -96,6 +98,13 @@ The following example illustrates the basic idea:
 
 (The use of `<script>` is basically because only a few elements allow the HTML parser to enter RCDATA mode, and `<script>` has already been used in this way. This doesn't necessarily mean that these rules should respect the `script-src` CSP directive.)
 
+For use cases where it is preferable to provide speculation rule sets without modifying a document, we introduce the `Speculation-Rules` HTTP header. The header's values are URLs to external resources containing speculation rule sets. The rules are expressed in the same JSON object format as the inline case. The MIME type of the resource must be `application/speculationrules+json`.
+
+For example, suppose a rule set is available at `https://example.com/speculation-rules.json`. The response headers for a request for `https://example.com/` could include:
+```
+Speculation-Rules: speculation-rules.json
```
jeremyroman, line 105: I believe the current Chromium implementation uses a [structured field](https://www.rfc-editor.org/rfc/rfc8941.html), which would require this to be quoted. SF are generally preferred for new headers, and it gives us flexibility to extend the syntax in the future (e.g., with parameters).

triggers.md
```diff
@@ -340,7 +353,17 @@ This has a few disadvantages:
...
...ld introduce a new [link type][link-type], "speculationrules." The header is semantically equivalent to the HTML `<link>` element, so using this header would mean we would have to support speculation rules loading with a `<link>` element as well. This seems sensible, especially given that web developers would already be familiar with loading declarative content with `<link>` elements (e.g. style sheets). However, this has a number of drawbacks. (1) For new kinds of resources to fetch, we want to require CORS. The `<link>` element has an unsuitable default in the form of the `crossorigin` attribute. (2) There are a number of other attributes which are undesirable or irrelevant for speculation rules (e.g. `type`). (3) The rel attribute can actually have multiple values, so this may lead to confusing combinations (e.g. rel="prefetch speculationrules"). (4) Since we already use `<script>` elements for inline speculation rules, it would be appropriate to support external speculation rules with `<script src=”...”>`, as noted [above](#external-speculation-rules-via-script-elements), but then we would have two ways of doing the same thing which could negatively impact developer experience.
```
jeremyroman, line 358: Could you reformat this ordered list so it gets the appropriate markdown styling, like:

```md
1. Foo
2. Bar
```

jeremyroman, line 358: nit: fancy quotation marks are used in some places but not others; probably should just use " everywhere?

triggers.md
kjmcnee: Done.

triggers.md
kjmcnee: Done.

kjmcnee: Done.

triggers.md
```diff
@@ -96,6 +98,13 @@ The following example illustrates the basic idea:
 
 (The use of `<script>` is basically because only a few elements allow the HTML parser to enter RCDATA mode, and `<script>` has already been used in this way. This doesn't necessarily mean that these rules should respect the `script-src` CSP directive.)
 
+For use cases where it is preferable to provide speculation rule sets without modifying a document, we introduce the `Speculation-Rules` HTTP header. The header's values are URLs to external resources containing speculation rule sets. The rules are expressed in the same JSON object format as the inline case. The MIME type of the resource must be `application/speculationrules+json`.
```
domenic, line 101: There is a sentence above this that probably should be modified:

> Currently, like import maps, script tags are only used for specifying speculation rules inline; future extensions may allow external ones.

domenic (approved):
Looks great! Was mostly only able to find typographical nits. Thanks for all the typo fixes too!

triggers.md
kjmcnee: Done

jeremyroman (approved):

PR #194, opened 2022-10-26
Meta: fix a cross-link to "global object"

domenic:
Follows https://github.com/whatwg/html/pull/8356.

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -19,17 +19,21 @@ jobs:
         mkdir -p upload
         mv out upload/head
     - name: Build (base)
+      id: build_base
+      continue-on-error: true
       run: |
         git checkout HEAD^
         make ci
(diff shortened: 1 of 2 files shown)


domenic:
I think the build failure here is because the PR preview is trying to build the previous revision and failing. @jeremyroman should we just merge and then hope this doesn't happen again? Or do you want to use this opportunity to fix the preview script?

jeremyroman:
Yeah, the preview build just doesn't understand that the base revision no longer builds. Not a ton of work to fix but also doable at some later point; happy for you to just merge this.

jeremyroman (approved):

jeremyroman:
Alright, it turned out to be a one-liner to fix this (plus 3 more to skip trying to diff against nothing). It now just shows you the head revision and omits base/diff if it couldn't build base.

PR #195, opened 2022-10-26
Update No-Vary-Search syntax and add an initial spec

domenic:
This updates the proposed syntax based on the discussions in [this doc](https://docs.google.com/document/d/1ae3lDSwScuSy83HPH3S1HTTgNO-E0MWIf8k-8lUo2rk/edit#), and also includes the various alternatives discussed there so that we can have more public discussion of them.

@liviutinta I thought a spec might be helpful while I was at it.


jeremyroman:
Rebased on the revision with the fixed build and added the .bs to the Makefile for you, to generate a preview. :)

jeremyroman:
lgtm

A couple choices were slightly different than I had in mind but the non-normative text justifies the choices well and I'm happy with the result.

jeremyroman (approved):

PR #202, opened 2022-11-04
Add spec for Speculation-Rules header

kjmcnee:
We add specification text for the processing of the Speculation-Rules HTTP response header and the handling of external speculation rule resources.

### DIFF
--- a/speculation-rules.bs
+++ b/speculation-rules.bs
@@ -249,18 +249,18 @@ Note: This section contains modifications to [[HTML]].
     1. If |item| is not a [=string=], then [=iteration/continue=].
     1. Let |parsedURL| be the result of [=basic URL parser|parsing=] |item| with |baseUrl|.
     1. If |parsedURL| is failure, then [=iteration/continue=].
-    1. Let |pendingResource| be a [=pending external speculation rule resource=] with [=pending external speculation rule resource/URL=] |parsedURL| and [=pending external speculation rule resource/controller=] null
+    1. Let |pendingResource| be a [=pending external speculation rule resource=] with [=pending external speculation rule resource/URL=] |parsedURL| and [=pending external speculation rule resource/controller=] null.
     1. [=list/Append=] |pendingResource| to |document|'s [=document/list of pending external speculation rule resources=].
   1. [=Process pending external speculation rule resources=] given |document|.
 </div>
 
-In [=shared document creation infrastructure=], before the step
+In <a spec="HTML">create and initialize a `Document` object</a>, before the step
 
   > 19. Return <var ignore>document</var>.
 
 add the following step
 
-  > 19. [=process the Speculation-Rules header|Process the Speculation-Rules header=] given <var ignore>document</var> and <var ignore>navigationParams</var>'s [=response=].
+  > 19. [=Process the Speculation-Rules header=] given <var ignore>document</var> and <var ignore>navigationParams</var>'s [=response=].
 
 <h3 id="external-speculation-rule-sets">External speculation rule sets</h3>
 
@@ -286,7 +286,7 @@ add the following step
   1. Let |settingsObject| be |document|'s [=relevant settings object=].
   1. Let |requestURL| be |pendingResource|'s [=pending external speculation rule resource/URL=].
   1. Let |request| be a new [=request=] whose [=request/URL=] is |requestURL|, [=request/client=] is |settingsObject|, and [=request/mode=] is "`cors`".
-  1. Let |controller| be the result of [=fetching=] given |request| with [=fetch/processResponseConsumeBody=] set to the following steps given [=response=] |response| and null, failure, or a [=byte sequence=] |body|:
+  1. Let |controller| be the result of [=fetching=] given |request| with <i>[=fetch/processResponseConsumeBody=]</i> set to the following steps given [=response=] |response| and null, failure, or a [=byte sequence=] |body|:
     1. [=list/Remove=] |pendingResource| from |pendingResources|.
     1. If any of the following conditions are true, abort these steps:
       * |body| is null.
--- a/speculation-rules.bs
+++ b/speculation-rules.bs
@@ -46,7 +46,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: urls-and-fetching.html
       text: valid non-empty URLs potentially surrounded by spaces; url: valid-non-empty-url-potentially-surrounded-by-spaces
     urlPrefix: document-lifecycle.html
-      text: load an HTML document; url: navigate-html
+      text: shared document creation infrastructure; url: shared-document-creation-infrastructure
 spec: urlpattern; urlPrefix: https://wicg.github.io/urlpattern/
   type: interface; text: URLPattern; url: urlpattern
   type: constructor; for: URLPattern; text: "URLPattern(input, baseURL)"; url: dom-urlpattern-urlpattern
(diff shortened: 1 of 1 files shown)


speculation-rules.bs
```diff
@@ -217,6 +233,85 @@ Inside the [=prepare the script element=] algorithm we make the following change
...
+  To <dfn for="pending external speculation rule resource">fetch</dfn> a [=pending external speculation rule resource=] |pendingResource| given a [=document=] |document|:
+  1. Let |pendingResources| be |document|'s [=document/list of pending external speculation rule resources=].
+  1. [=Assert=]: |pendingResources| [=list/contains=] |pendingResource|.
+  1. [=Assert=]: |pendingResource|'s [=pending external speculation rule resource/controller=] is null.
+  1. Let |settingsObject| be |document|'s [=relevant settings object=].
+  1. Let |requestURL| be |pendingResource|'s [=pending external speculation rule resource/URL=].
+  1. Let |request| be a new [=request=] whose [=request/URL=] is |requestURL|, [=request/client=] is |settingsObject|, and [=request/mode=] is "`cors`".
+  1. Let |controller| be the result of [=fetching=] given |request| with [=fetch/processResponseConsumeBody=] set to the following steps given [=response=] |response| and null, failure, or a [=byte sequence=] |body|:
+    1. [=list/Remove=] |pendingResource| from |pendingResources|.
+    1. If any of the following conditions are true, abort these steps:
```
kjmcnee, line 291: I used the wording "abort these steps" here as that's what a number of other nested algorithms seem to use. Is there a special meaning to it, or does it just mean return?

kjmcnee:
@isaboori FYI

jeremyroman:
ad-hoc preview since the other .bs files are broken by whatwg/html#6315 (I've filed issues; nontrivial changes required): https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/kjmcnee/nav-speculation/header-spec/speculation-rules.bs

speculation-rules.bs
```diff
@@ -111,10 +121,16 @@ A <dfn>speculation rule</dfn> is a [=struct=] with the following [=struct/items=
 
 The only valid string for [=speculation rule/requirements=] to contain is "`anonymous-client-ip-when-cross-origin`".
 
+The <dfn http-header><code>Speculation-Rules</code></dfn> HTTP response header is a [=structured header=] whose value must be a [=structured header/list=] whose members must be [=structured header/strings=] which are [=valid non-empty URLs potentially surrounded by spaces=].
```
jeremyroman, line 124: I'd probably just move this definition into the "The Speculation-Rules header" section.

I also think this should probably just use ["valid URL string"](https://url.spec.whatwg.org/#valid-url-string) rather than the more flexible "potentially surrounded by space", which I think is exclusively used for stuff parsed straight out of DOM attributes.

speculation-rules.bs
```diff
@@ -217,6 +233,85 @@ Inside the [=prepare the script element=] algorithm we make the following change
     </dd>
   </dl>
 
+<h3 id="speculation-rules-header">The [:Speculation-Rules:] header</h3>
+Note: This section contains modifications to [[HTML]].
+
+<div algorithm="process the Speculation-Rules header">
```
jeremyroman, line 239: You can just write `<div algorithm>` btw, it automatically looks for the `<dfn>` to form a default name.

speculation-rules.bs
```diff
@@ -217,6 +233,85 @@ Inside the [=prepare the script element=] algorithm we make the following change
...
+  1. Let |request| be a new [=request=] whose [=request/URL=] is |requestURL|, [=request/client=] is |settingsObject|, and [=request/mode=] is "`cors`".
+  1. Let |controller| be the result of [=fetching=] given |request| with [=fetch/processResponseConsumeBody=] set to the following steps given [=response=] |response| and null, failure, or a [=byte sequence=] |body|:
+    1. [=list/Remove=] |pendingResource| from |pendingResources|.
+    1. If any of the following conditions are true, abort these steps:
+      * |body| is null.
+      * |body| is failure.
+      * |response|'s [=response/status=] is not an [=ok status=].
+      * The result of [=header list/extracting a MIME type=] from |response|'s [=response/header list=] does not have an [=MIME type/essence=] of "`application/speculationrules+json`".
+    1. Otherwise, let |text| be the result of [=UTF-8 decoding=] |body|.
+    1. Let |ruleSet| be the result of [=parsing speculation rules=] given |text| and |requestURL|.
```
jeremyroman, line 297: I think this should be `|response|'s [=response/URL=]` (i.e., the URL after following redirects). This would be a mildly interesting case to check that the impl is correct for, and write a test.

speculation-rules.bs
```diff
@@ -217,6 +233,85 @@ Inside the [=prepare the script element=] algorithm we make the following change
...
+    1. If |ruleSet| is null, abort these steps.
+      <p class="note">User agents are encouraged to report failures of the previous steps.</p>
+    1. [=list/Append=] |ruleSet| to |document|'s [=document/list of speculation rule sets=]
```
jeremyroman, line 300: super nit: period at the end of this sentence

speculation-rules.bs
```diff
@@ -217,6 +233,85 @@ Inside the [=prepare the script element=] algorithm we make the following change
...
+  1. Set |pendingResource|'s [=pending external speculation rule resource/controller=] to |controller|.
+
+</div>
+
+<div algorithm="cancel and discard a pending external speculation rule resource">
```
jeremyroman, line 305: probably can rely on the inferred algorithm name here too

speculation-rules.bs
jeremyroman: AFAIK it's equivalent to `return`, though I think there's a stronger implication that the steps aren't an algorithm and that no value is returned. I don't think there's any actual normative distinction.

jeremyroman (approved):

speculation-rules.bs
```diff
@@ -217,6 +233,85 @@ Inside the [=prepare the script element=] algorithm we make the following change
...
+<div algorithm="process pending external speculation rule resources">
+
+  To <dfn>process pending external speculation rule resources</dfn> given a [=document=] |document|:
+
+  1. Let |pendingResources| be |document|'s [=document/list of pending external speculation rule resources=].
+  1. [=list/For each=] |pendingResource| of |pendingResources|:
+    1. Optionally, [=pending external speculation rule resource/cancel and discard=] |pendingResource| given |document|, then [=iteration/continue=].
+    1. If |pendingResource|'s [=pending external speculation rule resource/controller=] is not null, then [=iteration/continue=].
+    1. Optionally, [=pending external speculation rule resource/fetch=] |pendingResource| given |document|.
+      <p class="note">The user agent may schedule the fetch of external speculation rules at its discretion.</p>
```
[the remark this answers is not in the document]
domenic: Maybe to make it extra clear, say something like. "That is, if it skips the fetch during this call of [process pending external speculation rule resources], it might do the fetch during a later call."

domenic (approved):
Very nice. I didn't realize we were allowing a list of strings. The scheduling flexibility is also well-done.

speculation-rules.bs
kjmcnee: The current implementation uses the request URL. I've filed https://crbug.com/1382183 .

speculation-rules.bs
kjmcnee: Ack.

speculation-rules.bs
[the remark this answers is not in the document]
kjmcnee: Done.

speculation-rules.bs
kjmcnee: Done.

speculation-rules.bs
kjmcnee: Done

speculation-rules.bs
kjmcnee: Changed spec to use response URL.

speculation-rules.bs
kjmcnee: Done.

speculation-rules.bs
kjmcnee: Done

speculation-rules.bs
```diff
@@ -343,12 +438,17 @@ Inside the [=prepare the script element=] algorithm we make the following change
 
 A [=document=] has a <dfn for=document export>list of speculation rule sets</dfn>, which is an initially empty [=list=].
 
+A [=document=] has a <dfn for=document export>list of pending external speculation rule resources</dfn>, which is an initially empty [=list=] of [=pending external speculation rule resources=].
+
 <!-- TODO(domfarolino): Get rid of the `data-link-type="interface"` once we fix the dfn in HTML. -->
-Periodically, for any [=document=] |document|, the user agent may [=queue a global task=] on the <a data-link-type="interface">DOM manipulation task source</a> with |document|'s [=relevant global object=] to [=consider speculation=] for |document|.
+Periodically, for any [=document=] |document|, the user agent may [=queue a global task=] on the <a data-link-type="interface">DOM manipulation task source</a> with |document|'s [=relevant global object=] to [=consider speculation=] for |document|. The user agent may also [=queue a global task=] on the [=networking task source=] with |document|'s [=relevant global object=] to [=process pending external speculation rule resources=] given |document|.
```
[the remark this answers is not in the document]
kjmcnee: Done.

PR #205, opened 2022-11-04
Add explainer for the referrer_policy key in speculation rules

kjmcnee:


triggers.md
```diff
@@ -177,6 +178,24 @@ Note that if a page is truly unsure whether a given URL will be prerendered into
...
+  "prefetch": [
+    {"source": "list",
+     "urls": ["https://en.wikipedia.org/wiki/Lethe"],
+     "referrer_policy": "no-referrer"
+    }
+  ]
+}
+```
+
+For speculation actions that would be prevented by the [**sufficiently-strict referrer policy** requirement](./fetch.md#stripping-referrer-information) on a referring page with a lax policy, this allows the referring page to set a stricter policy specifically for the speculative request.
```
domenic, line 197: Maybe mention that we don't enforce referrer policy matching, so the above example's prefetch would still be used if the user clicked on `<a href=""https://en.wikipedia.org/wiki/Lethe" referrerpolicy="unsafe-url">`?

domenic (approved):
LGTM with suggested addition.

triggers.md
kjmcnee: Done.

triggers.md
```diff
@@ -177,6 +178,29 @@ Note that if a page is truly unsure whether a given URL will be prerendered into
 
 However, in implementations such as Chromium that need the target hint, this will prerender the page twice, and thus use twice as many resources. So this is best avoided if possible.
 
+### Explicit referrer policy
+
+By default, the referring document's referrer policy is used for the speculative request. If the future navigation will use a different referrer policy (e.g. using `rel="noreferrer"`), the policy to use for the speculative request can be specified in a rule using the `"referrer_policy"` key and the desired referrer policy string.
```
jeremyroman, line 183: The "If the future navigation will use a different referrer policy" bit seems to contradict the note at the bottom. I think it scans better if you simply say "The policy to use for the speculative request can be specified...".

jeremyroman (approved):

triggers.md
kjmcnee: Done

PR #207, opened 2022-11-07
Update prerendering for HTML's changes

### LINKED ISSUE #203: prerendering.bs doesn't build after whatwg/html#6315
It probably needs some fairly significant surgery to map onto the new concepts correctly.

domenic:
Closes #203. Notable:

* This removes activating prerendering in place of navigational redirects, as that is more complicated to do with the new spec, and wasn't implemented anyway. The absence of this feature is now noted by an XXX box.

* This fixes the non-rigorous session history merging, in a nice and elegant way.

* This makes it clear that a navigable-level "loading mode" is not enough to properly specify document.prerendering or document.visibilityState. Fixing this is left for the future, with an XXX box containing some notes on how to do so.

* This changes UA-initiated prerendering to also allow prerendering into an existing top-level traversable, now that it's clear that the way we make the back button work across prerendering is via traversables, and so we can't just create a new one when the user does URL bar prerendering with an existing page open.

---

@jeremyroman, are you up for reviewing this? Sections 1, 2, and to a lesser extent 3 are the interesting ones.

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -25,43 +25,43 @@ spec: ecma262; urlPrefix: https://tc39.es/ecma262/
     text: current realm; url: current-realm
 spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
-    urlPrefix: browsers.html
-      text: creating a new top-level browsing context; url: creating-a-new-top-level-browsing-context
-      for: browsing context
-        text: remove; url: bcg-remove;
-      text: fully active; url: fully-active
-      text: top level browsing context; url:
-    urlPrefix: history.html
-      text: session history; url: session-history
-    urlPrefix: browsing-the-web.html
-      for: session history entry
-        text: document; url: she-document
-      for: navigate
-        text: mature; url: concept-navigate-mature
-      for: navigation params
-        text: request; url: navigation-params-request
-        text: reserved environment; url: navigation-params-reserved-environment
-      text: abort; for: Document; url: abort-a-document
-      text: create and initialize a Document object; url: initialise-the-document-object
-      text: history handling behavior; url: history-handling-behavior
-      text: prompt to unload; url: prompt-to-unload-a-document
-      text: refused to allow the document to be unloaded; url: refused-to-allow-the-document-to-be-unloaded
-      text: traverse the history; url: traverse-the-history
-    urlPrefix: interaction.html
-      text: system focus; url: tlbc-system-focus
-    urlPrefix: urls-and-fetching.html
-      text: parse a URL; url: parse-a-url
-      text: resulting URL record
-    urlPrefix: web-messaging.html
-      text: BroadcastChannel; url: broadcastchannel
-    urlPrefix: webappapis.html
-      text: script; url: concept-script
-    urlPrefix: workers.html
-      text: The worker's lifetime; url: the-worker's-lifetime
-      text: active needed worker; url: active-needed-worker
-    urlPrefix: media.html
-      text: playing the media resource; url: playing-the-media-resource
-      text: current playback position; url: current-playback-position
+    text: active needed worker; url: workers.html#active-needed-worker
+    text: attempt to populate the history entry's document; url: browsing-the-web.html#attempt-to-populate-the-history-entry's-document
+    text: check if unloading is user-canceled; url: browsing-the-web.html#checking-if-unloading-is-user-canceled
+    text: create a new nested navigable; url: document-sequences.html#create-a-new-nested-navigable
+    text: create a new top-level traversable; url: document-sequences.html#creating-a-new-top-level-traversable
+    text: create navigation params by fetching; url: browsing-the-web.html#create-navigation-params-by-fetching
+    text: current playback position; url: media.html#current-playback-position
+    text: destroy a top-level traversable; url: document-sequences.html#destroy-a-top-level-traversable
+    text: finalize a cross-document navigation; url: browsing-the-web.html#finalize-a-cross-document-navigation
+    text: hand-off to external software; url: browsing-the-web.html#hand-off-to-external-software
+    text: history handling behavior; url: browsing-the-web.html#history-handling-behavior
+    text: navigation and traversal task source; url: webappapis.html#navigation-and-traversal-task-source
+    text: playing the media resource; url: media.html#playing-the-media-resource
+    text: session history entry; url: browsing-the-web.html#session-history-entry
+    text: the worker's lifetime; url: workers.html#the-worker's-lifetime
+    for: Document
+      text: abort; url: document-lifecycle.html#abort-a-document
+      text: destroy; url: document-lifecycle.html#destroy-a-document
+      text: is initial about:blank; url: dom.html#is-initial-about%3Ablank
+    for: document state
+      text: initiator origin; url: browsing-the-web.html#document-state-initiator-origin
+    for: navigable
+      text: active window; url: document-sequences.html#nav-window
+      text: ongoing navigation; url: browsing-the-web.html#ongoing-navigation
+      text: parent; url: document-sequences.html#nav-parent
+    for: navigation params
+      text: request; url: browsing-the-web.html#navigation-params-request
+      text: response; url: browsing-the-web.html#navigation-params-response
+    for: session history entry
+      text: document state; url: browsing-the-web.html#she-document-state
+      text: document; url: browsing-the-web.html#she-document
+    for: traversable navigable
+      text: append session history traversal steps; url: browsing-the-web.html#tn-append-session-history-traversal-steps
+      text: current session history step; url: document-sequences.html#tn-current-session-history-step
+      text: session history entries; url: document-sequences.html#tn-session-history-entries
+    for: Window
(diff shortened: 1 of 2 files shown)


prerendering.bs
```diff
@@ -333,140 +331,123 @@ Every {{Document}} has a <dfn for="Document">post-prerendering activation steps
...
 
-  1. [=In parallel=]:
+        <p class="note">Because we have moved the entire [=session history entry=], including the contained [=session history entry/document=]'s [=Document/browsing context=], this means traversing across the associated history entries will cause a browsing context group switch, similar to what happens with the [:Cross-Origin-Opener-Policy:] header. Importantly, this will sever any opener relationships, in the same way as COOP. Such traversal includes both the actual activation process here, where we make |successorEntry| active, but also e.g. the user pressing the back button (or the page using `history.back()`) after activation, which will switch in the other direction.
 
-    1. Update the user agent's user interface to replace |predecessorBC| with |successorBC|, e.g., by updating the tab/window contents and the browser chrome.
+      1. Update the user agent's user interface to reflect this change, e.g., by updating the tab/window contents and the browser chrome.
 
-    1. [=prerendering browsing context/Finalize activation=] for |successorBC| given |predecessorBC|'s [=active document=]'s [=Document/origin=] and |unsafe time|.
```
jeremyroman, line 450: I notice this timestamp is no longer plumbed through; is this basically because the getting of the timestamp and its use are so close together in time as for the difference to be inconsequential?

prerendering.bs
```diff
@@ -478,172 +459,146 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
...
 </div>
 
 <h2 id="navigation">Navigation and session history</h2>
 
 <h3 id="navigate-activation">Allowing activation in place of navigation</h3>
 
-<div algorithm="can activate a prerender">
-  We <dfn>can activate a prerender</dfn> given a [=browsing context=] |browsingContext|, a [=history handling behavior=] |historyHandling|, a string |navigationType|, and a [=request=] |request|, if the following steps return true:
+<div>
+  To <dfn>get the matching prerendering navigable</dfn> given a [=navigable=] |navigable|, a [=URL=] |url|, a string |cspNavigationType|, and a |documentResource|, if the following steps return true:
```
jeremyroman, line 486: Type of |documentResource|? With reference to HTML it seems to be POST resource, string or null?

prerendering.bs
```diff
@@ -478,172 +459,146 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
...
-  1. Return true if all of the following are true:
+  1. If any of the following conditions hold:
 
-      * |browsingContext| is a [=top-level browsing context=]
-      * |browsingContext| is not a [=prerendering browsing context=]
-      * |browsingContext| is not for a [Fenced Frame](https://github.com/shivanigithub/fenced-frame)
-         <p class="issue">The concept of a fenced frame browsing context is not yet defined but we need to link to it once it exists.</p>
-      * |historyHandling| is "`default`" or "`replace`"
-      * |navigationType| is "`other`"
-      * |request|'s [=request/method=] is \``GET`\`
```
jeremyroman, line 515: Is this obsoleted by the |documentResource| check? If so, why? I'm guessing that GET requests don't have them, POST requests always do, and no other type is possible in navigation? But it's not obvious.

prerendering.bs
```diff
@@ -478,172 +459,146 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
...
-      * |browsingContext|'s [=active document=]'s [=Document/prerendering browsing contexts map=][|request|'s [=request/URL=]] [=map/exists=] and is not [=prerendering browsing context/empty=]
-      * The result of calling [=should navigation request of type be blocked by content security policy?=] given |request| and <var ignore>navigationType</var> is "`Allowed`".
```
jeremyroman, line 517: It's not obvious to me why this was removed; can you clarify?

prerendering.bs
```diff
@@ -478,172 +459,146 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
...
 
-    1. [=prerendering browsing context/Activate=] |browsingContext| in place of <var ignore>sourceBrowsingContext</var> given |historyHandling|.
+  Append the following steps after the first sub-step under "While true:":
 
-    1. Return.
-
-  1. If |browsingContext| is a [=prerendering browsing context=]  and <var ignore>currentURL</var>'s [=url/origin=] is not [=/same site=] with |incumbentNavigationOrigin|, then:
+  1. If |navigable| is a [=prerendering navigable=] and <var ignore>currentURL</var>'s [=url/origin=] is not [=/same site=] with |initiatorOrigin|, then:
 
-    1. Run the [=environment discarding steps=] for <var ignore>navigationParams</var>'s [=navigation params/reserved environment=].
```
jeremyroman, line 556: Why is it no longer necessary to run the environment discarding steps?

prerendering.bs
```diff
@@ -478,172 +459,146 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
...
     1. Return.
 </div>
 
-<div algorithm="process a navigate URL scheme patch">
-  In <a spec=HTML>process a navigate URL scheme</a>, insert the following step before the step which displays inline content:
+<div algorithm="hand-off to external software patch">
+  In [=hand-off to external software=], prepend the following step:
 
-  1. Otherwise, if <var ignore>browsingContext</var> is a [=prerendering browsing context=], then return.
+  1. If <var ignore>navigable</var> is a [=prerendering navigable=], then return without invokign the external software package.
```
jeremyroman, line 570:
```suggestion
  1. If <var ignore>navigable</var> is a [=prerendering navigable=], then return without invoking the external software package.
```

prerendering.bs
domenic: I think the timestamp may have been previously separate because, before https://github.com/whatwg/html/pull/7720 , HTML had some special handling for unload timing? But yes, upon reevaluating this, it seems very unnecessary to grab the timestamp early. Also, I looked at the Chromium code, and the timestamp is computed during the task posted to the renderer process, at the same time we set `document.prerendering` and such.

prerendering.bs
domenic: Yes. Every POST request will have a POST resource as |documentResource|, in particular. So that subsumes the old check.

The case where |documentResource| is a string, occurs when navigating srcdoc iframes. We also want to bail out there, although we should never have a non-HTTP(S) URL in the map anyway.

prerendering.bs
domenic: Now that `navigate-to` is dead, this only matters for CSP `form-action` and `javascript:` URLs. We're already excluding those more generally, so I removed it.

This is not perfectly future proof, so maybe I should add it back. But:

- Doing so would be a bit of a pain, as we'd need to construct a synthetic response (since that's what CSP expects)
- It's arguably reasonable to block such navigations at prerendering time, and not necessarily check a second time when activating. Maybe?

I'll add it back if you don't find these excuses convincing :)

prerendering.bs
domenic: I think this step was incorrect in the first place. The environment discarding steps seem to only be used during redirects; overall failures to navigate, e.g. 204 or error documents, do not currently trigger these in HTML. I was thinking maybe they should? But that'd be a more general investigation.

prerendering.bs
```diff
@@ -478,172 +459,146 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
...
+  1. If |matchingPrerenderedNavigable| is not null, then:
 
-    1. Wait until all navigation attempts of |successorBC| have been [=navigate/mature|matured=].
+    1. Wait until either |matchingPrerenderedNavigable|'s [=navigable/active document=]'s [=Document/is initial about:blank=] is false, or |matchingPrerenderedNavigable| is [=destroy a top-level traversable|destroyed=].
 
-    1. If |successorBC| has not been [=discard|discarded=], then:
+    1. If |matchingPrerenderedNavigable| was not destroyed, then:
 
-      1. [=prerendering browsing context/Activate=] |successorBC| in place of |browsingContext| given |historyHandling|.
+      1. [=prerendering traversable/Activate=] |matchingPrerenderedNavigable| in place of |navigable| given <var ignore>historyHandling</var>.
```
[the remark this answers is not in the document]
domenic: Good catch. I think this is leftover from before I made activate happen entirely in the "in parallel" section.

I think the right behavior here is pretty unclear because of the WebDriver BiDi stuff. Also, UA-initiated prerendering is a bit more annoying. But threading the navigation ID through seems at least a bit more correct. I'll open a WebDriver BiDi issue and in the meantime thread things through, even though that'll give somewhat strange behavior.

prerendering.bs
jeremyroman: It wasn't totally obvious to me that a "document resource" refers to a POST response body (and associated metadata), though now that I know that it's clear. Could go either way.

jeremyroman (approved):
whoops, lost track of this and noticed only when I was annoyed that I fixed prefetch.bs and `make` still didn't succeed 😆

PR #208, opened 2022-11-08
Add "Content-Security-Policy" to speculation rules explainer

toyoshim:
The new section explains how the prerender and prefetch follow the Content-Security-Policy. They are not restricted by `script-src`, but done by `prefetch-src`. This makes speculationrules safer and flexible.


toyoshim:
let me re-issue this PR after setting up my W3C account :)

PR #212, opened 2022-11-17
Update prefetch spec after whatwg/html#6315.

jeremyroman:
This is a fairly comprehensive rewrite so hopefully it both respects the new navigables/etc infrastructure and reflects the previously specified functionality.

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -242,7 +242,7 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
     <p class="issue">It might be possible to use cache response headers to determine when a response can be used multiple times, but given the short lifetime of the prefetch buffer it's unclear whether this is worthwhile.</p>
 </div>
 
-<div>
+<div algorithm>
     To <dfn export>create navigation params from a prefetch record</dfn> given a [=navigable=] |navigable|, a [=document state=] |documentState|, a [=navigation id=] |navigationId|, a {{NavigationTimingType}} |navTimingType|, a [=request=] |request|, a [=prefetch record=] |record|, a [=target snapshot params=] |targetSnapshotParams|, and a [=source snapshot params=] |sourceSnapshotParams|, perform the following steps.
 
     1. Let |responseOrigin| be null.
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -209,7 +209,7 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
 </div>
 
 <div algorithm="perform prefetch response checks">
-    To <dfn export>perform prefetch response checks</dfn> given a [=navigable=] |navigable|, a [=document state=] |documentState| (this has a history policy container), a [=navigation id=] |navigationId|, a [=request=] |request|, a [=prefetch record=] |record|, a [=sandboxing flag set=] |sandboxFlags|, and a [=policy containers=] |sourcePolicyContainer|, perform the following steps.
+    To <dfn export>perform prefetch response checks</dfn> given a [=navigable=] |navigable|, a [=document state=] |documentState| (this has a history policy container), a [=navigation id=] |navigationId|, a {{NavigationTimingType}} |navTimingType|, a [=request=] |request|, a [=prefetch record=] |record|, a [=sandboxing flag set=] |sandboxFlags|, and a [=policy containers=] |sourcePolicyContainer|, perform the following steps.
     TODO(jbroman): which sandboxing flag set is this, the one from source snapshot params or target snapshot params?
 
     1. Let |responseOrigin| be null.
@@ -217,7 +217,7 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
     1. Let |coopEnforcementResult| b e a new [=cross-origin opener policy enforcement result=], with:
         :  [=cross-origin opener policy enforcement result/needs a browsing context group switch=]
         :: false
-        :  [=cross origin opener policy enforcement result/would need a browsing context group switch due to report-only=]
+        :  [=cross-origin opener policy enforcement result/would need a browsing context group switch due to report-only=]
         :: false
         :  [=cross-origin opener policy enforcement result/url=]
         :: |navigable|'s [=navigable/active document=]'s [=Document/URL=]
@@ -264,13 +264,13 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
         :  [=navigation params/navigable=]
         :: |navigable|
         :  [=navigation params/navigation timing type=]
-        :: "`navigate`"
+        :: |navTimingType|
         :  [=navigation params/fetch controller=]
-        :: null
+        :: |prefetchRecord|'s [=prefetch record/fetch controller=]
         :  [=navigation params/commit early hints=]
         :: null
 
-        TODO(jbroman): need to correct plumbing for: navTimingType, fetchController, commitEarlyHints
+            <p class="note">This implies that early hints delivered for prefetched documents won't be processed. This could be revised in the future if it is common for there to be useful resource hints which are not repeated in the main response headers. Since prefetches aren't currently served until the response headers arrive, early hints would not be processed any earlier than ordinary response headers. It might be possible for a future specification to allow early hints found during prefetching to be used in some way.</p>
     1. Return |navigationParams|.
 </div>
 
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -401,7 +401,7 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
             1. Set |request|'s [=request/replaces client id=] to <var ignore>navigable</var>'s [=navigable/active document=]'s [=relevant settings object=]'s [=environment/id=].
             1. Let |prefetchRecord| be the result of [=finding a matching prefetch record=] given <var ignore>navigable</var>'s [=navigable/active document=], <var ignore>entry</var>'s [=session history entry/URL=], and <var ignore>targetSnapshotParams</var>'s [=target snapshot params/sandboxing flags=].
             1. If <var ignore>documentResource</var> is null and |prefetchRecord| is not null:
-                1. Set <var ignore>navigationParams</var> to the result of [=creating navigation params from a prefetch record=] given <var ignore>navigable</var>, <var ignore>entry</var>'s [=session history entry/document state=], <var ignore>navigationId</var>, <var ignore>navTimingType</var>, <var ignore>request</var> |prefetchRecord|, <var ignore>targetSnapshotParams</var>, and <var ignore>sourceSnapshotParams</var>.
+                1. Set <var ignore>navigationParams</var> to the result of [=creating navigation params from a prefetch record=] given <var ignore>navigable</var>, <var ignore>entry</var>'s [=session history entry/document state=], <var ignore>navigationId</var>, <var ignore>navTimingType</var>, <var ignore>request</var>, |prefetchRecord|, <var ignore>targetSnapshotParams</var>, and <var ignore>sourceSnapshotParams</var>.
                 1. If |prefetchRecord|'s [=prefetch record/isolated partition key=] is not null, then [=copy prefetch cookies=] given |prefetchRecord|'s [=prefetch record/isolated partition key=] and <var ignore>navigationParams</var>'s [=navigation params/reserved environment=].
 
                     <div class="note">This copy is complete before continuing, in the sense that subresource fetches, {{Document/cookie|document.cookie}}, etc. can observe the cookies.</div>
@@ -443,7 +443,7 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
         1. If the result of [=should navigation request of type be blocked by Content Security Policy?=] given |request| and |cspNavigationType| is "`Blocked`", then set |response| to a [=network error=] and [=iteration/break=]. [[CSP]]
         1. If |shouldBlockNavigationRequest| is present and its result given |request| is "`Blocked`", then set |response| to a [=network error=] and [=iteration/break=]. <!-- these are new steps to allow requests to be blocked for additional reasons -->
         1. Set |response| to null.
-        1. If |fetchController| is null, then set |fetchController| to the result of [=fetching=] |request|, with processEarlyHintsResponse set to |processEarlyHintsResponse| as defined below, processResponse set to |processResponse| as defined below, and useParallelQueue set to true.
+        1. If |fetchController| is null, then set |fetchController| to the result of [=fetching=] |request|, with <i>[=fetch/processEarlyHintsResponse=]</i> set to |processEarlyHintsResponse| as defined below, <i>[=fetch/processResponse=]</i> set to |processResponse| as defined below, and <i>[=fetch/useParallelQueue=]</i> set to true.
 
             Let |processEarlyHintsResponse| be the following algorithm given a [=response=] |earlyResponse|:
                 1. If |processEarlyHints| is true and |commitEarlyHints| is null, then set |commitEarlyHints| to the result of [=processing early hint headers=] given |earlyResponse| and |request|'s [=request/reserved client=]. <!-- this has an additional condition to suppress extra fetches due to early hints -->
@@ -553,7 +553,7 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
         </div>
     1. [=Assert=]: |prefetchRecord|'s [=prefetch record/URL=]'s [=url/scheme=] is an [=HTTP(S) scheme=].
     1. [=list/Append=] |prefetchRecord| to |document|'s [=Document/prefetch records=]
-    1. Set |prefetchRecord|'s [=prefetch record/sandboxing flag set=] to the result of [=determining the creation sandboxing flags=] given |document|'s [=Document/browsing context=] and |document|'s [=node navigable=]'s [=navigable/container=].
+    1. Set |prefetchRecord|'s [=prefetch record/sandboxing flag set=] to the result of [=determining the creation sandboxing flags=] for |document|'s [=Document/browsing context=] given |document|'s [=node navigable=]'s [=navigable/container=].
     1. Let |isolationOrigin| be null.
     1. Let |documentState| be a new [=document state=] with
         :  [=document state/request referrer policy=]
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -425,13 +425,10 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
 
 <hr>
 
-Add the following [=struct/item=] to [=session history entry=]:
-* <dfn for="session history entry">prefetch record</dfn>, a [=prefetch record=] or null. If non-null, it indicates that a prefetch, not a normal navigation, is in progress.
-
-<div algorithm>
+<div algorithm="create navigation params by fetching">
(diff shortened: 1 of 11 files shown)


jeremyroman:
It builds locally; once the prerendering.bs fixes land I'll retrigger the preview bot. In the meantime at least this still works: https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jeremyroman/nav-speculation/prefetch-6315/prefetch.bs

jeremyroman:
@domenic ptal when you get the chance? I know it's kinda a huge diff.

prefetch.bs
```diff
@@ -252,233 +325,283 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
...
-            1. Add a parameter whose key is <a for="Sec-Purpose prefetch" lt="anonymous-client-ip">"`anonymous-client-ip`"</a> and whose value is true to the "`prefetch`" token in |purpose|.
-            1. The user agent must use a [=connection=] which anonymizes the client IP address (e.g., using a proxy) when fetching |request|, or return "`Blocked`".
+<hr>
 
-                <p class="issue">At the moment, how IP anonymization is achieved is handwaved. This will probably be done in an [=implementation-defined=] manner using some kind of proxy or relay. Ideally this would be plumbed down to [=obtain a connection=], and possibly even the mechanism could be further standardized.</p>
-        1. [=header list/Set a structured field value=] given (<a http-header>`` `Sec-Purpose` ``</a>, |purpose|) in |request|'s [=request/header list=].
+<div algorithm="attempt to populate the history entry's document">
+    In <a spec=HTML>attempt to populate the history entry's document</a>, replace the step which invokes <a spec=HTML>create navigation params by fetching</a> with the following:
```
domenic, line 409: To be clear, this means we're allowing prefetches to be used to fulfill traversals, not just push/replace navigations. I think that's OK but I wanted to make sure you agree. (If not, we should probably add some sort of guard.)

prefetch.bs
```diff
@@ -252,233 +325,283 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
...
+            1. Set |request|'s [=request/replaces client id=] to <var ignore>navigable</var>'s [=navigable/active document=]'s [=relevant settings object=]'s [=environment/id=].
+            1. Let |prefetchRecord| be the result of [=finding a matching prefetch record=] given <var ignore>navigable</var>'s [=navigable/active document=], <var ignore>entry</var>'s [=session history entry/URL=], and <var ignore>targetSnapshotParams</var>'s [=target snapshot params/sandboxing flags=].
+            1. If <var ignore>documentResource</var> is null and |prefetchRecord| is not null:
+                1. Set <var ignore>navigationParams</var> to the result of [=creating navigation params from a prefetch record=] given <var ignore>navigable</var>, <var ignore>entry</var>'s [=session history entry/document state=], <var ignore>navigationId</var>, <var ignore>navTimingType</var>, <var ignore>request</var> |prefetchRecord|, <var ignore>targetSnapshotParams</var>, and <var ignore>sourceSnapshotParams</var>.
```
domenic, line 421:
```suggestion
                1. Set <var ignore>navigationParams</var> to the result of [=creating navigation params from a prefetch record=] given <var ignore>navigable</var>, <var ignore>entry</var>'s [=session history entry/document state=], <var ignore>navigationId</var>, <var ignore>navTimingType</var>, <var ignore>request</var>, |prefetchRecord|, <var ignore>targetSnapshotParams</var>, and <var ignore>sourceSnapshotParams</var>.
```

prefetch.bs
```diff
@@ -252,233 +325,283 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
...
+
+    To <dfn>create navigation params by fetching</dfn> given a [=request=] |request|, a [=session history entry=] |entry|, a [=cross-origin opener policy enforcement result=] |coopEnforcementResult|, a navigable |navigable|, a [=source snapshot params=] |sourceSnapshotParams|, a [=target snapshot params=] |targetSnapshotParams|, a string |cspNavigationType|, a [=navigation ID=] or null |navigationId|, a {{NavigationTimingType}} |navTimingType|, an optional [=opaque origin=] or null (default null) |isolationOrigin|, an optional boolean |processEarlyHints| (default true), an optional boolean |reportCOOPViolations| (default true), an optional algorithm |shouldBlockNavigationRequest|, and an optional algorithm |processNavigationResponse|, perform the following steps. If given, |shouldBlockNavigationRequest| must be an algorithm accepting a [=request=] and returning a "`Blocked`" or "`Allowed`". They return a [=navigation params=], a [=non-fetch scheme navigation params=], or null. If given, |processNavigationResponse| must be an algorithm accepting a [=request=] and [=response=].
```
domenic, line 436: Oof, this got complex.

Some parameter declaration tweaks that might help readability a bit:

- Non-optional parameters in the order _entry_, _navigable_, _request_, _coopEnforcementResult_, _sourceSnapshotParams_, _targetSnapshotParams_, _cspNavigationType_, _navigationId_, _navTimingType_ (similar to today's HTML spec)

- Optional params can be named, i.e. declared here with something like `<dfn for="create navigation params by fetching">|reportCOOPViolations|</dfn>` and then referenced with `<i>[=create navigation params by fetching/reportCOOPViolations=]</i> set to true`.

Another option that might be worth exploring is if we could make the algorithm clearly have a "prefetch mode" and a "normal mode". E.g. instead of having fully-customizable _processNavigationResponse_, pass in a prefetch record or null, and if it's non-null, call "Update the response" directly. If this were able to collapse the 5 new optional params into 1 or 2 new ones, I think that would be a big win. Otherwise, it kind of seems like all 5 can vary independently, which is a lot of complexity for the reader to keep track of.

prefetch.bs
```diff
@@ -252,233 +325,283 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
...
+            1. Set |request|'s [=request/reserved client=] to null.
+            1. Set |commitEarlyHints| to null.
+        1. If |request|'s [=request/reserved client=] is null, then set |request|'s [=request/reserved client=] to the result of [=creating a reserved client=] given |navigable|, |currentURL| and |isolationOrigin|.
+        1. If the result of [=should navigation request of type be blocked by Content Security Policy?=] given |request| and |cspNavigationType| is "`Blocked`", then set |response| to a [=network error=] and [=iteration/break=]. [[CSP]]
+        1. If |shouldBlockNavigationRequest| is present and its result given |request| is "`Blocked`", then set |response| to a [=network error=] and [=iteration/break=]. <!-- these are new steps to allow requests to be blocked for additional reasons -->
+        1. Set |response| to null.
+        1. If |fetchController| is null, then set |fetchController| to the result of [=fetching=] |request|, with processEarlyHintsResponse set to |processEarlyHintsResponse| as defined below, processResponse set to |processResponse| as defined below, and useParallelQueue set to true.
```
domenic, line 463:
```suggestion
        1. If |fetchController| is null, then set |fetchController| to the result of [=fetching=] |request|, with <i>[=fetch/processEarlyHintsResponse=]</i> set to |processEarlyHintsResponse| as defined below, <i>[=fetch/processResponse=]</i> set to |processResponse| as defined below, and <i>[=fetch/useParallelQueue=]</i> set to true.
```

prefetch.bs
```diff
@@ -252,233 +325,283 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
...
+            :  [=document state/navigable target name=]
+            :: |oldDocState|'s [=document state/navigable target name=]
+        1. If |locationURL|'s [=url/scheme=] is not an [=HTTP(S) scheme=], then:
+            1. Set |entry|'s [=session history entry/document state=]'s [=document state/resource=] to null.
+            1. [=iteration/Break=].
+        1. Set |currentURL| to |locationURL|.
+        1. Set |entry|'s [=session history entry/URL=] to |currentURL|.
+    1. If |locationURL| is a [=request/URL=] whose [=url/scheme=] is not a [=fetch scheme=], then return a new [=non-fetch scheme navigation params=], with:
+        :  [=non-fetch scheme navigation params/initiator origin=]
+        :: |request|'s [=request/current URL=]
```
domenic, line 513: Gotta port over https://github.com/whatwg/html/commit/3769dd08c5bd43c287e7923f972bdd5f37f9e734

prefetch.bs
```diff
@@ -252,233 +325,283 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
...
+    1. Let |sourceSnapshotParams| be the result of [=snapshotting source snapshot params=] given |document|.
+    1. Let |targetSnapshotParams| be the result of [=snapshotting target snapshot params=] given |document|'s [=node navigable=].
+    1. Let |topLevelOrigin| be |prefetchRecord|'s [=prefetch record/URL=]'s [=url/origin=] if |document|'s [=node navigable=] is a [=top-level traversable=], and |document|'s [=relevant settings object=]'s [=environment/top-level origin=] otherwise.
+    1. Let |topLevelSite| be the result of [=obtaining a site=] given |topLevelOrigin|.
+    1. Let |secondKey| be null or an [=implementation-defined=] value.
+    1. Let |partitioningScheme| be "`same-partition`" if |sourcePartitionKey| is equal to (|topLevelSite|, |secondKey|), and "`uncredentialed`" otherwise.
+
+        <div class="note">
+            This determines whether the navigation would use the same network partition key. If it would, the prefetch is restricted to the same partition, and redirects which would leave the partition cause the prefetch to fail.
```
domenic, line 567:
```suggestion
            <p>This determines whether the navigation would use the same network partition key. If it would, the prefetch is restricted to the same partition, and redirects which would leave the partition cause the prefetch to fail.
```

prefetch.bs
```diff
@@ -252,233 +325,283 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
...
 
+            Otherwise it is uncredentialed, and redirects which would return to the original partition (thus ought to have credentials) will cause prefetch to fail.
```
domenic, line 569:
```suggestion
            <p>Otherwise it is uncredentialed, and redirects which would return to the original partition (thus ought to have credentials) will cause prefetch to fail.
```

prefetch.bs
```diff
@@ -252,233 +325,283 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
...
-    1. Let |isolatedEnvironment| be a new [=environment=] whose [=environment/id=] is a unique opaque string, [=environment/target browsing context=] is |browsingContext|, [=environment/creation URL=] is `about:blank`, [=environment/top-level creation URL=] is `about:blank`, and [=environment/top-level origin=] is |isolationOrigin|.
-    1. Let |isolatedPartitionKey| be the result of [=determining the network partition key=] given |isolatedEnvironment|.
-    1. Let |originalPartitionKey| be the result of [=determining the network partition key=] given |document|'s [=relevant settings object=].
-    1. Set |prefetchRecord|'s [=prefetch record/isolated partition key=] to |isolatedPartitionKey|.
-    1. Let |request| be a [=request=] as follows:
-
-        :  [=request/URL=]
+    1. [=list/Append=] |prefetchRecord| to |document|'s [=Document/prefetch records=]
+    1. Set |prefetchRecord|'s [=prefetch record/sandboxing flag set=] to the result of [=determining the creation sandboxing flags=] given |document|'s [=Document/browsing context=] and |document|'s [=node navigable=]'s [=navigable/container=].
```
domenic, line 573:
```suggestion
    1. Set |prefetchRecord|'s [=prefetch record/sandboxing flag set=] to the result of [=determining the creation sandboxing flags=] for |document|'s [=Document/browsing context=] given |document|'s [=node navigable=]'s [=navigable/container=].
```

prefetch.bs
```diff
@@ -252,233 +325,283 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
...
 
+            <div class="note">This is used to ensure a distinct network partition key is used.</div>
+
+        <div class="note">Though credentials are still included, they will be isolated such that no credentials are present to begin with.</div>
     1. Let |originsWithConflictingCredentials| be an empty [=ordered set=].
-    1. Let |shouldBlockNavigationRequest| be the following steps, given [=request=] |request|, [=string=] |navigationType| and [=environment=] |environment|:
-        1. [=Assert=]: |request|'s [=request/reserved client=] is |isolatedEnvironment| and not |environment|.
-        1. [=Assert=]: |navigationType| is "`other`".
+    1. Let |shouldBlockNavigationRequest| be an algorithm which, given a [=request=] |request|, performs the following steps.
         1. Let |purpose| be a <a spec="RFC8941">List</a> containing the <a spec="RFC8941">Token</a> "`prefetch`".
```
domenic, line 603:
```suggestion
        1. Let |purpose| be a <a spec="RFC8941">List</a> containing the <a spec="RFC8941">Token</a> `prefetch`.
```

prefetch.bs
```diff
@@ -252,233 +325,283 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
...
-        1. If |anonymizationPolicy| [=prefetch IP anonymization policy/requires anonymity=] for |request|, then:
+        1. If |prefetchRecord|'s [=prefetch record/anonymization policy=] [=prefetch IP anonymization policy/requires anonymity=] for |request|, then:
             1. Add a parameter whose key is <a for="Sec-Purpose prefetch" lt="anonymous-client-ip">"`anonymous-client-ip`"</a> and whose value is true to the "`prefetch`" token in |purpose|.
```
domenic, line 605:
```suggestion
            1. Add a parameter whose key is <a for="Sec-Purpose prefetch" lt="anonymous-client-ip">"`anonymous-client-ip`"</a> and whose value is true to the `prefetch` token in |purpose|.
```

prefetch.bs
```diff
@@ -488,61 +611,36 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
...
-        1. If |originalPartitionKey| is equal to |hypotheticalPartitionKey|, then return "`Blocked`".
-
-            <div class="note">The prefetch would end up in the same partition as the prefetch came from in this case. |environment| represents the environment that would ordinarily be used during navigation fetch.</div>
-        1. If there are [=credentials=] associated with |request|'s [=request/current URL=] and |hypotheticalPartitionKey|, then [=set/append=] |request|'s [=request/current URL=]'s [=url/origin=] to |originsWithConflictingCredentials|.
-        1. If |request|'s [=request/URL=] is not [=potentially trustworthy URL|potentially trustworthy=], then return "`Blocked`".
+        1. Let |proposedPartitionKey| be the result of [=determining the network partition key=] given |request|'s [=request/reserved client=].
+        1. If |partitioningScheme| is "`same-partition`" and |proposedPartitionKey| is not equal to |sourcePartitionKey|, then return "`Blocked`".
+        1. Otherwise, if |partitioningScheme| is "`uncredentialed`" and |proposedPartitionKey| is equal to |sourcePartitionKey|, then return "`Blocked`".
```
domenic, line 616: I think this blocks same-origin prefetches on totally-uncredentialed origins. Is that intentional?

domenic (commented):
Nice work!

prefetch.bs
jeremyroman: Any particular reason? The second `<p>`, at least, is generated by Bikeshed automatically. Going to commit them since I don't have a strong reason to oppose, but it seems unnecessary.

prefetch.bs
jeremyroman: Done.

jeremyroman (commented):
Responding to most comments. The question about whether to just make "create navigation params by fetching" explicitly aware of prefetch is an interesting one which I'll need to think a little more about.

prefetch.bs
```diff
@@ -199,40 +242,70 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
...
+        :: |navigable|'s [=navigable/active document=]'s [=Document/URL=]
+        :  [=cross-origin opener policy enforcement result/origin=]
+        :: |navigable|'s [=navigable/active document=]'s [=Document/origin=]
+        :  [=cross-origin opener policy enforcement result/cross-origin opener policy=]
+        :: |navigable|'s [=navigable/active document=]'s [=Document/cross-origin opener policy=]
+        :  [=cross-origin opener policy enforcement result/current context is navigation source=]
+        :: true if |navigable|'s [=navigable/active document=]'s [=Document/origin=] is [=same origin=] with |documentState|'s [=document state/initiator origin=], and false otherwise.
     1. Let |finalSandboxFlags| be an empty [=sandboxing flag set=].
-    1. Let |hasCrossOriginRedirects| be false.
+    1. Let |responsePolicyContainer| be null.
```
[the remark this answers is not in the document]
domenic: I have no idea what I was thinking here.

prefetch.bs
domenic: Fun!

Looks like the test got reverted; I hope we can add it back. It's weird that it would be failing on Macs only?

prefetch.bs
domenic: I'm pretty sure I looked at the generated source and it looked weird due to spacing issues, due to the lack of first `<p>`.

domenic (commented):
Looks like there's still some substantial feedback outstanding. Let me know if you want to box those away into `<p class="issue">` boxes so we can move forward; I think that's very reasonable.

prefetch.bs
jeremyroman: Tried this. It adds a bit more bulk to "create navigation params by fetching" than I'd like, though maybe some bits can be factored back out again. (I kinda leaned toward keeping them inline at least for now, in part because some of the handwaves like "use a connection which is IP-private!" probably ought not to be hidden in some "can this request continue prefetch" predicate.)

prefetch.bs
```diff
@@ -252,297 +324,315 @@ A <dfn export>cross-origin prefetch IP anonymization policy</dfn> has an <dfn ex
...
 
-            <div class="note">This is intended to both reduce the likelihood of prefetch traffic being visible to an on-path attacker, and to encourage the use of cryptographic schemes over public networks.</div>
-        1. If |request| cannot be fetched given |anonymizationPolicy| for an [=implementation-defined=] reason, then return "`Blocked`".
+<hr>
 
-            <div class="note">This explicitly acknowledges that implementations might have additional restrictions. For instance, anonymized traffic might not be possible to some hosts, such as those that are not publicly routable and those that have <a href="https://buettner.github.io/private-prefetch-proxy/traffic-advice.html">traffic advice</a> declining private prefetch traffic.
-        1. [=redirect chain/Append=] |request| to |prefetchRecord|'s [=prefetch record/redirect chain=].
-        1. Return "`Allowed`".
+Add the following [=struct/item=] to [=session history entry=]:
+* <dfn for="session history entry">prefetch record</dfn>, a [=prefetch record=] or null. If non-null, it indicates that a prefetch, not a normal navigation, is in progress.
```
domenic, line 429: It feels weird to store this in session history entry. It seems like it doesn't need to be there long term, and instead is just used because the relevant algorithms take a SHE. Can we instead just thread it to the relevant algorithms as an independent, sometimes-null parameter?

domenic (approved):
This looks great, I think merging into "create navigation params by fetching" worked well.

Giving you to the LGTM so that you can merge this at will, but I guess there might still be something worth checking about how we thread through the prefetch record.

prefetch.bs
jeremyroman: The entire SHE here is admittedly a little fictitious because we're not actually putting it in session history just yet.

Made the suggested change.

PR #214, opened 2022-11-18
Update speculation rules spec for navigables

kjmcnee:
The recent update to prerendering.bs for the navigable changes to the HTML spec requires additional changes in speculation-rules.bs for it to build and for the links to work.

Some of the links to the html spec still work due to redirects. We now link to the new concepts directly. We also rename "target browsing context name hint" to "target navigable name hint" accordingly.

We also fix an issue where the prerender case of consider speculation was iterating over prefetch candidates.


domenic (approved):
Thanks so much for spotting and fixing this!

PR #216, opened 2022-11-21
Add spec for referrer policy key

kjmcnee:
We parse the "referrer_policy" key from the given speculation rules input, and compute the referrer policy to use based on the rule and possibly a matching <a> or <area> element.

We pass this information through to a prerendering navigation, but leave the prefetch case for a later PR.

### DIFF
--- a/speculation-rules.bs
+++ b/speculation-rules.bs
@@ -380,7 +380,6 @@ add the following step
   1. Let |referrerPolicy| be the empty string.
   1. If |input|["`referrer_policy`"] [=map/exists=]:
     1. If |input|["`referrer_policy`"] is not a [=referrer policy=], then return null.
-    1. If |input|["`referrer_policy`"] is the empty string, then return null.
     1. Set |referrerPolicy| to |input|["`referrer_policy`"].
   1. Return a [=speculation rule=] with [=speculation rule/URLs=] |urls|, [=speculation rule/predicate=] |predicate|, [=speculation rule/requirements=] |requirements|, [=speculation rule/target navigable name hint=] |targetHint|, and [=speculation rule/referrer policy=] |referrerPolicy|.
 </div>
--- a/speculation-rules.bs
+++ b/speculation-rules.bs
@@ -17,6 +17,8 @@ Boilerplate: omit conformance
 <pre class="link-defaults">
 spec:html; type:element; text:a
 spec:html; type:element-attr; for:a; text:href
+spec:html; type:element-attr; for:a; text:referrerpolicy
+spec:html; type:element-attr; for:a; text:rel
 spec:html; type:element; text:link
 spec:html; type:element; text:script
 spec:selectors-4; type:selector; text::link
@@ -43,6 +45,12 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: rules for choosing a navigable; url: the-rules-for-choosing-a-navigable
     urlPrefix: semantics.html
       text: get an element's target; url: get-an-element's-target
+    urlPrefix: links.html
+      text: following a hyperlink; url: following-hyperlinks-2
+  type: attr-value
+    urlPrefix: links.html
+      for: a/rel
+        text: noreferrer; url: link-type-noreferrer
 spec: urlpattern; urlPrefix: https://wicg.github.io/urlpattern/
   type: interface; text: URLPattern; url: urlpattern
   type: constructor; for: URLPattern; text: "URLPattern(input, baseURL)"; url: dom-urlpattern-urlpattern
@@ -115,6 +123,7 @@ A <dfn>speculation rule</dfn> is a [=struct=] with the following [=struct/items=
 * <dfn for="speculation rule">predicate</dfn>, a [=document rule predicate=] or null
 * <dfn for="speculation rule">requirements</dfn>, an [=ordered set=] of [=strings=]
 * <dfn for="speculation rule">target navigable name hint</dfn>, a [=string=] or null
+* <dfn for="speculation rule">referrer policy</dfn>, a [=referrer policy=]
 
 The only valid string for [=speculation rule/requirements=] to contain is "`anonymous-client-ip-when-cross-origin`".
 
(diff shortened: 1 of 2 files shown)


speculation-rules.bs
```diff
@@ -368,7 +377,12 @@ add the following step
   1. If |input|["`target_hint`"] [=map/exists=]:
     1. If |input|["`target_hint`"] is not a [=valid navigable target name or keyword=], then return null.
     1. Set |targetHint| to |input|["`target_hint`"].
-  1. Return a [=speculation rule=] with [=speculation rule/URLs=] |urls|, [=speculation rule/predicate=] |predicate|, [=speculation rule/requirements=] |requirements|, and [=speculation rule/target navigable name hint=] |targetHint|.
+  1. Let |referrerPolicy| be the empty string.
+  1. If |input|["`referrer_policy`"] [=map/exists=]:
+    1. If |input|["`referrer_policy`"] is not a [=referrer policy=], then return null.
+    1. If |input|["`referrer_policy`"] is the empty string, then return null.
```
kjmcnee, line 383: domenic, offline you mentioned using the empty string to indicate falling back. Is that internally to the spec, or developer facing as well?

For example, should
```
{ "prefetch": [{
  "source": "list",
  "urls": ["https://example.com/"],
  "referrer_policy": ""
 }]
}
```
be treated as if the "referrer_policy" key were omitted, or should this be treated as an invalid rule? The implementation currently does the latter.

kjmcnee:
@a4sriniv FYI

speculation-rules.bs
```diff
@@ -43,6 +45,12 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: rules for choosing a navigable; url: the-rules-for-choosing-a-navigable
     urlPrefix: semantics.html
       text: get an element's target; url: get-an-element's-target
+    urlPrefix: links.html
+      text: following a hyperlink; url: following-hyperlinks-2
+  type: attr-value
+    urlPrefix: links.html
+      for: a/rel
+        text: noreferrer; url: link-type-noreferrer
```
domenic, line 53: Hmm, this shouldn't be necessary; HTML has it marked up correctly. But `bikeshed refs --text noreferrer --type attr-value` doesn't give any results. So, not sure what's going on there. No action for you.

speculation-rules.bs
domenic: I find it most elegant if we allow the empty string here. Everywhere else on the platform, it means to fall back. So I think it should here, too. It also gives us a nice default value in the parsed data model, such that the empty string is the same as omitted.

Finally, it's possible this could be useful for authors generating speculation rules in JavaScript (or a server-side language). The following code:

```js
const rule = {
  source: "list",
  urls: computeURLs(),
  referrer_policy: computeReferrerPolicy() // might return ""
};
```

is a bit simpler than


```js
const rule = {
  source: "list",
  urls: computeURLs(),
  referrer_policy: computeReferrerPolicy() // might return ""
};

if (rule.referrer_policy === "") {
  delete rule.referrer_policy;
}
```

speculation-rules.bs
```diff
@@ -368,7 +377,12 @@ add the following step
   1. If |input|["`target_hint`"] [=map/exists=]:
     1. If |input|["`target_hint`"] is not a [=valid navigable target name or keyword=], then return null.
     1. Set |targetHint| to |input|["`target_hint`"].
-  1. Return a [=speculation rule=] with [=speculation rule/URLs=] |urls|, [=speculation rule/predicate=] |predicate|, [=speculation rule/requirements=] |requirements|, and [=speculation rule/target navigable name hint=] |targetHint|.
+  1. Let |referrerPolicy| be the empty string.
+  1. If |input|["`referrer_policy`"] [=map/exists=]:
+    1. If |input|["`referrer_policy`"] is not a [=referrer policy=], then return null.
```
domenic, line 382: What if it's not a sufficiently-strict referrer policy? Probably that should return null too?

speculation-rules.bs
```diff
@@ -517,9 +547,10 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
   1. [=list/For each=] |prefetchCandidate| of |prefetchCandidates|:
     1. The user agent may run the following steps:
       1. Let |prefetchRecord| be a new [=prefetch record=] whose [=prefetch record/URL=] is |prefetchCandidate|'s [=prefetch candidate/URL=], [=prefetch record/anonymization policy=] is |prefetchCandidate|'s [=prefetch candidate/anonymization policy=], and [=prefetch record/label=] is "`speculation-rules`".
+        <p class="issue">Pass along |prefetchCandidate|'s [=prefetch candidate/referrer policy=].<p>
```
domenic, line 550: I guess we're waiting on #212 before we tackle this?

domenic (commented):
Sorry for the delay in review here!

speculation-rules.bs
domenic: Ah, this is https://github.com/tabatkins/bikeshed/issues/1752.

speculation-rules.bs
kjmcnee: Makes sense, thanks. Updated this to allow the empty string.

speculation-rules.bs
kjmcnee: Since we'll allow other policies for the same site case, I don't think we should enforce that here. Even for the cross-site case, it doesn't really matter if we accept it here, since the rule will still be ignored when we check the computed referrer policy.

speculation-rules.bs
kjmcnee: Yeah, to sketch the remaining plumbing:
- Add a field to prefetch records
- Set the prefetch algorithm's new document state's request referrer policy

The change to the existing prefetch spec would be different, so we may as well wait for that.

speculation-rules.bs
domenic: Great point.

domenic (approved):

speculation-rules.bs
kjmcnee: PR #234 handles this.

PR #217, opened 2022-11-25
Add explainer for relative_to

a4sriniv:
Adds a brief description and example of `"relative_to": "document"` in the document rules section.

### DIFF
--- a/triggers.md
+++ b/triggers.md
@@ -256,7 +256,7 @@ Any of these simple conditions can be negated and combined with conjunction and 
 
 An example of using these would be the following, which marks up as safe-to-prerender all same-origin pages except those known to be problematic:
 
-```js
+```json
 {
   "prerender": [
     {"source": "document",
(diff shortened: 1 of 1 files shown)


triggers.md
```diff
@@ -272,6 +272,12 @@ An example of using these would be the following, which marks up as safe-to-prer
 
 Note how this example uses a low `"score"` value to indicate that, although these links are _safe_ to prerender, they aren't necessarily that important or likely to be clicked on. In such a case, the browser would likely use its own heuristics, e.g. only performing the prerender on pointer-down. Additionally, the web developer might combine this with a higher-scoring rule that indicates which URLs they suspect are likely, which the browser could prerender ahead of time.
 
+For rule sets that are externally fetched, url patterns in `"href_matches"` will be parsed relative to the external resource's url. To parse these patterns relative to the document that fetched the rules instead, `"href_matches"` can be paired with `"relative_to": "document"` like so:
```
jeremyroman, line 275: Instead of "...the document that fetched the rules", I'd maybe say "...the document" or "...the document which contains the link"?

jeremyroman (approved):

triggers.md
a4sriniv: Updated to "the document's base url".

jeremyroman:
It occurs to me that we might want the same thing for list rules, for the same reasons.

```json
{"source": "list", "urls": ["/next"], "relative_to": "document"}
```

WDYT? If this seems plausible maybe we should consider it as an issue.

a4sriniv:
Yeah I agree (and I had a similar thought as well while writing this). I suppose at this point I wonder if it's necessary for `is_relative` to be paired with `href_matches`, and if we could just have `is_relative` as a general speculation rule option? I suppose it doesn't give the same level of granularity, but we aren't giving list rules the same level of granularity either.

jeremyroman:
A list rule can always be split into multiple list rules relatively easily to give any extra granularity needed, much as an `href_matches` predicate can be split into two with `or`. To me that seems more natural than doing it at the ruleset level.

a4sriniv:
Good point, thanks. I've updated the explainer text to now apply `relative_to` to list rules as well, PTAL.

PR #219, opened 2022-11-29
Add "relative_to" to Speculation Rules spec

a4sriniv:


speculation-rules.bs
```diff
@@ -398,7 +401,10 @@ add the following step
     1. If |clause| is null, then return null.
     1. Return a [=document rule negation=] whose [=document rule negation/clause=] is |clause|.
   1. If |predicateType| is "`href_matches`":
-    1. If |input| has any [=map/key=] other than "`href_matches`", then return null.
+    1. If |input| has any [=map/key=] other than "`href_matches`" and "`relative_to`", then return null.
+    1. If |input|["`relative_to`"] [=map/exists=], then:
```
a4sriniv, line 405: The case I'm not super sure about is:

```json
{
   "source": "document",
   "where": {"or": [{"href_matches": "/foo"}, {"href_matches": "/bar", "relative_to": "document"}]},
   "relative_to": "document"
}
```

It feels like the the speculation rule level `relative_to` here was probably accidentally added and could even be a bug, should we fail parsing to guard against this?

jeremyroman: Yeah, it's ambiguous and kinda weird since we haven't yet defined a keyword to change it back. I think my gut is to discard any document rules that use relative_to at the rule level, as invalid. We could always add it in the future if necessary without breaking compat with any valid rules.

a4sriniv: Yeah, that sounds reasonable to me. Updated the spec and and explainer to account for this.

speculation-rules.bs
```diff
@@ -339,21 +339,25 @@ add the following step
...
   1. If |input|["`source`"] does not [=map/exist=] or is neither the [=string=] "`list`" nor the [=string=] "`document`", then return null.
   1. Let |source| by |input|["`source`"].
   1. Let |urls| be an empty [=list=].
   1. Let |predicate| be null.
   1. If |source| is "`list`":
     1. If |input|["`where`"] [=map/exists=], then return null.
     1. If |input|["`urls`"] does not [=map/exist=], is not a [=list=], or has any element which is not a [=string=], then return null.
+    1. If |input|["`relative_to`"] [=map/exists=]:
+      1. If |input|["`relative_to`"] is not "`document`", then return null.
+      1. Set |baseURL| to the [=document base URL=].
```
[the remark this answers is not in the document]
a4sriniv: oops, I conveniently forgot about that, done!

jeremyroman (approved):

speculation-rules.bs
```diff
@@ -317,47 +318,51 @@ add the following step
...
+  1. If |input| has any [=map/key=] other than "`source`", "`urls`", "`where`", "`requires`", `"target_hint"`, and "`relative_to`", then return null.
   1. If |input|["`source`"] does not [=map/exist=] or is neither the [=string=] "`list`" nor the [=string=] "`document`", then return null.
   1. Let |source| by |input|["`source`"].
   1. Let |urls| be an empty [=list=].
   1. Let |predicate| be null.
   1. If |source| is "`list`":
     1. If |input|["`where`"] [=map/exists=], then return null.
     1. If |input|["`urls`"] does not [=map/exist=], is not a [=list=], or has any element which is not a [=string=], then return null.
+    1. If |input|["`relative_to`"] [=map/exists=]:
+      1. If |input|["`relative_to`"] is not "`document`", then return null.
```
domenic, line 352: It might be nice to have a value representing the default, e.g. `"relative_to": "speculation-rules"`. It wouldn't really be useful for anything besides documentation and teaching purposes.

domenic (approved):

speculation-rules.bs
a4sriniv: I added in a default value; I called it "ruleset" instead as we use the ruleset source's base URL (but I don't feel super strongly about this an can use "speculation-rules" if you think that is better).

domenic: LGTM.

I am unsure how I feel about "ruleset" since that term isn't exposed to web developers in any other place; I thought "speculation-rules" (or "speculationrules"??) might work because that's probably the most familiar term for "what goes inside a `<script type="speculationrules">`" and "what's inside the document at the other end of `Speculation-Rules:`".

But, I am happy with either, and this would be easy to change later (probably nobody will depend on this in any significant way). I'll let you and @jeremyroman make the call in your timezone. Please feel free to merge without further review from me :)

jeremyroman: Heh, the reason I didn't include this in my original proposal was exactly because I wasn't sure what to call it.

I think "ruleset" would be my mild preference (because it's relative to this particular ruleset, not all speculation rules globally) but this is only mild; "speculation-rules" totally works too (and @domenic's point is valid).

a4sriniv: Jeremy's rationale is why I didn't like "speculation-rules" as well. We do have the notion of a rule set in the specification, and it's also mention a few times in the explainer so it's not a complete unknown IMO.

PR #222, opened 2022-12-07
Add No-Vary-Search S&P questionnaire

domenic:
Feel free to push to the branch for any fixes and merge before I wake up.

### DIFF
--- a/no-vary-search-security-privacy-questionnaire.md
+++ b/no-vary-search-security-privacy-questionnaire.md
@@ -6,6 +6,8 @@ In theory, this feature can be used to give servers slightly more information ab
 
 In practice, this is not really exposing new information. Such a server can already know all sites that a given user has visited; this just lets requests to one URL (such as `https://example.com/?q=a`) also be used for checking on the status of related URLs (such as `https://example.com/?q=b`), instead of requiring two such requests.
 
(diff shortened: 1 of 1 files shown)


no-vary-search-security-privacy-questionnaire.md
```diff
@@ -0,0 +1,77 @@
+# No-Vary-Search Security & Privacy Questionnaire
+
+> 01.  What information might this feature expose to Web sites or other parties, and for what purposes is that exposure necessary?
+
+None. This feature allows websites to expose information about themselves to the browser, but not the other way around.
```
jeremyroman, line 5: I think it's innocuous, but anything that affects caching does at least provide the opportunity for exposure to some information about the history of previous fetches to that cache (because a request might not be sent in the event of a cache hit, the origin server can make an inference).

I think the differential privacy surface is very small  here, though, and mitigations that apply to caches remain in force with `No-Vary-Search` support as far as I can immediately tell.

jeremyroman (approved):

no-vary-search-security-privacy-questionnaire.md
domenic: Fair; added.

PR #224, opened 2022-12-13
Decode key strings when parsing No-Vary-Search

### LINKED ISSUE #221: No-Vary-Search parsing should make it possible to specify non-ASCII keys
Because the header is parsed as ASCII, we should have a way to specify non-ASCII keys, even though they may not be common.

The most natural way would seen to be applying [percent-decoding](https://url.spec.whatwg.org/#percent-decode) and then [UTF-8 decode without BOM](https://encoding.spec.whatwg.org/#utf-8-decode-without-bom).

I think we should probably also replace + with U+0020 SPACE because the [application/x-www-form-urlencoded parser](https://url.spec.whatwg.org/#concept-urlencoded-parser) and [serializer](https://url.spec.whatwg.org/#concept-urlencoded-serializer) both have this behavior (and the latter actually prefers to emit `+` over `%20` when serializing search params).

domenic:
This allows non-ASCII keys to be specified. Closes #221.

/cc @annevk in case he has any thoughts.

### DIFF
--- a/no-vary-search.bs
+++ b/no-vary-search.bs
@@ -102,14 +102,14 @@ The [=obtain a URL search variance=] algorithm ensures that all [=URL search var
         1. Set |result|'s [=URL search variance/vary params=] to [=URL search variance/no-vary params/wildcard=].
     1. Otherwise, if |value|["`params`"] is a [=list=]:
       1. If any [=list/item=] in |value|["`params`"] is not a [=string=], then return the [=default URL search variance=].
-      1. Set |result|'s [=URL search variance/no-vary params=] to |value|["`params`"].
+      1. Set |result|'s [=URL search variance/no-vary params=] to the result of [=parsing a key=] given each [=list/item=] in |value|["`params`"].
       1. Set |result|'s [=URL search variance/vary params=] to [=URL search variance/no-vary params/wildcard=].
     1. Otherwise, return the [=default URL search variance=].
   1. If |value|["`except`"] [=map/exists=]:
     1. If |value|["`params`"] is not true, then return the [=default URL search variance=].
     1. If |value|["`except`"] is not a [=list=], then return the [=default URL search variance=].
     1. If any [=list/item=] in |value|["`except`"] is not a [=string=], then return the [=default URL search variance=].
-    1. Set |result|'s [=URL search variance/vary params=] to |value|["`except`"].
+    1. Set |result|'s [=URL search variance/vary params=] to the reslt of [=parsing a key=] given each [=list/item=] in |value|["`except`"].
   1. Return |result|.
 
   <p class="note">In general, this algorithm is strict and tends to return the [=default URL search variance=] whenever it sees something it doesn't recognize. This is because the [=default URL search variance=] behavior will just cause fewer cache hits, which is an acceptable fallback behavior.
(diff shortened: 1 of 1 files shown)


no-vary-search.bs
```diff
@@ -251,7 +280,7 @@ Two [=URLs=] |urlA| and |urlB| are <dfn export>equivalent modulo search variance
   <table>
     <thead>
       <tr>
-        <th>Equivalent URLs
+        <th>Equivalent URL strings
```
domenic, line 283: URLs themselves in their parsed form cannot have non-ASCII query components. It's only the parsed query list of (key, value) pairs that contain potentially-non-ASCII strings. But, people can send non-ASCII inputs to the URL parser, which is what I think is valuable to exemplify here and in the newly-added example.

no-vary-search.bs
```diff
@@ -102,14 +102,14 @@ The [=obtain a URL search variance=] algorithm ensures that all [=URL search var
...
-      1. Set |result|'s [=URL search variance/no-vary params=] to |value|["`params`"].
+      1. Set |result|'s [=URL search variance/no-vary params=] to the result of [=parsing a key=] given each [=list/item=] in |value|["`params`"].
       1. Set |result|'s [=URL search variance/vary params=] to [=URL search variance/no-vary params/wildcard=].
     1. Otherwise, return the [=default URL search variance=].
   1. If |value|["`except`"] [=map/exists=]:
     1. If |value|["`params`"] is not true, then return the [=default URL search variance=].
     1. If |value|["`except`"] is not a [=list=], then return the [=default URL search variance=].
     1. If any [=list/item=] in |value|["`except`"] is not a [=string=], then return the [=default URL search variance=].
-    1. Set |result|'s [=URL search variance/vary params=] to |value|["`except`"].
+    1. Set |result|'s [=URL search variance/vary params=] to the reslt of [=parsing a key=] given each [=list/item=] in |value|["`except`"].
```
annevk, line 112:
```suggestion
    1. Set |result|'s [=URL search variance/vary params=] to the result of [=parsing a key=] given each [=list/item=] in |value|["`except`"].
```

no-vary-search.bs
```diff
@@ -189,6 +189,35 @@ The [=obtain a URL search variance=] algorithm ensures that all [=URL search var
...
+
+  1. Let |keyBytesDecoded| be the [=byte sequence/percent-decoding=] of |keyBytes|.
+
+  1. Let |keyStringDecoded| be the [=UTF-8 decode without BOM|UTF-8 decoding without BOM=] of |keyBytesDecoded|.
+
+  1. Return |keyStringDecoded|.
+</div>
+
+<div class="example" id="example-parse-a-key">
+  The [=parse a key=] algorithm allows encoding non-ASCII key strings in the ASCII strucured header format, similar to how the <a><code>application/x-www-form-urlencoded</code></a> format allows encoding an entire entry list of keys and values in ASCII URL format. For example,
```
annevk, line 207:
```suggestion
  The [=parse a key=] algorithm allows encoding non-ASCII key strings in the ASCII structured header format, similar to how the <a><code>application/x-www-form-urlencoded</code></a> format allows encoding an entire entry list of keys and values in ASCII URL format. For example,
```

no-vary-search.bs
```diff
@@ -189,6 +189,35 @@ The [=obtain a URL search variance=] algorithm ensures that all [=URL search var
...
+
+  <pre highlight="http">No-Vary-Search: params=("%C3%A9+%E6%B0%97")</pre>
+
+  will result in a [=URL search variance=] whose [=URL search variance/vary params=] are « "`é 気`" ». As explained in <a href="#example-equivalence-canonicalization">a later example</a>, the canonicalization process during [=equivalent modulo search variance|equivalence testing=] means this will treat as equivalent URL strings such as:
```
annevk, line 211:
```suggestion
  will result in a [=URL search variance=] whose [=URL search variance/vary params=] is « "`é 気`" ». As explained in <a href="#example-equivalence-canonicalization">a later example</a>, the canonicalization process during [=equivalent modulo search variance|equivalence testing=] means this will treat as equivalent URL strings such as:
```

no-vary-search.bs
```diff
@@ -189,6 +189,35 @@ The [=obtain a URL search variance=] algorithm ensures that all [=URL search var
...
+
+  * `https://example.com/?é 気=1`
+  * `https://example.com/?é+気=2`
+  * `https://example.com/?%C3%A9%20気=3`
+  * `https://example.com/?%C3%A9+%E6%B0%97=4`
```
annevk, line 216: Apart from the last one you shouldn't really see the first three at this layer though, right? Those can be input to `<a>`, but by the time we hit the caches and network they are normalized.

no-vary-search.bs
```diff
@@ -102,14 +102,14 @@ The [=obtain a URL search variance=] algorithm ensures that all [=URL search var
         1. Set |result|'s [=URL search variance/vary params=] to [=URL search variance/no-vary params/wildcard=].
     1. Otherwise, if |value|["`params`"] is a [=list=]:
       1. If any [=list/item=] in |value|["`params`"] is not a [=string=], then return the [=default URL search variance=].
-      1. Set |result|'s [=URL search variance/no-vary params=] to |value|["`params`"].
+      1. Set |result|'s [=URL search variance/no-vary params=] to the result of [=parsing a key=] given each [=list/item=] in |value|["`params`"].
```
annevk, line 105: I find this a tad confusing as apparently the result is a list and while "each" gives that away somewhat it's rather magical.

annevk (commented):
Overall this looks correct, but I found some nits.

jeremyroman (approved):

no-vary-search.bs
domenic: "are" is grammatically correct when used after a plural.

no-vary-search.bs
domenic: Yeah, we don't have a good wording for `list.map(operation)`. I'll try

> to the result of applying [parse a key] to each [item] in ...

no-vary-search.bs
domenic: Right, this relates to my below comment, and why I called these "URL strings" instead of URLs. Maybe it's too misleading though and we should overhaul all the examples to only operate on ASCII URL strings...

no-vary-search.bs
domenic: It seems like a plural to me? "params"

PR #226, opened 2022-12-15
Add an overview of an upcoming experiment.

jeremyroman:
I wanted to externalize this overview (so I can link to it from intents etc) and it seems fine to put it here rather than in a Google doc or something harder to find.

### DIFF
--- a/chrome-2023-experiment-overview.md
+++ b/chrome-2023-experiment-overview.md
@@ -1,0 +1,101 @@
+# Chrome 2023 Experiment Overview
+
+In early 2023, Google Chrome will be running an origin trial to allow developers to try out several new enhancements related to speculation rules and prefetching. These enhancements are focused around allowing the browser to automatically assist with prefetching with minimal developer work, and allowing developers to measure when their pages are successfully prefetched.
+
+## Opting into the origin trial
+
+Developers can sign up for an [origin trial](https://developer.chrome.com/docs/web-platform/origin-trials/) token.
+
+> **Note**
+> This trial is not yet available. We'll update this with a link when it launches.
+
+Most developers will only need a token for their own origin, and can use it the usual way.
+
+```http
+Origin-Trial: [token issued to your origin]
+```
+
+```html
+<meta http-equiv="origin-trial" content="[token issued to your origin]">
+```
+
+For this trial, Chrome will also accept a third-party token issued to another origin, as long as either:
+
+* an external script loaded from that origin adds it, or
+* the server delivers the document response with a `Speculation-Rules` HTTP response header which loads speculation rules and an `Origin-Trial` HTTP response header issued to that origin
+
+```http
+Origin-Trial: [third-party token for https://rules-provider.example]
+Speculation-Rules: "https://rules-provider.example/speculationrules.json"
+```
+
+## HTTP-fetched speculation rules
+
+Some developers prefer to deliver speculation rules out of line. You can do this using the `Speculation-Rules` response header.
+
+```http
+Speculation-Rules: "/speculationrules.json"
+```
+
+This will be fetched and cached like any other subresource, though at a lower priority since it doesn't block the current page load. This resource must use the correct MIME type and, if it is cross-origin, pass a [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) check.
+
+```http
+Content-Type: application/speculationrules+json
+Access-Control-Allow-Origin: *
+```
+
+## Automatic link finding
+
+With this enhancement, the browser can automatically react find links in your page which the user is likely to click. The developer simply needs to communicate which links are appropriate to prefetch, since some servers might exhibit side effects for `GET` requests.
+
+You can do this by using [URL patterns](https://developer.mozilla.org/en-US/docs/Web/API/URL_Pattern_API) and simple boolean expressions using "and", "or", and "not". Support for selecting links using CSS selectors is planned. (Let us know if this would make a difference to you!)
+
+```json
+{"prefetch": [
+   {"source": "document",
+    "where": {"or": [
+       {"href_matches": "/articles/*\\?*",
+        "relative_to": "document",
+        "referrer_policy": "strict-origin-when-cross-origin"},
+       {"href_matches": "/products/*\\?*",
+        "relative_to": "document",
+        "referrer_policy": "strict-origin-when-cross-origin"}
+    ]}}
+]}
+```
+
+## When automatic prefetching occurs
+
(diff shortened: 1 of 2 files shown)


chrome-2023-experiment-overview.md
```diff
@@ -0,0 +1,101 @@
+# Chrome 2023 Experiment Overview
```
domenic, line 1: Maybe 2023Q1? We'll probably run more than one experiment this year.

chrome-2023-experiment-overview.md
```diff
@@ -0,0 +1,101 @@
...
+
+In early 2023, Google Chrome will be running an origin trial to allow developers to try out several new enhancements related to speculation rules and prefetching. These enhancements are focused around allowing the browser to automatically assist with prefetching with minimal developer work, and allowing developers to measure when their pages are successfully prefetched.
```
domenic, line 3: I think much of this stuff works for prerendering too, right? Should we use "preloading" here and throughout?

chrome-2023-experiment-overview.md
```diff
@@ -0,0 +1,101 @@
...
+
+* an external script loaded from that origin adds it, or
+* the server delivers the document response with a `Speculation-Rules` HTTP response header which loads speculation rules and an `Origin-Trial` HTTP response header issued to that origin
+
+```http
+Origin-Trial: [third-party token for https://rules-provider.example]
+Speculation-Rules: "https://rules-provider.example/speculationrules.json"
+```
+
+## HTTP-fetched speculation rules
```
domenic, line 32: I'd group all of these under some header with an intro sentence that makes it clear they're the features which your page gets access to when you have added the token.

chrome-2023-experiment-overview.md
```diff
@@ -0,0 +1,101 @@
...
+Access-Control-Allow-Origin: *
+```
+
+## Automatic link finding
+
+With this enhancement, the browser can automatically react find links in your page which the user is likely to click. The developer simply needs to communicate which links are appropriate to prefetch, since some servers might exhibit side effects for `GET` requests.
+
+You can do this by using [URL patterns](https://developer.mozilla.org/en-US/docs/Web/API/URL_Pattern_API) and simple boolean expressions using "and", "or", and "not". Support for selecting links using CSS selectors is planned. (Let us know if this would make a difference to you!)
+
+```json
```
domenic, line 53: I'd add a blocklist example, blocklisting logout. That's probably the most common case.

chrome-2023-experiment-overview.md
```diff
@@ -0,0 +1,101 @@
...
+       {"href_matches": "/articles/*\\?*",
+        "relative_to": "document",
+        "referrer_policy": "strict-origin-when-cross-origin"},
+       {"href_matches": "/products/*\\?*",
+        "relative_to": "document",
+        "referrer_policy": "strict-origin-when-cross-origin"}
+    ]}}
+]}
+```
+
```
domenic, line 66: If this works for prerender too, then I'd add a sentence like:

> This works for prerendering too.

or maybe

> This works for prerendering too, although as usual, making your content [safe to prerender](https://docs.google.com/document/d/1_9XkDUKMGf2f3tDt1gvQQjfliNLpGyFf36BB1-NUZ98/edit#heading=h.312tpmkwsg83) can take extra work.

chrome-2023-experiment-overview.md
```diff
@@ -0,0 +1,101 @@
...
+## When automatic prefetching occurs
+
+In Chrome 110, this experiment will cause Chrome to prefetch up to some small number (currently 5) of URLs when the pointer is down (i.e., mouse down or touch start), which typically occurs between 100 and 200 milliseconds before click, if your speculation rules permit it.
```
domenic, line 69: I find this sentence a bit convoluted. Here's my attempt:

> In Chrome 110, when the pointer is down on an `<a>` element that points to a URL that appears in a `"document"`-source speculation rule, Chrome will start prefetching that link. "Pointer is down" means either mouse down, or touch start, and typically this occurs between 100 and 200 milliseconds before a click.

(we could mention `<area>` too, assuming that's supported? But probably not worthwhile.)

domenic (commented):
Biggest feedback is about whether this should be generalized more to talk about preloading.

chrome-2023-experiment-overview.md
```diff
@@ -0,0 +1,101 @@
...
+This will be fetched and cached like any other subresource, though at a lower priority since it doesn't block the current page load. This resource must use the correct MIME type and, if it is cross-origin, pass a [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) check.
+
+```http
+Content-Type: application/speculationrules+json
+Access-Control-Allow-Origin: *
+```
+
+## Automatic link finding
+
+With this enhancement, the browser can automatically react find links in your page which the user is likely to click. The developer simply needs to communicate which links are appropriate to prefetch, since some servers might exhibit side effects for `GET` requests.
```
tunetheweb, line 49: "react"?

```suggestion
With this enhancement, the browser can automatically prefetch links in your page which the user is likely to click. The developer simply needs to communicate which links are appropriate to prefetch, since some servers might exhibit side effects for `GET` requests.
```

Include prerender if that is included.

chrome-2023-experiment-overview.md
```diff
@@ -0,0 +1,101 @@
...
+
+You can do this by using [URL patterns](https://developer.mozilla.org/en-US/docs/Web/API/URL_Pattern_API) and simple boolean expressions using "and", "or", and "not". Support for selecting links using CSS selectors is planned. (Let us know if this would make a difference to you!)
+
+```json
+{"prefetch": [
+   {"source": "document",
+    "where": {"or": [
+       {"href_matches": "/articles/*\\?*",
```
tunetheweb, line 57: It may be worth commenting on this "weird" URL pattern syntax. Confused me for a long while, until I finally spotted point one in the caveats below. Might be worth linking to that to explain it just after this code block?

jeremyroman: Done. Originally it was closer to this, but the addition of some caveats not in the previous version of this document made that awkward. A link is a good compromise.

chrome-2023-experiment-overview.md
jeremyroman: Added that link above, and added a small parenthetical referring to prerendering. Since the current heuristic is pointerdown, I'd nudge developers toward prefetching for that use case since they're unlikely to realize the differential benefits of prerendering without more lead time.

domenic (approved):

PR #239, opened 2023-01-16
update prerender ClientHints-related algo

Clqsin45:
The "Client Hints Infrastructure" specification updates the logic handling CH Cache[1], so this PR updates the logic on the prerender2 side accordingly to be consistent with the CH specification.

Including:
1. processing a request:
- Before: the processed `hintSet` is the union of `prerender navigable's accept-CH cache` and the global `accept-CH cache`.
- After: `prerender navigable's accept-ch cache` will override the global one.

2.  update `prerender navigable's accept-ch cache`:
- Before:  `prerender navigable's accept-ch cache` was set to the union of global `accept-CH cache` and the tokens specified by the 'Accept-CH' header.
- After: `prerender navigable's accept-ch cache` will be set to the tokens specified by the 'Accept-CH' header, because they are the newest.

[1] https://github.com/WICG/client-hints-infrastructure/pull/135

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -82,13 +82,14 @@ spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
 spec: client-hints-infrastructure; urlPrefix: https://wicg.github.io/client-hints-infrastructure
   type: dfn
     text: Accept-CH cache; url: accept-ch-cache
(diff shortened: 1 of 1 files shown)


domenic (approved):
Great work; thank you! I'll wait to merge this until after #237 lands so we can have a series of green builds on the main branch :)

PR #243, opened 2023-01-24
Speculation rules: add spec text for 'eagerness'.

jeremyroman:


jeremyroman:
This aims to elaborate on [the explainer](https://github.com/WICG/nav-speculation/blob/main/triggers.md#eagerness).

@isaboori ptal as well

speculation-rules.bs
```diff
@@ -345,7 +348,7 @@ add the following step
 <div algorithm="parse a speculation rule">
   To <dfn>parse a speculation rule</dfn> given a [=map=] |input|, a [=document=] |document|, and a [=URL=] |baseURL|, perform the following steps. They return a [=speculation rule=] or null.
 
-  1. If |input| has any [=map/key=] other than "`source`", "`urls`", "`where`", "`requires`", `"target_hint"`,`"referrer_policy"`, and `"relative_to"`, then return null.
+  1. If |input| has any [=map/key=] other than "`source`", "`urls`", "`where`", "`requires`", "`target_hint`","`referrer_policy`", "`relative_to`", and "`eagerness`" then return null.
```
domenic, line 351: (Not actionable but...) I'm starting to wonder if this policy is a mistake. People being unable to use eagerness without browser version sniffing is kind of bad.

domenic (approved):

speculation-rules.bs
jeremyroman: I had the same thought. At the same time, some other properties probably should be mandatory if specified, which is kinda stronger.

Maybe some sort of forward-looking syntax, either for indicating that a key is ignorable if unknown, or an @supports-style feature detection thing? Unfortunately the web's feature detection approach is kinda all over the place so it's not obvious what precedent we would follow if we wanted to.

domenic: I wonder if we could get away with `"requires"` being the only place where unknown causes rejection?

jeremyroman: Maybe so, because we could always cram a `"requires": ["eagerness-support"]` or something like that in the future to reverse it. Would you like to file an issue (or shall I)?

jeremyroman: Filed #244.

PR #245, opened 2023-01-26
speculation-rules spec patch for CSP extension

toyoshim:


toyoshim:
@jeremyroman
cc: @domenic @mkwst

toyoshim:
@jeremyroman 
seems like the preview/build failure was for an infra issue? how can I rerun it?

toyoshim:
cc: @mikewest (sorry, the first mention mistakenly used the different ID)

speculation-rules.bs
```diff
@@ -649,6 +649,27 @@ The same mitigations available to other features also apply here. In particular,
...
+    / "<dfn>'wasm-unsafe-eval'</dfn>" / "<dfn>'inline-speculation-rules'</dfn>"
+</pre>
+
+Also, we make a following change for a note in [[CSP#should-block-inline]].
+
+Note: The valid values for `type` are "`script`", "`script attribute`", "`script speculationrules`", "`style`", and "`style attribute`".
+
+Also we make the following change for [[CSP#directive-script-src]].
+
+  3.  Inline <{script}> blocks MUST pass through [[CSP#should-block-inline]]. Their behavior will be blocked unless every policy allows inline script, either implicitly by not specifying a `script-src` (or `default-src`) directive, or explicitly, by specifying "`unsafe-inline`", a <a grammar>nonce-source</a> or a <a grammar>hash-source</a> that matches the inline block, or specifying "`inline-speculation-rules`" on the type attribute being "`speculationrules`".
```
mikewest, line 671: I think you'd also want to adjust https://w3c.github.io/webappsec-csp/#match-element-to-source-list to actually perform the check you're suggesting here. The text you've edited isn't actually part of any algorithm, but expresses the intention behind the algorithms below. :)

speculation-rules.bs
```diff
@@ -649,6 +649,27 @@ The same mitigations available to other features also apply here. In particular,
 
 The possibility of leaking link URLs via this mechanism is additionally mitigated by the fact that prefetch and prerender to plaintext HTTP (other than to localhost) is not permitted, and so such an on-path attacker could not directly observe preloading request URLs, but would only have access to metadata and traffic analysis. This does not, however, replace standard XSS protections.
 
+<h3 id="security-inline-scripting">Inline scripting</h3>
```
jeremyroman, line 652: I don't think this belongs in the Security Considerations section, which is mostly "here's how we thought about security when we wrote this specification". We could certainly mention this in the section on XSS (Cross-site scripting), but I believe the normative patches to CSP should be in an earlier section. I'd suggest making it a top-level section called something like "CSP Patches", between the Speculation Rules and Security Considerations sections.

speculation-rules.bs
jeremyroman: +1, this will need to patch either that or [allows all inline behavior](https://w3c.github.io/webappsec-csp/#source-list-allows-all-inline-behavior), not sure which.

speculation-rules.bs
```diff
@@ -649,6 +649,27 @@ The same mitigations available to other features also apply here. In particular,
...
+<pre dfn-type="grammar" link-type="grammar">
+  <dfn>keyword-source</dfn> = "<dfn>'self'</dfn>" / "<dfn>'unsafe-inline'</dfn>" / "<dfn>'unsafe-eval'</dfn>"
+    / "<dfn>'strict-dynamic'</dfn>" / "<dfn>'unsafe-hashes'</dfn>"
+    / "<dfn>'report-sample'</dfn>" / "<dfn>'unsafe-allow-redirects'</dfn>"
+    / "<dfn>'wasm-unsafe-eval'</dfn>" / "<dfn>'inline-speculation-rules'</dfn>"
+</pre>
+
+Also, we make a following change for a note in [[CSP#should-block-inline]].
+
+Note: The valid values for `type` are "`script`", "`script attribute`", "`script speculationrules`", "`style`", and "`style attribute`".
```
jeremyroman, line 667: Does anything actually supply the "script speculationrules" value right now? Do we need an additional patch to plumb that through?

jeremyroman (commented):
Re-ran the build automatically for you.

speculation-rules.bs
toyoshim: https://w3c.github.io/webappsec-csp/#ref-for-should-block-inline
There are two references, and probably I should modify this description.
Currently, we don't have a clear statement how each value is chosen. Let me try adding it.

speculation-rules.bs
toyoshim: My last CL contains patches for both sections.

toyoshim:
Hi, @mikewest and @jeremyroman .
Sorry for taking a long time, but now I will focus on making this patch polished to submit.
Can you take another look?

jeremyroman:
> Also, I consistently see the build bot failure around the dfn for user-agent, but I think our document doesn't contain the keyword, and this error cannot be fixed in our repository. Does this error come from other spec repositories' issue?

Upstream change caused the prerendering.bs file's definition of "user agent" to be ambiguous. I've resolved it in the WICG main branch now; if you update your branch to include that, the build should work again.

toyoshim:
thank you for the fix!
now my patch was rebased, and it looks all checks have passed!

speculation-rules.bs
```diff
@@ -627,6 +629,97 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
   1. [=Assert=]: This step is not reached.
 </div>
 
+<h2 id="content-security-policy">Content-Security-Policy</h2>
```
domenic, line 632:
```suggestion
<h2 id="content-security-policy">Content Security Policy</h2>
```

speculation-rules.bs
```diff
@@ -627,6 +629,97 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
...
+<h3 id="content-security-policy-patches">Required Spec Patches</h3>
+Inline script use is also restricted by the [[CSP]], but there is a case a developer doesn't want to permit general inline scripts, but want to permit inline speculation rules. So, we make the following change that extends [[CSP]] to support a new source keyword for inline speculation rules.
```
domenic, line 634:
```suggestion
Inline script use is restricted by [[CSP]], but there are cases where a developer doesn't want to permit general inline scripts, but wants to permit inline speculation rules. So, we make the following change that extends [[CSP]] to support a new source keyword for inline speculation rules.
```

speculation-rules.bs
```diff
@@ -627,6 +629,97 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
...
+
+<h3 id="content-security-policy-patches-directives">Directives</h3>
+In the [[CSP#framework-directives]], we define several associated algorithms, and an <dfn dfn-type="grammar" link-type="grammar">inline check</dfn> algorithm needs a patch to define how to choose the type argument.
```
domenic, line 637: Here and throughout, you can use `<dfn grammar>` instead of `<dfn dfn-type="grammar" link-type="grammar">`.

Also, you should use `noexport` on all definitions that are duplicates of definitions from the CSP spec. So in total, `<dfn grammar noexport>`, or `<dfn for="directive" noexport>`.

speculation-rules.bs
```diff
@@ -627,6 +629,97 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
...
+
+Note: The valid values for <a>`type`</a> are "`script`", "`script attribute`", "`script speculationrules`", "`style`", and "`style attribute`".
+
+<h3 id="content-security-policy-patches-script-src">script-src</h3>
+Also we make the following change for [[CSP#directive-script-src]].
+
+  3.  Inline <{script}> blocks MUST pass through [[CSP#should-block-inline]]. Their behavior will be blocked unless every policy allows inline script, either implicitly by not specifying a `script-src` (or `default-src`) directive, or explicitly, by specifying "`unsafe-inline`", a <a grammar>nonce-source</a> or a <a grammar>hash-source</a> that matches the inline block, or specifying "`inline-speculation-rules`" on the type attribute being "`speculationrules`".
+
+<h3 id="content-security-policy-patches-allow-all-inline">Does a source list allow all inline behavior for type?</h3>
+The algorighm needs a patch to handle the "`script speculationrules`" type.
```
domenic, line 673:
```suggestion
The algorithm needs a patch to handle the "`script speculationrules`" type.
```

speculation-rules.bs
```diff
@@ -627,6 +629,97 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
...
+          <a grammar>`hash-source`</a> grammar, return "`Does Not Allow`".
+
+      2.  If <a>`type`</a> is "`script`", "`script attribute`" or "`navigation`"
+          and |expression| matches the <a grammar>keyword-source</a>
+          "<a grammar>`'strict-dynamic'`</a>", return "`Does Not Allow`".
+
+          Note: `'strict-dynamic'` only applies to scripts, not other resource
+          types. Usage is explained in more detail in [[CSP#strict-dynamic-usage]].
+
+      3.  If <a>`type`</a> is "`script speculationrules`" and |expression| matches the
```
domenic, line 687: For inserted things inside multi-step algorithms, it would be good to mark them up with `<ins>` / `</ins>` to make it clearer to the reader.

domenic (commented):
Normative content looks solid

speculation-rules.bs
```diff
@@ -627,6 +629,96 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
...
+      if the {{Element}} is {{HTMLScriptElement}} and the {{HTMLScriptElement/type}}
+      attribute is "`speculationrules`", or set to
+      "<dfn dfn-type="grammar" link-type="grammer">`script attribute`</dfn>" for
+      JavaScript event handlers. Otherwise, it is set to 
+      "<dfn dfn-type="grammar" link-type="grammer">`script`</dfn>".
+
+<h3 id="content-security-policy-patches-source-lists">Source Lists</h3>
+In the [[CSP#framework-directive-source-list]], we change the following definition to have a new source, `inline-speculation-rules`.
+
+<pre dfn-type="grammar" link-type="grammar">
```
toyoshim, line 654: Does this also need an extra attribute not to export dfn tags used inside the pre tag?
or should I also add noexport for each dfn tag?

speculation-rules.bs
toyoshim: Resolved

speculation-rules.bs
domenic: I think you can put the `noexport` on this container.

domenic (approved):

speculation-rules.bs
toyoshim: Done!

toyoshim:
@jeremyroman @mikewest @domenic 
now all comments are addressed. can you take another look?

speculation-rules.bs
```diff
@@ -627,6 +629,96 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
   1. [=Assert=]: This step is not reached.
 </div>
 
+<h2 id="content-security-policy">Content Security Policy</h2>
+Inline script use is restricted by [[CSP]], but there are cases where a developer doesn't want to permit general inline scripts, but wants to permit inline speculation rules. So, we make the following change that extends [[CSP]] to support a new source keyword for inline speculation rules.
+
+<h3 id="content-security-policy-patches-directives">Directives</h3>
+In the [[CSP#framework-directives]], we define several associated algorithms, and an <dfn grammar noexport>inline check</dfn> algorithm needs a patch to define how to choose the type argument.
```
jeremyroman, line 636: nit

```suggestion
In [[CSP#framework-directives]], we define several associated algorithms, and an <dfn grammar noexport>inline check</dfn> algorithm needs a patch to define how to choose the type argument.
```

Similar below.

speculation-rules.bs
```diff
@@ -627,6 +629,96 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
...
+      <ins>The <a>`type`</a> string is set to
+      "<dfn grammar noexport>`script speculationrules`</dfn>"
+      if the {{Element}} is {{HTMLScriptElement}} and the {{HTMLScriptElement/type}}
+      attribute is "`speculationrules`", or set to
+      "<dfn grammar noexport>`script attribute`</dfn>" for
+      JavaScript event handlers. Otherwise, it is set to 
+      "<dfn grammar noexport>`script`</dfn>".</ins>
+
+<h3 id="content-security-policy-patches-source-lists">Source Lists</h3>
+In the [[CSP#framework-directive-source-list]], we change the following definition to have a new source, `inline-speculation-rules`.
```
jeremyroman, line 652:
```suggestion
In [[CSP#framework-directive-source-list]], we change the following definition to have a new source, `inline-speculation-rules`.
```

speculation-rules.bs
```diff
@@ -627,6 +629,96 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
...
+
+<h3 id="content-security-policy-patches-directives">Directives</h3>
+In the [[CSP#framework-directives]], we define several associated algorithms, and an <dfn grammar noexport>inline check</dfn> algorithm needs a patch to define how to choose the type argument.
+
+  3.  An <dfn for="directive" noexport>inline check</dfn>, which takes an {{Element}}, a
+      <dfn>`type`</dfn> string, a policy, and a source string as arguments,
+      and is executed during [[CSP#should-block-inline]] and during
+      [[CSP#should-block-navigation-request]] for `javascript:` requests. This
+      algorithm returns "`Allowed`" unless otherwise specified.
+      <ins>The <a>`type`</a> string is set to
```
jeremyroman, line 643: This kinda makes an assertion about the calling algorithm, which isn't true without further patches. Supplying "`script attribute`" happens in the code for [event handler content attributes](https://html.spec.whatwg.org/multipage/webappapis.html#event-handler-content-attributes), but nothing in this spec yet invokes the inline check algorithm with the "`script speculationrules`" type.

Presumably this requires a change to the "prepare the script element" patch to update step 19 (which invokes "Should element's inline behavior be blocked by Content Security Security Policy?") to supply "`script speculationrules`" instead of "`script`" in appropriate cases.

speculation-rules.bs
jeremyroman: Love this btw. I'd totally forgetten `<ins>` even existed.

speculation-rules.bs
toyoshim: I think the `event handler content attributes` is the spec for inline javascripts in event handlers, such as &lt;button onlick="run();"&gt;. So it would not be a speculationrules. That's why I skipped to patch it for handling "`script speculationrules`" type.

toyoshim: Also, the spec you pointed as the step 19 is the one in the algorithm below?
https://html.spec.whatwg.org/multipage/scripting.html#prepare-the-script-element

So, I agreed that I need to add breakdown steps that setup the `type` and call the algorithm with it to return when it returns "Blocked".

I will update my PR to have a patch to the scripting spec.

toyoshim:
As the main branch could not pass the build test due to the possible multiple references again, I sent another PR to fix the build.
https://github.com/WICG/nav-speculation/pull/250
@domenic @jeremyroman Can you review and merge it to solve the build issue?

speculation-rules.bs
toyoshim: Ah, OK. Our spec already have a patch for the "prepare the script element".
So, what I should do is to modify the existing patch to be aligned.

toyoshim: Now the latest patch contains the mod for the existing patch section.

toyoshim:
thanks! the multiple reference issues were solved.
the latest patch resolves all comments, and rebased on the top of the last upstream fix.

speculation-rules.bs
```diff
@@ -239,6 +241,14 @@ Inside the [=prepare the script element=] algorithm we make the following change
     </dd>
   </dl>
 
+* Replace "script" argument with |type| that is set through the following steps, and passed to [[CSP#should-block-inline]] algorithm.
```
jeremyroman, line 244:
```suggestion
* Replace "script" argument with |cspType| that is set through the following steps, and passed to [[CSP#should-block-inline]] algorithm.
```

speculation-rules.bs
```diff
@@ -239,6 +241,14 @@ Inside the [=prepare the script element=] algorithm we make the following change
     </dd>
   </dl>
 
+* Replace "script" argument with |cspType| that is set through the following steps, and passed to [[CSP#should-block-inline]] algorithm.
+  <dl>
+    <dd>
+      1. Let |cspType| be "script".
+      1. If |type| is "`speculationrules`", then set |cspType| to "`script speculationrules`".
```
jeremyroman, line 248:
```suggestion
      1. If <var ignore>type</var> is "`speculationrules`", then set |cspType| to "`script speculationrules`".
```

This is probably the fix to the build issue.

toyoshim:
thanks!
ping @mikewest if you have another feedback

speculation-rules.bs
```diff
@@ -629,6 +639,97 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
   1. [=Assert=]: This step is not reached.
 </div>
 
+<h2 id="content-security-policy">Content Security Policy</h2>
+Inline script use is restricted by [[CSP]], but there are cases where a developer doesn't want to permit general inline scripts, but wants to permit inline speculation rules. So, we make the following change that extends [[CSP]] to support a new source keyword for inline speculation rules.
+Also, this changes affect
```
mikewest, line 644: Nit: This sentence trails off...

mikewest (approved):
One nit, otherwise LGTM. Thanks for taking another pass!

speculation-rules.bs
toyoshim: oh, thanks. this line was mistakenly added in the rev d98e...

toyoshim:
thanks everyone!
@jeremyroman may I ask you to merge this PR?

PR #256, opened 2023-03-22
Use hyperlink url instead of href

### LINKED ISSUE #255: Recent HTTPS-skipping commit is not correct
https://github.com/WICG/nav-speculation/commit/18a76e0601d028f01e5c1ac07e4d362d0c36ba70

|href| is a string, not a URL record, so it does not have a scheme.

You should instead look at the element's [url](https://html.spec.whatwg.org/#concept-hyperlink-url).

@a4sriniv

a4sriniv:
Fixes #255.

### DIFF
--- a/speculation-rules.bs
+++ b/speculation-rules.bs
@@ -528,7 +528,7 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
 
         <div class="note">This corresponds to the elements which match '':any-link'' [=pseudo-class=], or which appear in the {{Document/links}} collection.</div>
 
-    1. Let |url| be |descendant|'s associated [=url=].
+    1. Let |url| be |descendant|'s [=url=].
     1. If |url| is null, or its [=url/scheme=] is not an [=HTTP(S) scheme=], [=iteration/continue=].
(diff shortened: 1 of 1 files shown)


jeremyroman:
While here, can you fix the other places that use the href getter but should use url? (Not sure if this changed since I wrote that, or if it was always wrong.)

Also, if you merge with the main branch, you should pick up a spec build fix I merged earlier today.

jeremyroman (approved):

speculation-rules.bs
```diff
@@ -51,6 +51,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: get an element's target; url: get-an-element's-target
     urlPrefix: links.html
       text: following a hyperlink; url: following-hyperlinks-2
+      text: url; url: concept-hyperlink-url
```
domenic, line 54: It doesn't really make a difference, but it might be nicer to do `for: HTMLHyperlinkElementUtils` and then change the call sites to `[=HTMLHyperlinkElementUtils/url=]`.

domenic (approved):
Nice, thank you!!

speculation-rules.bs
a4sriniv: Thanks, done!

PR #257, opened 2023-03-22
Fix the spec build.

jeremyroman:
(1) Apparently bikeshed now requires that `<pre>` be the first thing on
    the line. Not sure why.

(2) It is now necessary to backslash-escape the CSP 'foo' references
    because otherwise they are interpreted as CSS autolinks (since this
    file does have CSS markup shorthand enabled).

### DIFF
--- a/no-vary-search.bs
+++ b/no-vary-search.bs
@@ -125,17 +125,20 @@ The [=obtain a URL search variance=] algorithm ensures that all [=URL search var
         <th>Result</th>
     <tbody>
       <tr>
-        <td><pre highlight="http">No-Vary-Search: params</pre>
+        <td>
+          <pre highlight="http">No-Vary-Search: params</pre>
         <td>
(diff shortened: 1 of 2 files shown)


domenic:
cc @TabAtkins as more evidence of breakage from the new parser. The `<pre>` thing is especially bad.

tabatkins:
For (1), the reasoning behind the warning is that the markdown-block parser in Bikeshed skips the contents of pre elements *if the line starts with the tag*, but if there's non-whitespace preceding the tag it doesn't, so you can end up getting the contents accidentally mangled as markdown. Firing on single-line `pre`s was indeed overzealous.

(2) was a simple mistake that should have been caught; I discuss in [my post-mortem](https://github.com/speced/bikeshed/issues/1773#issuecomment-1481598821) why this was missed.

But yeah the changes are all reverted for now so this is moot.

PR #262, opened 2023-03-30
Adjust redirect following behavior.

jeremyroman:
This removes the barrier between same- and cross-partition prefetches, instead requiring that for each request (including redirects), we isolate storage for anything that would belong outside the current partition.

I should probably write more about the security and privacy considerations of this, but this is hopefully ready enough to be an improvement already.

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -192,24 +192,16 @@ A <dfn export>prefetch record</dfn> is a [=struct=] with the following [=struct/
 
   <div class="note">"`canceled`" indicates that the prefetch was aborted by the author or user, or terminated by the user agent.</div>
 * <dfn export for="prefetch record">fetch controller</dfn>, a [=fetch controller=] (a new [=fetch controller=] by default)
-* <dfn for="prefetch record">partition state</dfn>, which is a [=same-partition prefetch state=], a [=cross-partition prefetch state=], or null (the default)
 * <dfn export for="prefetch record">sandboxing flag set</dfn>, a [=sandboxing flag set=]
 * <dfn export for="prefetch record">redirect chain</dfn>, a [=redirect chain=] (empty by default)
 * <dfn export for="prefetch record">expiry time</dfn>, a {{DOMHighResTimeStamp}} (0.0 by default)
+* <dfn export for="prefetch record">source partition key</dfn>, a [=network partition key=] or null (the default)
+* <dfn export for="prefetch record">isolated partition key</dfn>, a [=network partition key=] whose first item is an [=opaque origin=] and which represents a separate partition in which cross-partition state can be temporarily stored, or null (the default)
+* <dfn export for="prefetch record">origins with conflicting credentials</dfn>, an [=ordered set=] of [=origins=] (initially empty)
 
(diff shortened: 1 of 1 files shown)


prefetch.bs
```diff
@@ -637,23 +624,14 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
...
     1. Let |coopEnforcementResult| be the result of [=creating a cross-origin opener policy enforcement result for navigation=] given |document| and |document|'s [=Document/origin=].
     1. Let |global| be |document|'s [=relevant global object=].
     1. [=In parallel=]:
         1. Let |navigationParams| be the result of [=creating navigation params by fetching=] given |request|, |entry|, |coopEnforcementResult|, |document|'s [=node navigable=], |sourceSnapshotParams|, |targetSnapshotParams|, "`other`", null (navigationId), "`navigate`", and <a href="#create-navigation-params-by-fetching-prefetchRecord"><i>prefetchRecord</i></a> |prefetchRecord|.
         1. If |navigationParams|'s [=navigation params/response=] does not [=support prefetch=], then set |navigationParams| to null.
-        1. If |prefetchRecord|'s [=prefetch record/partition state=] is a [=cross-partition prefetch state=] whose [=cross-partition prefetch state/origins with conflicting credentials=] is not [=set/empty=], then set |navigationParams| to null.
+        1. If |prefetchRecord|'s [=prefetch record/origins with conflicting credentials=] is not [=set/empty=], then set |navigationParams| to null.
```
domenic, line 633: Does this mean we still follow the redirects, even after we know the end result will be failure? Is that a good idea?

It seems like instead of tracking the list of origins with conflicting credentials, which we only ever do a size check on (right?), we could just bail out with a network error instead of appending to the list.

jeremyroman: It could be a boolean; it's been a list just because it's about as long and seemed to me a little clearer about what's going on, and UAs might want to report it anyway. But that might be overkill, I'm not attached to that.

As far as bailing out with a network error, the problem is that it would reveal whether the user has credentials to an origin that shouldn't know. Consider:

a.example wants to know if the user has credentials on b.example, and there is a URL on b.example which redirects to a.example (unconditionally).

Then a page on a.example can request a prefetch of https://b.example/redirect-to/a and observe whether or not it sees a request for a.example. This should not be possible, because we're not in the b.example partition, until such time as the user actually navigates.

The private prefetching proxy also sends these requests that we know (at least, before Supports-Loading-Mode: uncredentialed-prefetch) will not be used, so that this information is not leaked at prefetch time.

domenic: This makes perfect sense. We should document that in a note, probably at the point at which we add to the set, but maybe with another small note here.

I think a boolean is probably a bit cleaner, personally. Otherwise I as the reader am wondering if I missed something. Or I guess we could add some text to make use of the origins, like "User agents may [=report a warning to the console=] describing how |prefetchRecord|'s [=prefetch record/origins with conflicting credentials=] prevented the prefetch from being used."

kjmcnee:
> btw @kjmcnee you've looked at the referrer policy stuff before, so if you have a moment to skim that and make sure it makes sense that would be great (if not, no worries)

Yeah, the referrer policy part looks correct.

prefetch.bs
jeremyroman: Switched to a boolean.

domenic (approved):

PR #263, opened 2023-04-04
Address redirect privacy in more detail

jeremyroman:
The privacy considerations section here merits some expansion now. This change adds some detail and links to it from the prefetch spec doc as well.


jeremyroman:
I'm not suuuper happy with the clarity of my explanation here. I've kinda struggled to express it both clearly and concisely. Suggestions welcome.

speculation-rules.bs
```diff
@@ -805,7 +805,16 @@ While efforts have been made to minimize the privacy impact of prefetching, some
...
+
+A notable exception to this policy is that in order to determine if using an empty partition affected the resource returned by a cross-partition request, one of two conditions have to be met:
```
domenic, line 810: This long sentence is the most unclear thing here. I'm not clear how one of these conditions being met constitutes an exception to the policy stated above.

This _might_ be a correct rephrasing?

> Using this third empty partition will only give correct results if one of the following conditions is met:
> 
> 1. the response declares that the resource fetched with no credentials can be used, even if a normal (non-prefetch) navigation would have been done with credentials; or
> 2. the destination partition contains no credentials.
> 
> We can only determine if the latter condition is met by examining the destination partition's state during the prefetch process. To avoid this...

jeremyroman: I'm trying to call extra attention to the fact that otherwise we can simply say "we never examine cross-partition state", but in this case we do (cautiously).

domenic (approved):

Issue #266, opened 2023-04-07
Make prefetch records not Document-scoped


hiroshige-g:
Currently prefetch records are scoped by its Documents, but we might want to make the scope broader.

Possible benefits would be:

- Prefetched results are more broadly available
- Chromium implementation might be simpler (not so sure though)

Candidate scoping would be:

- Per-[network partition keys](https://fetch.spec.whatwg.org/#network-partition-key) -- perhaps better aligning with HTTP caches etc.
- Per-User agent (Per-profile in Chrome)

Filing this issue here to track this idea and discussion.
cc/ @domenic @nyaxt @jeremyroman

nyaxt:
> Per-[network partition keys](https://fetch.spec.whatwg.org/#network-partition-key) -- perhaps better aligning with HTTP caches etc.

I'm currently exploring this option.
We have a write up on the topic - https://docs.google.com/document/d/1z3BTb3jEfyvUJIFJCc7e9UKB5dIZlGRHFhpU8R_s66M/edit

We are currently trying out an Chromium implementation under a feature flag to see how it looks like.

PR #267, opened 2023-04-10
Use sourceDocument if set for prefetch lookup

hiroshige-g:
- Spec issue: https://github.com/WICG/nav-speculation/issues/261
- WPT change: https://github.com/web-platform-tests/wpt/pull/39518

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -413,7 +413,7 @@ Update all creation sites to supply an empty string, except for any in this docu
 <hr>
 
 <div algorithm="attempt to populate the history entry's document">
-    In <a spec=HTML>attempt to populate the history entry's document</a>, add an optional {{Document}} |sourceDocument| argument, plumb it from <a spec=HTML>navigate</a>, and replace the step which invokes <a spec=HTML>create navigation params by fetching</a> with the following:
+    In <a spec=HTML>attempt to populate the history entry's document</a>, add an optional {{Document}} |sourceDocument| argument, and replace the step which invokes <a spec=HTML>create navigation params by fetching</a> with the following:
 
     1. Otherwise, if both of the following are true:
         * <var ignore>entry</var>'s [=session history entry/URL=]'s [=url/scheme=] is a [=fetch scheme=]; and
@@ -423,8 +423,9 @@ Update all creation sites to supply an empty string, except for any in this docu
 
             1. Let |request| be the result of [=creating a navigation request=] given <var ignore>entry</var>, <var ignore>sourceSnapshotParams</var>'s [=source snapshot params/fetch client=], <var ignore>navigable</var>'s [=navigable/container=], and <var ignore>sourceSnapshotParams</var>'s [=source snapshot params/has transient activation=].
             1. Set |request|'s [=request/replaces client id=] to <var ignore>navigable</var>'s [=navigable/active document=]'s [=relevant settings object=]'s [=environment/id=].
-            1. If <var ignore>navigable</var>'s [=navigable/loading mode=] is "`prerender`", then set |referringDocument| to |sourceDocument|.
+            1. If |sourceDocument| is given, then set |referringDocument| to |sourceDocument|.
             1. Otherwise, set |referringDocument| to <var ignore>navigable</var>'s [=navigable/active document=].
+               <p class="issue">Or should we fallback to null instead? A notable case where this shouldn't be null is <a spec=HTML>reload</a>, but I'm not sure about other cases.</p>
             1. Let |prefetchRecord| be the result of [=finding a matching prefetch record=] given |referringDocument|, <var ignore>entry</var>'s [=session history entry/URL=], and <var ignore>targetSnapshotParams</var>'s [=target snapshot params/sandboxing flags=].
             1. If <var ignore>documentResource</var> is null and |prefetchRecord| is not null:
                 1. Set <var ignore>navigationParams</var> to the result of [=creating navigation params from a prefetch record=] given <var ignore>navigable</var>, <var ignore>entry</var>'s [=session history entry/document state=], <var ignore>navigationId</var>, <var ignore>navTimingType</var>, <var ignore>request</var>, |prefetchRecord|, <var ignore>targetSnapshotParams</var>, and <var ignore>sourceSnapshotParams</var>.
@@ -435,6 +436,8 @@ Update all creation sites to supply an empty string, except for any in this docu
                 1. Let |coopEnforcementResult| be the result of [=creating a cross-origin opener policy enforcement result for navigation=] given <var ignore>navigable</var>'s [=navigable/active document=] and <var ignore>entry</var>'s [=session history entry/document state=]'s [=document state/initiator origin=].
                 1. Set <var ignore>navigationParams</var> to the result of [=creating navigation params by fetching=] given |request|, <var ignore>entry</var>, |coopEnforcementResult|, <var ignore>navigable</var>, <var ignore>sourceSnapshotParams</var>, <var ignore>targetSnapshotParams</var>, <var ignore>cspNavigationType</var>, <var ignore>navigationId</var>, and <var ignore>navTimingType</var>.
 
+    The |sourceDocument| argument should be plumbed from <a spec=HTML>navigate</a>'s |sourceDocument| and <a spec=HTML>traverse the history by a delta</a>'s |sourceDocument|.
+    <p class="issue">How about other callers, i.e. <a spec=HTML>apply pending history changes</a>, <a spec=HTML>finalize a same-document navigation</a> and <a spec=HTML>finalize a cross-document navigation</a>?</p>
 </div>
 
 <hr>
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -121,6 +121,7 @@ spec: COOKIES; urlPrefix: https://httpwg.org/specs/rfc6265.html
 spec: nav-speculation; urlPrefix: prerendering.html
   type: dfn
     text: getting the supported loading modes; url: get-the-supported-loading-modes
+    text: loading mode; for: navigable; url: navigable-loading-mode
     text: uncredentialed-prefetch; for: Supports-Loading-Mode; url: supports-loading-mode-uncredentialed-prefetch
 spec: resource-timing; urlPrefix: https://w3c.github.io/resource-timing/
   type: dfn; for: PerformanceResourceTiming; text: delivery type; url: dfn-delivery-type
@@ -412,7 +413,7 @@ Update all creation sites to supply an empty string, except for any in this docu
 <hr>
 
 <div algorithm="attempt to populate the history entry's document">
-    In <a spec=HTML>attempt to populate the history entry's document</a>, replace the step which invokes <a spec=HTML>create navigation params by fetching</a> with the following:
+    In <a spec=HTML>attempt to populate the history entry's document</a>, add an optional {{Document}} |sourceDocument| argument, plumb it from <a spec=HTML>navigate</a>, and replace the step which invokes <a spec=HTML>create navigation params by fetching</a> with the following:
 
     1. Otherwise, if both of the following are true:
         * <var ignore>entry</var>'s [=session history entry/URL=]'s [=url/scheme=] is a [=fetch scheme=]; and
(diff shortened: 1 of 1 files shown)


domenic (approved):
This looks good, but what do you think about unconditionally using _sourceDocument_, instead of only using it for prerendering?

For example, I think that would be necessary to make `<a href="prefetched.html" target="_blank">` work, which is desirable.

hiroshige-g:
> unconditionally using sourceDocument

Perhaps this makes sense.
See https://chromium-review.googlesource.com/c/chromium/src/+/4422475 for behavior changes.

Yesterday I and @nyaxt discussed not to make this generalization because this would dissappear after https://github.com/WICG/nav-speculation/issues/266, but the updates after that are:

- The issue for `<a href="prefetched.html" target="_blank">` remains even after https://github.com/WICG/nav-speculation/issues/266 (at least for cross-site URLs) because the navigable's active document anyway doesn't belong to the initiator's site.
- I found this is implementable, as the `sourceDocument` seems to correspond to `NavigationRequest::initiator_document_`. See https://chromium-review.googlesource.com/c/chromium/src/+/4372403/.

prefetch.bs
```diff
@@ -412,7 +413,7 @@ Update all creation sites to supply an empty string, except for any in this docu
 <hr>
 
 <div algorithm="attempt to populate the history entry's document">
-    In <a spec=HTML>attempt to populate the history entry's document</a>, replace the step which invokes <a spec=HTML>create navigation params by fetching</a> with the following:
+    In <a spec=HTML>attempt to populate the history entry's document</a>, add an optional {{Document}} |sourceDocument| argument, and replace the step which invokes <a spec=HTML>create navigation params by fetching</a> with the following:
```
domenic, line 416: Since it is only used for looking up prefetches, I think we should name it something like _sourceDocumentForPrefetches_ or _prefetchSourceDocument_ or similar.

Alternately, maybe we should only plumb the prefetch record through. Then it is more obvious what the purpose of all this is. The downside would be duplicating a couple steps in each call site, hmm.

hiroshige-g:
Uploaded a commit that uses sourceDocument if set, but as also I added as inline issue comments in the commit, I'm not so sure how to plumb `sourceDocument` in some cases -- maybe we just don't plumb `sourceDocument` and fallback to navigable's active document instead?

prefetch.bs
```diff
@@ -432,6 +436,8 @@ Update all creation sites to supply an empty string, except for any in this docu
                 1. Let |coopEnforcementResult| be the result of [=creating a cross-origin opener policy enforcement result for navigation=] given <var ignore>navigable</var>'s [=navigable/active document=] and <var ignore>entry</var>'s [=session history entry/document state=]'s [=document state/initiator origin=].
                 1. Set <var ignore>navigationParams</var> to the result of [=creating navigation params by fetching=] given |request|, <var ignore>entry</var>, |coopEnforcementResult|, <var ignore>navigable</var>, <var ignore>sourceSnapshotParams</var>, <var ignore>targetSnapshotParams</var>, <var ignore>cspNavigationType</var>, <var ignore>navigationId</var>, and <var ignore>navTimingType</var>.
 
+    The |sourceDocument| argument should be plumbed from <a spec=HTML>navigate</a>'s |sourceDocument| and <a spec=HTML>traverse the history by a delta</a>'s |sourceDocument|.
+    <p class="issue">How about other callers, i.e. <a spec=HTML>apply pending history changes</a>, <a spec=HTML>finalize a same-document navigation</a> and <a spec=HTML>finalize a cross-document navigation</a>?</p>
```
domenic, line 440: - apply pending history changes: only if we want to support prefetching reloads. Which I'm not sure we do?
- finalize a same-document navigation: no, prefetching is not relevant there
- finalize a cross-document navigation: yes, plumbing it through here is necessary for the path navigate -> finalize a cross-document navigation -> apply the history step.

prefetch.bs
```diff
@@ -121,6 +121,7 @@ spec: COOKIES; urlPrefix: https://httpwg.org/specs/rfc6265.html
 spec: nav-speculation; urlPrefix: prerendering.html
   type: dfn
     text: getting the supported loading modes; url: get-the-supported-loading-modes
+    text: loading mode; for: navigable; url: navigable-loading-mode
```
domenic, line 124: Is this necessary?

domenic (commented):
Overall strategy looks good, just some ideas on how to make it cleaner.

prefetch.bs
hiroshige-g: > support prefetching reloads

There is a WPT already for reload (https://github.com/WICG/nav-speculation/pull/212#discussion_r1035155217), but maybe just to check the Chromium behavior?
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/speculation-rules/prefetch/prefetch-traverse-reload.sub.html

> navigate -> finalize a cross-document navigation -> apply the history step

Do you have examples/test cases where prefetched results would be used in this code path? (I'm not familiar with how "finalize a cross-document navigation" works)

prefetch.bs
hiroshige-g: No longer needed. Removed.

prefetch.bs
hiroshige-g: Sounds good > sourceDocumentForPrefetches.

> only plumb the prefetch record ... The downside would be duplicating a couple steps in each call site

+1 > downside -- also https://github.com/WICG/nav-speculation/issues/266 would affect how prefetch records are obtained, so plumbing a Document to https://wicg.github.io/nav-speculation/prefetch.html#find-a-matching-prefetch-record and getting prefetch records there (either from Document / network partition / etc.) is simpler.

hiroshige-g: Done > renaming to _sourceDocumentForPrefetches_.

prefetch.bs
domenic: > There is a WPT already for reload ([#212 (comment)](https://github.com/WICG/nav-speculation/pull/212#discussion_r1035155217)), but maybe just to check the Chromium behavior?

Ah OK, I am happy to keep it then.

> Do you have examples/test cases where prefetched results would be used in this code path? (I'm not familiar with how "finalize a cross-document navigation" works)

This is the most basic code path. E.g. this is the code path which clicking a link or doing `location.href` goes through.

hiroshige-g: https://html.spec.whatwg.org/multipage/browsing-the-web.html#reload says:

> It is intentional that the resulting call to [apply the history step](https://html.spec.whatwg.org/multipage/browsing-the-web.html#apply-the-history-step) does not pass [sourceSnapshotParams](https://html.spec.whatwg.org/multipage/browsing-the-web.html#apply-history-step-source-snapshot) or [initiatorToCheck](https://html.spec.whatwg.org/multipage/browsing-the-web.html#apply-history-step-initiator).

So probably we shouldn't pass `sourceDocumentForPrefetches` as well?

Also, how about passing `sourceDocumentForPrefetches` (or its prefetch records) via `sourceSnapshotParams`?

Pros: It might be consistent with the existing decisions whether we should pass sourceDocument-related informatio n (`sourceSnapshotParams` or `sourceDocumentForPrefetches`) or not

Cons?: this means not passing `sourceDocumentForPrefetches` in `finalize a cross-document navigation`. IIUC `sourceDocumentForPrefetches` will be still passed for ordinal cross-document navigations via navigate -> [attempt to populate the history entry's document](https://html.spec.whatwg.org/multipage/browsing-the-web.html#attempt-to-populate-the-history-entry's-document) -> [creating navigation params by fetching](https://html.spec.whatwg.org/multipage/browsing-the-web.html#create-navigation-params-by-fetching), even if we don't pass `sourceDocumentForPrefetches` via navigate -> `finalize a cross-document navigation` though (I'm not sure the differences between the two paths, but probably the path via [attempt to populate the history entry's document](https://html.spec.whatwg.org/multipage/browsing-the-web.html#attempt-to-populate-the-history-entry's-document) is the primary one?).

domenic: Sorry for the delay; I got sick :(.

> So probably we shouldn't pass `sourceDocumentForPrefetches` as well?

Hmm, I'm not sure. Mostly, sourceSnapshotParams and initiatorToCheck are used for security checks which (for some reason) browsers decide to not perform on reloads. The only thing that's kind of similar to how sourceDocumentForPrefetches is used is sourceSnapshotParams's fetch client, which is used to determine referrer and such. (E.g. for reloads, it always looks like the page is reloading itself, according to the `Referer` header.)

> Cons?: this means not passing `sourceDocumentForPrefetches` in `finalize a cross-document navigation`. IIUC `sourceDocumentForPrefetches` will be still passed for ordinal cross-document navigations via navigate -> [attempt to populate the history entry's document](https://html.spec.whatwg.org/multipage/browsing-the-web.html#attempt-to-populate-the-history-entry's-document) -> [creating navigation params by fetching](https://html.spec.whatwg.org/multipage/browsing-the-web.html#create-navigation-params-by-fetching), even if we don't pass `sourceDocumentForPrefetches` via navigate -> `finalize a cross-document navigation` though (I'm not sure the differences between the two paths, but probably the path via [attempt to populate the history entry's document](https://html.spec.whatwg.org/multipage/browsing-the-web.html#attempt-to-populate-the-history-entry's-document) is the primary one?).

Sorry, I realize I was misleading in what I said earlier. The paths are:

- push/replace navigation: `navigate` -> `attempt to populate` -> `finalize a cross-document navigation` -> `apply the history step`. In `apply the history step`, targetEntry's document will be present, because of the earlier `attempt to populate` step.

- reload navigation: `reload` -> `apply pending history changes` -> `apply the history step` -> `attempt to populate`. This time we reach `attempt to populate` from `apply the history step`, because targetEntry's document state's reload pending is true 

- back/forward navigation, bfcache: `traverse the history by a delta` -> `apply the history step`. In `apply the history step`, targetEntry's document will be present, because bfcache.

- back/forward navigation, non-bfcache: `traverse the history by a delta` -> `apply the history step` -> `attempt to populate`. This time we reach `attempt to populate` from `apply the history step`, because targetEntry's document is null, because no bfcache.

Does that make sense? Feel free to grab me in person or in realtime if it doesn't; I know this stuff is complicated and I don't want to be confusing about it again.

So, I think your idea of putting sourceDocumentForPrefetches or the prefetch records in source snapshot params mostly works, and feels elegant. The consequence would be that reloads can only use prefetch records from their own document though, e.g. if you had this setup:

- Parent loads `a.html` in an iframe
- Later, parent prefetches `a.html`
- Parent calls `frames[0].reload()`

then the prefetch would not be used. Maybe that's fine, because it's congruent with how the fetch client is set to `frames[0]` instead of to the parent?

Another elegant part of putting the prefetch records (not sourceDocumentForPrefetches) in source snapshot params is that if we start storing the prefetch records on a wider scope, per https://github.com/WICG/nav-speculation/issues/266, it will not change very much: it will just change the snapshotting call to grab them from a different place.

hiroshige-g: Hmm, thank you for inputs and clarifications! Then I'll draft the spec to put prefetch records to source snapshot params.

I'm also less sure about reload's behavior here, but I feel we can prioritize (at least for now) the elegance/consistency of the spec architecture by putting the prefetch records to source snapshot params, and then later (after #266) we can revisit the reload behavior if actually needed in the wild.

If we make prefetch records scoped by network partition keys, I imagine prefetch records would be partitioned by the result of `determine the network partition key given request’s client` (because `sourceDocumentForPrefetches` basically corresponds to source snapshot params' client), while HTTP caches for navigation requests are partitioned by the result of `determine the network partition key given request’s reserved client`, which is perhaps simple.

hiroshige-g:
@ArthurSonzogni FYI.
`sourceSnapshotParams` fallbacks to `navigable's active document`'s snapshot at Step 12.6.4 of https://html.spec.whatwg.org/multipage/browsing-the-web.html#apply-the-history-step, which might be similar fallback you mentioned in the codereview comment https://chromium-review.googlesource.com/c/chromium/src/+/4372403/comment/bc534a9a_458b492c/.

prefetch.bs
```diff
@@ -411,6 +410,14 @@ Update all creation sites to supply an empty string, except for any in this docu
 
 <hr>
 
+Add an additional item to [=source snapshot params=] as follows:
+:  <dfn for="source snapshot params">prefetch records</dfn>
+:: a a [=list=] of [=prefetch records=].
+
+When to [=snapshot source snapshot params=], set [=source snapshot params/prefetch records=] to <var ignore>sourceDocument</var>'s [=Document/prefetch records=].
```
domenic, line 417: Probably make a [=list/clone=] here (https://infra.spec.whatwg.org/#list-clone) to reflect the idea we're snapshotting.

hiroshige-g: I'm not so sure: in general creating a snapshot would be a good idea to avoid unintentional race conditions, while in this case not taking a clone might be also an option,

- As [=list/clone=] is a shallow copy, this isn't a complete frozen snapshot of the prefetch records.
- When a prefetch is still ongoing at the time of navigation start and we want to use the prefetch result after response is received after navigation start, we might need intentional interactions via (non-snapshot) prefetch records (IIUC  such case is not yet specified so though)

domenic: Yeah, I can see your concerns. However, I think it'd most fit the spirit of source snapshot params to start with a shallow clone for now, reflecting the idea that any prefetches added after the prerender navigation starts, should not influence the prerender navigation. Similar to how, e.g., changes to the page's origin or sandboxing flags after navigation starts, should not affect the (normal, non-prerender) navigation.

I agree that it's a little strange that we're not doing a deep clone. But that's kind of the nature of prefetch records being things that are updated over time. I think that part is OK.

PR #268, opened 2023-04-19
Allow inline-speculation-rules use with nonce, hash, and strict-dynamic

toyoshim:
"Does a source list allow all inline behavior for type?" defines an algorithm, and it is designed to ignore 'unsafe-inline' and 'inline-speculation-rules' when one of nonce, hash, or strict-dynamic presents.

For 'inline-speculation-rules', this is not beneficial. So, this patch updates the spec patch not to ignore 'inline-speculation-rules' in the algorithm.

### DIFF
--- a/speculation-rules.bs
+++ b/speculation-rules.bs
@@ -684,25 +684,32 @@ Also we make the following change for [[CSP#directive-script-src]].
 <h3 id="content-security-policy-patches-allow-all-inline">Does a source list allow all inline behavior for type?</h3>
 The algorithm needs a patch to handle the "`script speculationrules`" type.
 
-  2.  <a for=set>For each</a> |expression| of list:
+  1.  Let |allow all inline| be `false`.
+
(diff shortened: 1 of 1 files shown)


toyoshim:
This is a feedback from https://bugs.chromium.org/p/chromium/issues/detail?id=1433616
@jeremyroman can you take a look?
@mikewest can you also take a look from the CSP perspective?

jeremyroman:
Since I missed this at the time and I think there's been further discussion since, can I just check whether this is still current? If so I can take a look.

domenic:
We got pushback from @mikewest internally on this, suggesting that instead we use workarounds to accomplish the same goals. I don't agree with that pushback, as I think it makes the platform less comprehensible, but I think we're currently stuck here.

toyoshim:
Thanks @domenic to share the context here.
I changed the status to Draft as we would not go without resolving CSP people's concerns, and I could not spend enough time on this for now.

PR #272, opened 2023-05-16
Add explainer for expects_no_vary_search

kjmcnee:
The explainer currently describes a proposed extension for a No-Vary-Search hint. We now make this part of the current proposal and describe the details of the solution.


domenic (approved):
Great stuff, including the new paragraph about whether exact matches are required.

jeremyroman (commented):
This is now a weirdly late review comment because GitHub was serving me 500 errors yesterday, but nonetheless...

PR #274, opened 2023-06-02
Append allow-2119 class to the note

toyoshim:
As bikeshed starts annoying against RFC2119 keyword in non-normative section.


domenic:
We should fix the note instead. See some suggestions in the last paragraph of https://infra.spec.whatwg.org/#conformance

toyoshim:
As the PR's title isn't suitable now, I will close this and make another PR instead.

Issue #165, opened 2022-06-17
Web audio is not restricted while prerendering


domenic:
My read of the current spec at https://wicg.github.io/nav-speculation/prerendering.html is that we have gated audio/video elements, and web speech API, but not any part of web audio.

domenic:
I'm not sure what all the web audio entry points that might play sound are. One might be the `AudioContext` constructor, which currently has a vague, implementation-defined [allowed to start](https://webaudio.github.io/web-audio-api/#allowed-to-start). We could say that prerendering pages are never allowed to start.

toyoshim:
AudioContext can be constructed synchronously, but it doesn't start audio playback if it is constructed without user gesture. In such a case, AudioContext::resume() is needed to start playing. So, we can make the method deferred while prerendering? My assumption is user gestures could not be made while prerendering.

toyoshim:
fyi, @hoch

domenic:
I think that approach sounds reasonable (and indeed, user gestures cannot be made while prerendering). I'm a bit unsure how it fits into the Web Audio spec. https://webaudio.github.io/web-audio-api/ is very hard to read (lots of insertions and deletions for some reason) but it only mentions activation once: https://webaudio.github.io/web-audio-api/#allowed-to-start . It sounds like it's implementation-defined?

Maybe we can just modify https://webaudio.github.io/web-audio-api/#allowed-to-start to say you're never "allowed to start" during prerendering?

toyoshim:
Ah, I see. The user gesture requirement was added later. So, the spec just says it is kind of implementation specific for now even though all browsers are changed to require a user gesture now.
I agreed that our spec patch would say `you're never "allowed to start" during prerendering`.

As far as I know, an explicit call of resume() is needed, and the Web Audio best practice is
 - Construct AudioContext with a user gesture
 - If code was already there and change is difficult, insert AudioContext::resume() at the first user gesture point
 
 resume() may have two choices.
1) reject with InvalidStateError (maybe no patch is needed if we don't allow us to start during prerendering at the AudioContext construction time)
2) [DelayWhilePrerendering]

From the implementation perspective, it just works without any change if we take 1) ?

toyoshim:
Looked at our implementation.
It seems we define AutoplayPolicy, and it can be AutoplayPolicy::kNoUserGestureRequired.
With this case, the AudioContext always starts playback, and users may not expect resume() call. So if prerender disallow it, the page may be broken?
[...]

PR #275, opened 2023-06-02
Spec patch for the Web Audio API restriction

toyoshim:
This is based on the discussion at #165

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -96,18 +96,11 @@ spec: client-hints-infrastructure; urlPrefix: https://wicg.github.io/client-hint
 spec: web-audio; urlPrefix: https://webaudio.github.io/web-audio-api/
   type: dfn
     text: allowed to start; url: allowed-to-start
-    text: AudioContext; url: AudioContext
-    text: AudioContext constructor; url: AudioContext-constructors
     text: control message; url: control-message
     text: [[control thread state]]; url: dom-baseaudiocontext-control-thread-state-slot
     text: running; url: dom-audiocontextstate-running
     text: queue a control message; url: queuing
-    text: resume(); url: dom-audiocontext-resume
     text: [[pending resume promises]]; url: dom-audiocontext-pending-resume-promises-slot
-    text: AudioScheduledSourceNode; url: AudioScheduledSourceNode
-    text: start(when); url: dom-audioscheduledsourcenode-start
-    text: AudioBufferSourceNode; url: AudioBufferSourceNode
-    text: start(when, offset, duration); url: dom-audiobuffersourcenode-start
 
 </pre>
 <pre class="biblio">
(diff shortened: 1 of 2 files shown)


prerendering.bs
```diff
@@ -787,6 +794,20 @@ Add {{[DelayWhilePrerendering]}} to {{Navigator/getUserMedia()}}, {{MediaDevices
   1. If [=this=]'s [=relevant global object=]'s [=Window/navigable=] is a [=prerendering navigable=], then return.
 </div>
 
+<h4 id="web-audio-patch">Web Audio API</h4>
+
+The concept, [=allowed to start=], is used in the spec, but details are not defined as implementation details.
+To restrict auto playback while prerendering, add the following note and steps in the [=AudioContext=] interface section.
+
+<p class="note">
```
domenic, line 802: `class="note"` is for informative notes, not normative requirements. So just use a normal paragraph and say "the following requirements" instead.

prerendering.bs
```diff
@@ -787,6 +794,20 @@ Add {{[DelayWhilePrerendering]}} to {{Navigator/getUserMedia()}}, {{MediaDevices
...
+The [=AudioContext=] is never allowed to start while prerendering.
```
domenic, line 803:
```suggestion
The [=AudioContext=] is never [=allowed to start=] while prerendering.
```

domenic (commented):
Overall, this looks pretty reasonable to me. I left a few editorial comments.

I am a little scared of "If [=allowed to start=] algorithm runs and returns false only due to the prerendering". It is very general.

The alternative would be to only patch the AudioContext constructor. But, I am not sure if that is right either... the other three call sites (in `resume()` and `start(when)` and `start(when, offset, duration)`) might also be important. Do they need the same behavior, of setting the control thread state and queueing a control message to resume? Hmm.

So, I am OK with the text as-is, but would be interested in your thoughts. (Or thoughts from @hoch, if possible.)

toyoshim:
`allow to start` needs to return `false` for all callers during prerendering, but the post-prerendering activation patch is probably needed only in the AudioContext constuctor case.

Users may or may not call other resume or start methods, but their promises would not be resolved until the post-prerendering activation steps run. When our step runs, the resume control message is processed, and it resolves all pending resume promises.

So, it seems to work without  compatibility breakages. I will revise my PR in this direction.

prerendering.bs
toyoshim: done.

prerendering.bs
toyoshim: done.

hoch:
As you already saw, the construction of AudioContext, Autoplay control, and the resume behavior is rather messy.

I believe the last comment from @toyoshim looks correct.

prerendering.bs
```diff
@@ -787,6 +802,30 @@ Add {{[DelayWhilePrerendering]}} to {{Navigator/getUserMedia()}}, {{MediaDevices
...
+<h4 id="web-audio-patch">Web Audio API</h4>
+
+The concept, [=allowed to start=], is used in the spec, but details are not defined as implementation details.
+To restrict auto playback while prerendering, add the following rule in the [=AudioContext=] interface section.
+
+The [=AudioContext=] is never [=allowed to start=] while prerendering.
+
+Also modify [=AudioContext constructor=] steps to have the following step before returning the constructed object.
+
+1. Else if the context is not [=allow to start=] only due to the prerendering, then append the following steps to [=this=]'s [=Document/post-prerendering activation steps list=].
```
hoch, line 814: What does it mean by "Else if" here?

prerendering.bs
```diff
@@ -787,6 +795,30 @@ Add {{[DelayWhilePrerendering]}} to {{Navigator/getUserMedia()}}, {{MediaDevices
...
+  1. Set the [=[[control thread state]]=] on the {{AudioContext}} to [=running=].
+  1. [=Queue a control message=] to resume the {{AudioContext}}
+
+<p class="note">
+Users might call {{AudioContext/resume()}} that ensurs the context is not [=allow to start=] and appends a {{Promise}} to [=[[pending resume promises]]=].
+The {{Promise}} will be resolved in the activation steps above.
+</p>
+
+<p class="note">
+The {{AudioScheduledSourceNode}} interface has {{AudioScheduledSourceNode/start(when)}} method and the {{AudioBufferSourceNode}} interface has {{AudioBufferSourceNode/start(when, offset, duration)}} method that might affect [=[[control thread state]]=].
```
domenic, line 817:
```suggestion
The {{AudioScheduledSourceNode}} interface has a {{AudioScheduledSourceNode/start(when)}} method and the {{AudioBufferSourceNode}} interface has a {{AudioBufferSourceNode/start(when, offset, duration)}} method that might affect [=[[control thread state]]=].
```

prerendering.bs
toyoshim: Yes.

toyoshim:
Thanks!

prerendering.bs
toyoshim: @hoch Are you ok with the current PR, or do you think we need some editing here to clarify the patch's context a little more?

hoch: The current PR looks good to me. Thanks for asking!

hoch (approved):
LGTM

PR #278, opened 2023-06-14
Add spec for No-Vary-Search hint in speculation rules

kjmcnee:
We introduce the "expects_no_vary_search" field for speculation rules and specify how it is parsed.

When matching prefetch records, we now have it:
* perform inexact matches based on the response's No-Vary-Search value
* block on the availability of headers for potential matches
* have the identification of potential matches incorporate the No-Vary-Search hint

Note that this spec text describes blocking on any available prefetch. The current chromium implementation blocks only on a single ongoing prefetch.

### DIFF
--- a/no-vary-search.bs
+++ b/no-vary-search.bs
@@ -22,6 +22,7 @@ spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
       text: dictionary; url: name-dictionaries
       text: boolean; url: name-boolean
       text: inner list; url: name-inner-lists
+    text: parsing structured fields; url: #text-parse
 </pre>
 <style>
 #example-equivalence-canonicalization table {
@@ -82,15 +83,14 @@ The [=obtain a URL search variance=] algorithm ensures that all [=URL search var
 <h2 id="parsing">Parsing</h2>
 
 <div algorithm>
-  To <dfn>obtain a URL search variance</dfn> given a [=response=] |response|:
+  To <dfn>parse a URL search variance</dfn> given a [=map=] |value|:
 
-  1. Let |value| be the result of [=header list/getting a structured field value=] given [:No-Vary-Search:] and "`dictionary`" from |response|'s [=response/header list=].
-  1. If |value| is null, then return the [=default URL search variance=].
-  1. If |value|'s [=map/keys=] [=list/contains=] anything other than "`key-order`", "`params`", or "`except`", then return the [=default URL search variance=].
+  1. If |value| is null, then return null.
+  1. If |value|'s [=map/keys=] [=list/contains=] anything other than "`key-order`", "`params`", or "`except`", then return null.
   1. Let |result| be a new [=URL search variance=].
   1. Set |result|'s [=URL search variance/vary on key order=] to true.
   1. If |value|["`key-order`"] [=map/exists=]:
-    1. If |value|["`key-order`"] is not a [=boolean=], then return the [=default URL search variance=].
+    1. If |value|["`key-order`"] is not a [=boolean=], then return null.
     1. Set |result|'s [=URL search variance/vary on key order=] to the boolean negation of |value|["`key-order`"].
   1. If |value|["`params`"] [=map/exists=]:
     1. If |value|["`params`"] is a [=boolean=]:
@@ -101,16 +101,26 @@ The [=obtain a URL search variance=] algorithm ensures that all [=URL search var
         1. Set |result|'s [=URL search variance/no-vary params=] to the empty list.
         1. Set |result|'s [=URL search variance/vary params=] to [=URL search variance/no-vary params/wildcard=].
     1. Otherwise, if |value|["`params`"] is a [=list=]:
-      1. If any [=list/item=] in |value|["`params`"] is not a [=string=], then return the [=default URL search variance=].
+      1. If any [=list/item=] in |value|["`params`"] is not a [=string=], then return null.
       1. Set |result|'s [=URL search variance/no-vary params=] to the result of applying [=parse a key=] to each [=list/item=] in |value|["`params`"].
       1. Set |result|'s [=URL search variance/vary params=] to [=URL search variance/no-vary params/wildcard=].
-    1. Otherwise, return the [=default URL search variance=].
+    1. Otherwise, return null.
   1. If |value|["`except`"] [=map/exists=]:
-    1. If |value|["`params`"] is not true, then return the [=default URL search variance=].
-    1. If |value|["`except`"] is not a [=list=], then return the [=default URL search variance=].
-    1. If any [=list/item=] in |value|["`except`"] is not a [=string=], then return the [=default URL search variance=].
+    1. If |value|["`params`"] is not true, then return null.
+    1. If |value|["`except`"] is not a [=list=], then return null.
+    1. If any [=list/item=] in |value|["`except`"] is not a [=string=], then return null.
     1. Set |result|'s [=URL search variance/vary params=] to the result of applying [=parse a key=] to each [=list/item=] in |value|["`except`"].
   1. Return |result|.
+
+</div>
+
+<div algorithm>
+  To <dfn>obtain a URL search variance</dfn> given a [=response=] |response|:
+
+  1. Let |fieldValue| be the result of [=header list/getting a structured field value=] given [:No-Vary-Search:] and "`dictionary`" from |response|'s [=response/header list=].
+  1. Let |result| be the result of [=parsing a URL search variance=] given |fieldValue|.
+  1. If |result| is null, then return the [=default URL search variance=].
+  1. Otherwise return |result|.
 
   <p class="note">In general, this algorithm is strict and tends to return the [=default URL search variance=] whenever it sees something it doesn't recognize. This is because the [=default URL search variance=] behavior will just cause fewer cache hits, which is an acceptable fallback behavior.
 </div>
@@ -187,6 +197,18 @@ The [=obtain a URL search variance=] algorithm ensures that all [=URL search var
         <td><pre highlight="http">No-Vary-Search: key-order=?0</pre>
         <td>(omit the header)
   </table>
+</div>
+
+<div algorithm>
+  To <dfn>obtain a URL search variance hint</dfn> given a [=string=] |hintValue|:
+
+  1. Let |fieldValue| be the result of [=parsing structured fields=] given |hintValue| and "`dictionary`".
+  1. If parsing failed, then return null.
+  1. Let |result| be the result of [=parsing a URL search variance=] given |fieldValue|.
+  1. If |result| is null, then return null.
+  1. Otherwise return |result|.
+
+  <p class="note">Speculation rule parsing is strict, so the return value of this algorithm surfaces parsing errors instead of silently returning the [=default URL search variance=].
 </div>
 
 <div algorithm>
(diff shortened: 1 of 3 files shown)


speculation-rules.bs
```diff
@@ -586,7 +604,7 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
   1. End the [=synchronous section=], continuing the remaining steps [=in parallel=].
   1. [=list/For each=] |prefetchCandidate| of |prefetchCandidates|:
     1. The user agent may run the following steps:
```
kjmcnee, line 606: The existing spec text seems general enough to cover the incorporation of NVS hint into user agent heuristics (for prefetching a non-eager candidate when a predicted URL inexactly matches the candidate). But perhaps a non-normative note near the eagerness description on line 616 would be useful?

prefetch.bs
```diff
@@ -194,6 +200,7 @@ A <dfn export>prefetch record</dfn> is a [=struct=] with the following [=struct/
 * <dfn export for="prefetch record">fetch controller</dfn>, a [=fetch controller=] (a new [=fetch controller=] by default)
 * <dfn export for="prefetch record">sandboxing flag set</dfn>, a [=sandboxing flag set=]
 * <dfn export for="prefetch record">redirect chain</dfn>, a [=redirect chain=] (empty by default)
+* <dfn export for="prefetch record">creation time</dfn>, a {{DOMHighResTimeStamp}} (0.0 by default)
```
jeremyroman, line 203: Is this the time that the prefetch starts? If so, is "start time" more direct?

prefetch.bs
```diff
@@ -225,34 +250,66 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
...
+        1. Let |completeRecord| be the result of [=finding a matching complete prefetch record=] given |document|, |url|, and |sandboxFlags|.
+        1. If |completeRecord| is not null, return |completeRecord|.
+        1. Let |potentialRecords| be an empty [=list=].
+        1. [=list/For each=] |record| of |document|'s [=Document/prefetch records=]:
+            1. If all of the following are true, then [=list/append=] |record| to |potentialRecords|:
+                * |record|'s [=prefetch record/state=] is "`ongoing`".
+                * |record| [=prefetch record/possibly matches a URL=] given |url|.
+                * |record|'s [=prefetch record/sandboxing flag set=] is not empty or |sandboxFlags| is empty.
+                * |record|'s [=prefetch record/expiry time=] is greater than the [=current high resolution time=] for the [=relevant global object=] of |document|.
+                * |record|'s [=prefetch record/redirect chain=] does not [=redirect chain/has updated credentials|have updated credentials=].
```
jeremyroman, line 304: Can we actually change how much time we wait based on this? If the updated credentials are cross-site, at least, then this seems like it might reveal cross-site browsing activity (albeit only activity occurring during the prefetch window).

prefetch.bs
```diff
@@ -225,34 +250,66 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
...
+                * |cutoffTime| is null or |record|'s [=prefetch record/creation time=] is less than |cutoffTime|.
+        1. If |potentialRecords| [=list/is empty=], return null.
+        1. Wait until the [=prefetch record/state=] of any element of |document|'s [=Document/prefetch records=] changes.
+        1. If |cutoffTime| is null and any element of |potentialRecords| has a [=prefetch record/state=] that is not "`ongoing`", set |cutoffTime| to the [=current high resolution time=] for the [=relevant global object=] of |document|.
```
jeremyroman, line 308: Shouldn't this be set before the wait, possibly outside the loop? I assume the goal here is "don't block on prefetches which started after we first started looking"? The current semantics are, at least, confusing to me.

prefetch.bs
```diff
@@ -225,34 +250,66 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
...
+        1. If |recordToUse|'s [=prefetch record/redirect chain=] [=redirect chain/has updated credentials=], return null.
+        1. Return |recordToUse|.
     1. Return null.
 
     <p class="note">It's not obvious, but this doesn't actually require that the prefetch have received the complete body, just the response headers. In particular, a navigation to a prefetched response might nonetheless not load instantaneously.</p>
 
     <p class="issue">It might be possible to use cache response headers to determine when a response can be used multiple times, but given the short lifetime of the prefetch buffer it's unclear whether this is worthwhile.</p>
 </div>
 
+<div algorithm="find a matching prefetch record">
```
jeremyroman, line 289: You can omit the value of the algorithm attr, and bikeshed will infer it from the <dfn>.

prefetch.bs
```diff
@@ -225,34 +250,66 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
...
+    To <dfn export>find a matching prefetch record</dfn> given a {{Document}} |document|, [=URL=] |url|, and [=sandboxing flag set=] |sandboxFlags|, perform the following steps.
```
jeremyroman, line 290: Should we say "wait for a matching prefetch record" or similar, to make it clearer at the call site that this isn't always immediate?

prefetch.bs
```diff
@@ -225,34 +250,66 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
...
+        1. [=list/For each=] |record| of |document|'s [=Document/prefetch records=]:
+            1. If all of the following are true, then [=list/append=] |record| to |potentialRecords|:
+                * |record|'s [=prefetch record/state=] is "`ongoing`".
+                * |record| [=prefetch record/possibly matches a URL=] given |url|.
+                * |record|'s [=prefetch record/sandboxing flag set=] is not empty or |sandboxFlags| is empty.
+                * |record|'s [=prefetch record/expiry time=] is greater than the [=current high resolution time=] for the [=relevant global object=] of |document|.
+                * |record|'s [=prefetch record/redirect chain=] does not [=redirect chain/has updated credentials|have updated credentials=].
+                * |cutoffTime| is null or |record|'s [=prefetch record/creation time=] is less than |cutoffTime|.
+        1. If |potentialRecords| [=list/is empty=], return null.
+        1. Wait until the [=prefetch record/state=] of any element of |document|'s [=Document/prefetch records=] changes.
```
jeremyroman, line 307: Is this algorithm invoked in parallel, so that this doesn't block things?

no-vary-search.bs
```diff
@@ -189,6 +199,18 @@ The [=obtain a URL search variance=] algorithm ensures that all [=URL search var
...
+<div algorithm>
+  To <dfn>obtain a URL search variance hint</dfn> given a [=string=] |hintValue|:
+
+  1. Let |fieldValue| be the result of [=parsing structured fields=] given |hintValue| and "`dictionary`".
+  1. If parsing failed, then return null.
+  1. Let |result| be the result of [=parsing a URL search variance=] given |fieldValue|.
+  1. If |result| is null, then return null.
+  1. Otherwise return |result|.
+
+  <p class="note">Speculation rule parsing is strict, so the return value of this algorithm surfaces parsing errors instead of silently returning the [=default URL search variance=].
```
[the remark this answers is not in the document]
kjmcnee: Ah, I think I see your point. Also the motivation for the strictness described in "parse speculation rules" talks about misunderstood rules leading to more speculation than was intended. But here, we know the consequence of an invalid hint and it would not cause more speculation.

For debuggability, we should presumably still have the implementation warn about this.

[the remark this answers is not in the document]
kjmcnee: Tagging @liviutinta in case of any concerns.

prefetch.bs
kjmcnee: The note below this explains the motivation. The goal is to allow prefetches started after we started blocking, which is why the cutoff isn't set until this point. Would it help to move the note next to this line?

prefetch.bs
kjmcnee: Yes. See the assert at the beginning of the algorithm. Would it help to move the assert here?

no-vary-search.bs
```diff
@@ -189,6 +199,18 @@ The [=obtain a URL search variance=] algorithm ensures that all [=URL search var
...
 </div>
 
+<div algorithm>
+  To <dfn>obtain a URL search variance hint</dfn> given a [=string=] |hintValue|:
+
+  1. Let |fieldValue| be the result of [=parsing structured fields=] given |hintValue| and "`dictionary`".
+  1. If parsing failed, then return null.
+  1. Let |result| be the result of [=parsing a URL search variance=] given |fieldValue|.
+  1. If |result| is null, then return null.
+  1. Otherwise return |result|.
```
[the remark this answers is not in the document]
kjmcnee: Done.

no-vary-search.bs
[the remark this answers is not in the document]
kjmcnee: I've updated the parsing of the hint to fallback to the default variance. Note that in "parse a speculation rule" this still treats non-string values for expects_no_vary_search as invalid.

prefetch.bs
kjmcnee: Done.
(Yes. Creation and starting happen at the same time.)

prefetch.bs
```diff
@@ -207,6 +214,24 @@ A [=prefetch record=]'s <dfn export for="prefetch record">response</dfn> is the
...
+    A [=prefetch record=] |prefetchRecord| <dfn export for="prefetch record">matches a URL</dfn> given a [=URL=] |url| if the following algorithm returns true:
+    1. If |prefetchRecord|'s [=prefetch record/URL=] is equal to |url|, return true.
+    1. If |prefetchRecord|'s [=prefetch record/response=] is not null:
+      1. Let |searchVariance| be the result of [=obtaining a URL search variance=] given |prefetchRecord|'s [=prefetch record/response=].
+      1. If |prefetchRecord|'s [=prefetch record/URL=] and |url| are [=equivalent modulo search variance=] given |searchVariance|, return true.
+    1. Otherwise, return false.
+</div>
+
+<div algorithm>
+    A [=prefetch record=] |prefetchRecord| <dfn export for="prefetch record">possibly matches a URL</dfn> given a [=URL=] |url| if the following algorithm returns true:
```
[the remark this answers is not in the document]
kjmcnee: Done.

prefetch.bs
kjmcnee: Done.

prefetch.bs
kjmcnee: Done.

prefetch.bs
kjmcnee: Done.

prefetch.bs
jeremyroman: Ah, okay, missed that.

prefetch.bs
jeremyroman: Discussed a little aloud. I see the informative note below, but I'd appreciate it if it could be expanded to be explicit about the intuitive understanding of this heuristic (blocking only on prefetch requests which started before a prefetch completed), why we're using it (we want to allow a slightly later prefetch which is a better match, but don't want a fixed timeout that might cause issues with slow servers), and what its properties are (e.g., in the worst case, we block on two nearly-consecutive prefetches and then the actual navigation if we still failed to match).

Parentheticals are merely inspiration from my understanding and you can write a more cohesive note. :)

speculation-rules.bs
jeremyroman: +1 to a non-normative note.

jeremyroman (approved):

prefetch.bs
kjmcnee: Done.

speculation-rules.bs
kjmcnee: Done.

PR #279, opened 2023-08-09
Restrict updated credentials checks to cross-partition.

jeremyroman:
These are the cases where conflicting credentials at prefetch time would have prevented the prefetch in the first place. Other kinds of credential updates (within the network partition) are left for authors to handle, though tools for facilitating that could be added later.

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -669,7 +669,7 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
     1. [=In parallel=]:
         1. Let |navigationParams| be the result of [=creating navigation params by fetching=] given |request|, |entry|, |coopEnforcementResult|, |document|'s [=node navigable=], |sourceSnapshotParams|, |targetSnapshotParams|, "`other`", null (navigationId), "`navigate`", and <a href="#create-navigation-params-by-fetching-prefetchRecord"><i>prefetchRecord</i></a> |prefetchRecord|.
         1. If |navigationParams|'s [=navigation params/response=] does not [=support prefetch=], then set |navigationParams| to null.
-        1. If |prefetchRecord| [=prefetch record/had conflicting credentials=], then set |navigationParams| to null.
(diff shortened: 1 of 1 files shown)


jeremyroman:
This better matches what Chromium does, contrary to what I previously thought.

prefetch.bs
```diff
@@ -689,7 +669,7 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
     1. [=In parallel=]:
         1. Let |navigationParams| be the result of [=creating navigation params by fetching=] given |request|, |entry|, |coopEnforcementResult|, |document|'s [=node navigable=], |sourceSnapshotParams|, |targetSnapshotParams|, "`other`", null (navigationId), "`navigate`", and <a href="#create-navigation-params-by-fetching-prefetchRecord"><i>prefetchRecord</i></a> |prefetchRecord|.
         1. If |navigationParams|'s [=navigation params/response=] does not [=support prefetch=], then set |navigationParams| to null.
-        1. If |prefetchRecord| [=prefetch record/has conflicting credentials=], then set |navigationParams| to null.
+        1. If |prefetchRecord| [=prefetch record/had conflicting credentials=], then set |navigationParams| to null.
```
domenic, line 672:
```suggestion
        1. If |prefetchRecord|'s [=prefetch record/had conflicting credentials=] is true, then set |navigationParams| to null.
```

domenic (approved):
Huh, I didn't realize all of this spec text was used only for the same-partition case.

I guess the big difference is that for cross-partition cases, we don't want to bother comparing cookie data; we just want to bail if there are any cookies at all?

jeremyroman:
> Huh, I didn't realize all of this spec text was used only for the same-partition case.
> 
> I guess the big difference is that for cross-partition cases, we don't want to bother comparing cookie data; we just want to bail if there are any cookies at all?

Yeah, if we only care about whether there are credentials at all, that's much more concise to check. And if the purpose of this is only to check that the uncredentialed prefetch assumption still holds, then that's all that is required, I think.

PR #281, opened 2023-11-01
Add the 'immediate' eagerness level.

jeremyroman:


jeremyroman:
As you proposed, @domenic. PTAL?

domenic (approved):
Spec changes LGTM, but we should also update triggers.md. (Maybe just replacing some of its contents with a link to the spec, to avoid duplication?)

PR #283, opened 2023-11-02
Delay registerProtocolHandler() until activation

nhiroki:
This delays registerProtocolHandler() and unregisterProtocolHandler() until activation. To handle SyntaxError DOMException and SecurityError DOMException during normalizing parameters, this adds custom steps instead of [DelayWhilePrerendering].

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -865,6 +865,24 @@ Add {{[DelayWhilePrerendering]}} to {{SpeechRecognition/start()}}, {{SpeechRecog
 
 Add {{[DelayWhilePrerendering]}} to {{LockManager/request(name, callback)}}, {{LockManager/request(name, options, callback)}}, and {{LockManager/query()}}.
 
+<h4 id="custom-scheme-handlers-patch">Custom Scheme Handlers</h4>
+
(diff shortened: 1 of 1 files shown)


nhiroki:
Regarding the IPR failure, I'm now requesting to join the group.

domenic:
> To handle SyntaxError DOMException and SecurityError DOMException during normalizing parameters, this adds custom steps instead of [DelayWhilePrerendering].

We have previously had some discussions about this and thought it was OK to delay even such basic argument validation. See https://github.com/WICG/nav-speculation/pull/63#pullrequestreview-663708345 .

But, I don't think anyone felt too strongly about this. Do you think we should do the validation early?

jeremyroman:
I think the difference is that these aren't promise-returning operations.

If they were, then the exception would still be handled by the calling code. Since they're synchronous, though, we can't let author code handle the exception if we don't throw it synchronously, here. Instead it would be either unhandled or become an `error` event, both of which seem less convenient and more likely to break author code.

domenic:
Oh, that's a great point, and I'm sorry for missing it. Then, this LGTM!

nhiroki:
Jeremy perfetcly summarized what I wanted to say. Thanks!

PR #285, opened 2023-11-11
Update speculation rules security and privacy questionnaire.

jeremyroman:
This update mostly reflects progress in reflecting security and privacy considerations in the spec; the rest of the answers hold up pretty well.

### DIFF
--- a/speculation-rules-security-privacy-questionnaire.md
+++ b/speculation-rules-security-privacy-questionnaire.md
@@ -104,3 +104,11 @@ Yes:
 ### Do features in your specification enable origins to downgrade default security protections?
 
 No.
+
+### What happens when a document that uses your feature is kept alive in BFCache (instead of getting destroyed) after navigation, and potentially gets reused on future navigations back to the document?
+
+The prefetches and prerenders already triggered may remain in memory and be used after the document is restored from bfcache, if not expired. Additional speculation can occur after that point.
+
+### What happens when a document that uses your feature gets disconnected?
+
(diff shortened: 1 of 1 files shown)


speculation-rules-security-privacy-questionnaire.md
```diff
@@ -96,10 +96,19 @@ Browsers could choose to behave more conservatively, but doing so may make it po
...
 
 No.
+
+### What happens when a document that uses your feature is kept alive in BFCache (instead of getting destroyed) after navigation, and potentially gets reused on future navigations back to the document?
+
+The prefetches and prerenders already triggered may remain in memory and be used after the document is restored from bfcache, if not expired. Additional speculation can occur after that point.
+
+### What happens when a document that uses your feature gets disconnected?
+
+User agents should not initiate preloading in a document which is not fully active.
```
domenic, line 114:
```suggestion
User agents must not initiate preloading in a document which is not fully active.
```

? Is that specified somewhere we could link to?

jeremyroman: Yes, it is here: https://wicg.github.io/nav-speculation/speculation-rules.html#consider-speculation

> ⌛ If document is not [fully active](https://html.spec.whatwg.org/multipage/document-sequences.html#fully-active), then return.

There's a note/issue that we could be more explicit in our handling of bfcached documents, but the question relates to the fully active check, which we do have.

domenic (approved):

PR #289, opened 2023-11-23
Document rules: skip links which are not being rendered or in skipped contents

### LINKED ISSUE #246: Document rules should be limited to rendered links
see also https://chromium-review.googlesource.com/c/chromium/src/+/4226934

Users aren't likely to click on such links, and for selectors, Chromium's style engine (and likely others) avoid computing style for unrendered subtrees (due to display:none or shadow roots). So we should probably only consider links which are [being rendered](https://html.spec.whatwg.org/multipage/rendering.html#being-rendered).

cc @a4sriniv

jeremyroman:
It's hard for the user to interact with these links, and the author may not even intend them to be part of the document's palpable content.

In at least Chromium (and likely other implementations), these are also elements where the author is trying to skip work related to selector matching etc, which makes selector_matches applying to them particularly problematic. Better to ignore them altogether.

Fixes #246.

### DIFF
--- a/speculation-rules.bs
+++ b/speculation-rules.bs
@@ -545,6 +545,9 @@ A <dfn>prerender candidate</dfn> is a [=struct=] with the following [=struct/ite
 
         <div class="note">This corresponds to the elements which match '':any-link'' [=pseudo-class=], or which appear in the {{Document/links}} collection.</div>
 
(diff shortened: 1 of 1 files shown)


jeremyroman:
@a4sriniv ptal?

a4sriniv:
LGTM, thank you!

PR #295, opened 2024-01-10
Infer a speculation rule's source from the other keys

### LINKED ISSUE #169: Consider removing "source" from speculation rules objects

jeremyroman:
Fixes #169.

### DIFF
--- a/speculation-rules.bs
+++ b/speculation-rules.bs
@@ -371,10 +371,10 @@ add the following step
   To <dfn>parse a speculation rule</dfn> given a [=map=] |input|, a [=document=] |document|, and a [=URL=] |baseURL|, perform the following steps. They return a [=speculation rule=] or null.
 
   1. If |input| has any [=map/key=] other than "`source`", "`urls`", "`where`", "`requires`", "`target_hint`","`referrer_policy`", "`relative_to`", "`eagerness`", and "`expects_no_vary_search`" then return null.
+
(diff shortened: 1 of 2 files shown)


jeremyroman:
What do you think? First commit is my first instinct (if present validate it as now, just allow it to be absent). Second was your slight preference (ignore it entirely).

Honestly I think I have a slight preference for the former but don't feel super strongly.

This is pretty trivial and I'd hope we can use a lightweight process if this makes sense to you.

jeremyroman:
One case that ignoring it would break (that our unit tests caught) -- a catch-all could previously be written `{"source":"document"}` with no conditions, but that's not possible if implicit `"source"` is the only option. At least not unless we make `{}` also universal, which feels excessively weird.

domenic (approved):
I'm convinced, let's infer it!

Maybe also consider explainer updates in this PR?

domenic (approved):
Love it!!

PR #310, opened 2024-04-19
Spec the use of No-Vary-Search with prerender

kjmcnee:
The NVS hint from the speculation rule, if any, is now passed to the prerender creation.

The map of URLs to prerendering traversables is replaced with a list of a struct containing additional information about the prerender to support more flexible matching.

When the initial navigation of a prerendering traversable commits, we record its No-Vary-Search value.

When getting a prerender to activate, we now wait on potentially multiple ongoing prerenders that could match based on the NVS hint. This logic is similar to the prefetch case.

During activation, we update the prerender's URL if there is a difference in the prerendered document's URL and the URL navigated to in the case of inexact matching due to NVS.

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -267,15 +267,21 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
 
      <p class="note">We treat this initial navigations as |prerenderingTraversable| navigating itself, which will ensure all relevant security checks pass.
 
-  1. When the user indicates they wish to commit the navigation and create a new user-visible [=top-level traversable=] while doing so (e.g., by pressing <kbd><kbd>Shift</kbd>+<kbd>Enter</kbd></kbd> after typing in the address bar):
+  1. Let |record| be a [=prerender record=] with [=prerender record/starting URL=] |startingURL| and [=prerender record/prerendering traversable=] |prerenderingTraversable|.
 
-    1. Update the user agent's user interface to present |prerenderingTraversable|, e.g., by creating a new tab/window.
+  1. When the user indicates they wish to commit a navigation to an |activationUrl| which is a [=URL=] such that |record| [=prerender record/matches a URL=] given |activationUrl|:
 
-    1. [=prerendering traversable/Finalize activation=] for |prerenderingTraversable| given |startingURL|'s [=url/origin=].
+    1. If the user indicates they wish to create a new user-visible [=top-level traversable=] while doing so (e.g., by pressing <kbd><kbd>Shift</kbd>+<kbd>Enter</kbd></kbd> after typing in the address bar):
 
-  1. Alternately, when they use indicates they wish to commit the navigation into an existing [=top-level traversable=] |predecessorTraversable| (e.g., by pressing <kbd>Enter</kbd> after typing in the address bar):
+      1. TODO update URL
 
-    1. [=prerendering traversable/Activate=] |prerenderingTraversable| in place of |predecessorTraversable| given "`push`".
+      1. Update the user agent's user interface to present |prerenderingTraversable|, e.g., by creating a new tab/window.
+
+      1. [=prerendering traversable/Finalize activation=] for |prerenderingTraversable| given |startingURL|'s [=url/origin=].
+
+    1. Alternately, if the user indicates they wish to commit the navigation into an existing [=top-level traversable=] |predecessorTraversable| (e.g., by pressing <kbd>Enter</kbd> after typing in the address bar):
+
+      1. [=prerendering traversable/Activate=] |prerenderingTraversable| in place of |predecessorTraversable| given "`push`", |startingURL|, and |activationUrl|.
 
   <p class="note">The user might never indicate such a commitment, or might take long enough to do so that the user agent needs to reclaim the resources used by the prerender for some more immediate task. In that case the user agent can [=destroy a top-level traversable|destroy=] |prerenderingTraversable|.
 </div>
@@ -316,7 +322,7 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
 </div>
 
 <div algorithm>
-TODO activate call sites
+
   To <dfn for="prerendering traversable">activate</dfn> a [=prerendering traversable=] |successorTraversable| in place of a [=top-level traversable=] |predecessorTraversable| given a [=history handling behavior=] |historyHandling|, a [=URL=] |startingUrl|, a [=URL=] |activationUrl|, an optional [=navigation ID=] |navigationId|:
 
   1. [=Assert=]: |successorTraversable|'s [=navigable/active document=]'s [=Document/is initial about:blank=] is false.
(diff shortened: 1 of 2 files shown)


prerendering.bs
```diff
@@ -259,23 +304,48 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
...
+  1. Let |successorDocument| be |successorTraversable|'s [=navigable/active document=].
+
+  1. If |startingURL| equals |successorDocument|'s [=Document/URL=] and |startingURL| does not equal |activationURL|:
+
+    1. [=Assert=]: |successorDocument| [=can have its URL rewritten=] to |activationURL|.
+
+    1. Let |navigation| be |successorDocument|'s [=relevant global object=]'s [=navigation API=].
+
+    1. Let |continue| be the result of [=firing a push-replace-reload navigate event=] at |navigation| given "[=NavigationType/replace=]", |activationURL|, and true.
+       <p class="issue">Should we do this or just do the update without the event? TODO: resolve before submit
```
kjmcnee, line 338: @domenic in particular, for thoughts on whether a navigate event is appropriate here.

domenic: I feel moderately strongly that a navigate event is a good idea here. (Haven't looked at the rest of the PR yet.)

prerendering.bs
```diff
@@ -366,9 +438,28 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
...
+    1. Let |exactRecord| be null.
+    1. Let |inexactRecord| be null.
+    1. [=list/For each=] |record| of |predecessorDocument|'s [=Document/prerender records=]:
+        1. If |record|'s [=prerender record/prerendering traversable=]'s [=navigable/active document=]'s [=Document/is initial about:blank=] is true, then [=iteration/continue=].
+        1. If |record|'s [=prerender record/starting URL=] is equal to |url|:
+            1. Set |exactRecord| to |record|.
+            1. [=iteration/Break=].
+        1. If |inexactRecord| is null and |record| [=prerender record/matches a URL=] given |url|:
+            1. Set |inexactRecord| to |record|.
+    1. Let |recordToUse| be |exactRecord| if |exactRecord| is not null, otherwise |inexactRecord|.
```
jeremyroman, line 453: I think this works the same if we use a single variable, right?

```suggestion
    1. Let |recordToUse| be null.
    1. [=list/For each=] |record| of |predecessorDocument|'s [=Document/prerender records=]:
        1. If |record|'s [=prerender record/prerendering traversable=]'s [=navigable/active document=]'s [=Document/is initial about:blank=] is true, then [=iteration/continue=].
        1. If |record|'s [=prerender record/starting URL=] is equal to |url|:
            1. Set |recordToUse| to |record|.
            1. [=iteration/Break=].
        1. If |recordToUse| is null and |record| [=prerender record/matches a URL=] given |url|:
            1. Set |recordToUse| to |record|.
```

prerendering.bs
domenic: I agree canceling it would be a bit strange. However I think there's no reason to disallow that, similar to how we don't prevent a page from using `history.replaceState()` on activation to change its URL "back".

kjmcnee: The latest patch fires the event and removes this TODO.

prerendering.bs
```diff
@@ -190,6 +203,8 @@ By default, a [=navigable=]'s [=navigable/loading mode=] is "`default`". A navig
 
 <p class="note">Although there are only two values for the [=navigable/loading mode=], we use a flexible structure in anticipation of other future loading modes, such as those provided by fenced frames, portals, and uncredentialed (cross-site) prerendering. It's not yet clear whether that anticipation is correct; if, as those features gain full specifications, it turns out not to be, we will instead convert this into a boolean.
 
+Every [=prerendering traversable=] has an <dfn for="prerendering traversable">initial prerender search variance</dfn>, which is a [=URL search variance=] or null, and is initially null.
```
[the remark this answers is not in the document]
kjmcnee: Right. Renamed to "prerender initial response search variance" for clarity.

prerendering.bs
```diff
@@ -259,23 +304,48 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
 
      <p class="note">Currently, cross-site prerendering is not specified or implemented, although we have various ideas about how it could work in this repository's explainers.
 
-  1. If |referrerDoc|'s [=Document/prerendering traversables map=][|startingURL|] [=map/exists=], then return.
+  1. [=list/For each=] |record| of |referrerDoc|'s [=Document/prerender records=]:
+      1. If |record|'s [=prerender record/starting URL=] is |startingURL|, then return.
```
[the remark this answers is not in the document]
kjmcnee: On the prefetch side, it doesn't look like we spec any deduplication. I suppose it could be implicit from the prefetch steps themselves being optional?

Implementation-wise, currently both prefetch and prerender only dedupe based on the exact URL.

So I suppose the options are:
1. Remove the explicit deduplication from this prerender spec
2. Add explicit deduplication based on exact matches to the prefetch spec
3. Add explicit deduplication based on NVS matches to the prefetch and prerender specs and update the chromium implementation
4. Add explicit deduplication based on exact matches to the prefetch spec and add explicitly optional NVS based deduplication to prefetch and prerender specs

Can we think of a scenario where duplicate requests would be desirable for a UA to allow? If not, option 3 seems appropriate. WDYT?

prerendering.bs
```diff
@@ -259,23 +304,47 @@ Every {{Document}} has an <dfn for="Document">activation start time</dfn>, which
...
+
+  1. Let |successorDocument| be |successorTraversable|'s [=navigable/active document=].
+
+  1. If |startingURL| equals |successorDocument|'s [=Document/URL=] and |startingURL| does not equal |activationURL|:
+
+    1. [=Assert=]: |successorDocument| [=can have its URL rewritten=] to |activationURL|.
+
+    1. Let |navigation| be |successorDocument|'s [=relevant global object=]'s [=navigation API=].
+
+    1. Let |continue| be the result of <a>firing a push/replace/reload <code>navigate</code> event</a> at |navigation| given "[=NavigationType/replace=]", |activationURL|, and true.
```
kjmcnee, line 337: I can't figure out how to make the regular bikeshed linking work when there's a slash in the text of the definition, so I've ended up using the manual syntax.

kjmcnee:
Sorry for the delay. I think this is ready for further review. PTAL.

prerendering.bs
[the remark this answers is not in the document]
jeremyroman: I can imagine a scenario (while the site _can_ adapt to the query param being different, an exact match provides a "better" response which is fully ready faster), but I'm not sure it's worth indexing on. I have a leaning toward 3 as well, but maybe this should fork into an issue in case we want to weigh this a little more (and ask others)?

jeremyroman (approved):

prerendering.bs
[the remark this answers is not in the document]
kjmcnee: Filed #315

PR #325, opened 2024-07-17
Allow window.close() to work in prerendering navigables

domenic:
It turns out this only requires deleting text, as everything already works. For example, prerendering navigables always have a single-entry session history, so they are always script-closeable.

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -676,19 +676,6 @@ To <dfn export>get the supported loading modes</dfn> for a [=response=] |respons
 
 1. Return a [=list=] containing all elements of |slmHeader| that are [=structured header/tokens=].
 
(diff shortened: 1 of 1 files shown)


jeremyroman (approved):
lgtm as constructed, but is there any context to the motivation?

domenic:
The motivation is partners that want to shut down a prerender if it's no longer useful, e.g. if some other prerender is deemed more useful. (Imagine a calendar that prerenders the upcoming events; once the event is in the past, you can get rid of it.)

We've discussed this sort of thing before, in the context of giving the referrer page control via some sort of handle object. But that's a large design space. For the cases where the prerendered page has enough information by itself, we can just make `window.close()` work, without having to dip into that design space.

PR #328, opened 2024-08-21
Update triggers.md

robertlin-chromium:
Update the section `Window name targeting hints` about intention of having target_hint field, and
add a link to the progress tracker.
The changes are for https://github.com/w3ctag/design-reviews/issues/931


triggers.md
```diff
@@ -160,6 +160,7 @@ To help with this, `"prerender"` rules can have a `"target_hint"` field, which c
 ```
 
 This is just a hint, and is not binding on the implementation. Indeed, we hope that it one day becomes unnecessary, and all implementations can activate prerendered content into any target window. At that point, the field can be safely ignored, and removed from the specification. But at least for Chromium, getting to that point might take a year or so of engineering effort, so in the meantime `"target_hint"` gives developers a way to use prerendering in combination with new windows.
+To check the progress of removing target_hint field, please to [progress tracker](https://crbug.com/361129302).
```
domenic, line 163:
```suggestion

To check Chromium's progress on removing the `target_hint` field, see [this issue](https://crbug.com/361129302).
```

domenic (approved):
LGTM with nit, will fix and merge

PR #338, opened 2024-09-27
Editorial: Move No-Vary-Search to IETF HTTPbis

jeremyroman:


jeremyroman:
I've already migrated the labeled issues from both here and jeremyroman/http-no-vary-search, so hopefully this gives enough breadcrumbs to lead to a single working copy.

jeremyroman:
One thing this _doesn't_ yet do is adjust algorithms which refer to the No-Vary-Search spec to point at the IETF draft. I should probably do that, too, unless you disagree.

jeremyroman:
Okay, now it does do that. Do you know offhand how to make the IETF version win in the generation of https://github.com/tobie/specref/blob/main/refs/browser-specs.json (which feeds Bikeshed etc bibliography generation)? I'd update it manually except that I think it might just get automatically overwritten by a bot.

no-vary-search.bs
```diff
@@ -45,9 +45,7 @@ spec: RFC8941; urlPrefix: https://www.rfc-editor.org/rfc/rfc8941.html
 
 <h2 id="status-and-venue">Status and venue note</h2>
 
-This document is being written as a web-style specification in the WICG for now, because that's the tooling and venue the author is familiar with. Its purpose is to nail down some details of the processing model in order to make writing and testing prototypes easier.
-
-In the longer term, we envision this header being specified in a HTTPWG RFC, alongside whatever portion of the processing model can be shared among its various consumers. (That is, between both web platform specifications such as [[FETCH]], and HTTP specifications such as future modifications to [[RFC9111]].) It's just incubating in WICG for now.
+Work on this document has moved to the IETF HTTP working group. The <a href="https://httpwg.org/http-extensions/draft-ietf-httpbis-no-vary-search.html">latest draft</a> and <a href="https://github.com/httpwg/http-extensions/labels/no-vary-search">issue tracker</a> can be found there.
 
 <h2 id="header-definition">HTTP header field definition</h2>
```
domenic, line 50: I think we should gut this spec entirely, maybe replacing it with a (client-side) redirect.

domenic (approved):

PR #340, opened 2024-10-04
Exempt the Speculation-Rules header from CSP

kjmcnee:
Per discussions, this is outside of CSP's threat model.


domenic (approved):
Lovely comprehensive update, including a future-facing note.

Issue #315, opened 2024-05-07
Should the spec be explicit about deduplicating prefetches/prerenders?


kjmcnee:
Spinning off discussion in https://github.com/WICG/nav-speculation/pull/310#discussion_r1575338094

In the chromium implementation, we currently dedupe prefetches and prerenders based on the exact URL. That is, if we have an existing prefetch/prerender to a given URL, we don't start another request to the same URL.

In the prefetch spec, we don't explicitly mention deduplication. In the prerender spec, we explicitly dedupe based on an exact URL match. In either case, the steps are optional, so there's presumably an implicit option for the UA to dedupe.

If we do want to be explicit in the spec about deduplication, should this also incorporate No-Vary-Search?

Options here could be:
1. Remove the explicit deduplication from this prerender spec
2. Add explicit deduplication based on exact matches to the prefetch spec
3. Add explicit deduplication based on NVS matches to the prefetch and prerender specs and update the chromium implementation
4. Add explicit deduplication based on exact matches to the prefetch spec and add explicitly optional NVS based deduplication to prefetch and prerender specs

domenic:
I am in favor of option 3, unless there is some subtle problem there I am not realizing.

kjmcnee:
Option 3 sounds like everyone's preferred option.

It would also allow us to simplify the matching logic to not need to deal with multiple potential matches.

kjmcnee:
Unassigning myself due to a team change.

domenic:
https://github.com/WICG/nav-speculation/pull/353 ends up containing fixes for this, I believe, because I was working on nearby areas of the spec and it seemed broken.

PR #353, opened 2025-01-29
Add an explainer for speculation rules tags

domenic:


domenic:
I think attempting to do the spec in this same PR is not going to work well as I am uncovering lots of other interesting issues (like #315). Tentative plan:

- Reduce this PR to just the explainer
- One spec PR for making all prerenders go through the prefetch path
- One spec PR for fixing all prefetch and prerender matching to incorporate no vary search (hint)s, and making the deduplication logic explicit
- Then a spec PR for tags on top of those

PR #356, opened 2025-01-31
Clarify interaction between prefetch and early hints

### LINKED ISSUE #339: Clarify 103 response code support
Was asked if 103 early hints prevents navigation speculation. We tested and they don't (at least for prefetch), as it's the final response status that counts, but it could maybe be made clearer in the spec which has things like this:

> A [response](https://fetch.spec.whatwg.org/#concept-response) response _supports prefetch_ if the following steps return true:
> 1. Let _status_ be _response_’s [status](https://fetch.spec.whatwg.org/#concept-response-status).
> 2. [Assert](https://infra.spec.whatwg.org/#assert): _status_ is not a [redirect status](https://fetch.spec.whatwg.org/#redirect-status).
> 3. If _status_ is not an [ok status](https://fetch.spec.whatwg.org/#ok-status), then return false.
> 4. Return true.

With a link to [fetch's definition of ok-status](https://fetch.spec.whatwg.org/#ok-status):

> **2.2.3. Statuses**
> A **status** is an integer in the range 0 to 999, inclusive.
> A **null body status** is a [status](https://fetch.spec.whatwg.org/#concept-status) that is 101, 103, 204, 205, or 304.
> An **ok status** is a [status](https://fetch.spec.whatwg.org/#concept-status) in the range 200 to 299, inclusive.
> A **redirect status** is a [status](https://fetch.spec.whatwg.org/#concept-status) that is 301, 302, 303, 307, or 308.

204 and 205 have special exclusions in [the spec for prerendering](https://wicg.github.io/nav-speculation/prerendering.html#:~:text=s%20status%20is-,204%20or%20205,-%2C), 304 is not yet supported as per #330, 101 is not really revelant for prerendering navigations, which leaves 103 as kind of unclear.

Given both 103 Early Hints and Nav Speculations are performance optimizations it would be good to be clearer about this.

domenic:
Closes #339.

@tunetheweb PTAL?

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -357,10 +357,9 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
            <p class="note">In the case where the above <span class=allow-2119>optional</span> step is taken, and thus |record| could end up generating multiple [=navigation params=] via multiple calls to this algorithm, it is OK that all such [=navigation params=] share the same [=navigation params/fetch controller=]. This is because the only use of a [=navigation params=]'s [=navigation params/fetch controller=] is to calculate navigation timing information, which we want to be the same for any reuses.</p>
         :  [=navigation params/commit early hints=]
         :: null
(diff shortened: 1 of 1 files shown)


tunetheweb (approved):
LGTM. Thanks.

PR #360, opened 2025-02-03
Explicitly dedupe prefetches and prerenders

### LINKED ISSUE #315: Should the spec be explicit about deduplicating prefetches/prerenders?

domenic:
Closes #315. The rule is that the first candidate wins.

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -319,7 +319,7 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
 
     1. If |recordUnderConsideration|'s [=prefetch record/No-Vary-Search hint=] is not equal to |recordNVS|, then [=iteration/continue=].
 
-    1. If |recordUnderConsideration| [=prefetch record/URL=] and |record|'s [=prefetch record/URL=] are [=equivalent modulo search variance=] given |recordNVS|, then return true.
+    1. If |recordUnderConsideration|'s [=prefetch record/URL=] and |record|'s [=prefetch record/URL=] are [=equivalent modulo search variance=] given |recordUnderConsideration|'s [=prefetch record/No-Vary-Search hint=], then return true.
 
   1. Return false.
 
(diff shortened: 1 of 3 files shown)


prefetch.bs
```diff
@@ -299,6 +303,47 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
...
+
+    1. Let |recordNVS| be null.
+
+    1. If |record|'s [=prefetch record/response=] is not null, then set |recordNVS| to the result of [=obtaining a URL search variance=] given |record|'s [=prefetch record/response=].
+
+    1. Otherwise, set |recordNVS| to |record|'s [=prefetch record/No-Vary-Search hint=].
+
+    1. If |recordUnderConsideration|'s [=prefetch record/No-Vary-Search hint=] is not equal to |recordNVS|, then [=iteration/continue=].
+
+    1. If |recordUnderConsideration| [=prefetch record/URL=] and |record|'s [=prefetch record/URL=] are [=equivalent modulo search variance=] given |recordNVS|, then return true.
```
hiroshige-g, line 322: nit: `|recordUnderConsideration|'s [=prefetch record/URL=]` ?

Also `[=equivalent modulo search variance=] given |recordUnderConsideration|'s [=prefetch record/No-Vary-Search hint=]` is a little clearer that the equivalence relation is the same during the iteration (this doesn't affect the behavior though).

hiroshige-g:
LGTM.

There are related issues (switching from NVS hint to NVS header, whether to merge or cancel prefetch requests on deduplication) but they are orthogonal, so it's fine to land this PR to land the deduping condition so far.

PR #361, opened 2025-02-06
Add Clear-Site-Data Clear Cache interaction with prerender

srwei:
Corresponding CL: https://chromium-review.googlesource.com/c/chromium/src/+/6173154

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -644,9 +644,11 @@ This stores which client hints each origin has opted into receiving, until it ca
 
 <h3 id="interaction-with-clear-site-data-clear-cache">Interaction with Clear-Site-Data Clear Cache</h3>
 
-Modify [$clear cache for origin$] to add the following step at the end.
+<div algorithm="clear cache for origin patch">
+Modify [$clear cache for <var ignore>origin</var>$] to add the following step at the end.
 1. [=list/For each=] [=navigable/top-level traversable=] |traversable| in the user agent's [=top-level traversable set=]: 
   1. If |traversable|'s [=navigable/active document=]'s [=Document/origin=] is [=same origin=] with |origin|, and |traversable| is a [=prerendering traversable=], then [=destroy a top-level traversable|destroy=] |traversable|.
+</div>
 
 <h2 id="supports-loading-mode">The \`<dfn export http-header><code>Supports-Loading-Mode</code></dfn>\` HTTP response header</h2>
 
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -45,6 +45,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     text: session history entry; url: browsing-the-web.html#session-history-entry
     text: the worker's lifetime; url: workers.html#the-worker's-lifetime
     text: URL and history update steps; url: browsing-the-web.html#url-and-history-update-steps
+    text: top-level traversable set; url: document-sequences.html#top-level-traversable-set
     for: Document
       text: abort; url: document-lifecycle.html#abort-a-document
       text: destroy; url: document-lifecycle.html#destroy-a-document
@@ -119,6 +120,9 @@ spec: web-audio; urlPrefix: https://webaudio.github.io/web-audio-api/
     text: running; url: dom-audiocontextstate-running
     text: queue a control message; url: queuing
     text: [[pending resume promises]]; url: dom-audiocontext-pending-resume-promises-slot
+spec: webappsec-clear-site-data; urlPrefix: https://w3c.github.io/webappsec-clear-site-data/
+  type: dfn
+    text: Clear cache for origin; url: clear-cache
 
 </pre>
 <pre class="biblio">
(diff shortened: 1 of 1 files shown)


prerendering.bs
```diff
@@ -119,6 +120,9 @@ spec: web-audio; urlPrefix: https://webaudio.github.io/web-audio-api/
     text: running; url: dom-audiocontextstate-running
     text: queue a control message; url: queuing
     text: [[pending resume promises]]; url: dom-audiocontext-pending-resume-promises-slot
+spec: webappsec-clear-site-data; urlPrefix: https://w3c.github.io/webappsec-clear-site-data/
+  type: dfn
+    text: Clear cache for origin; url: clear-cache
```
domenic, line 125: I think you can remove this line as long as you use `[$clear cache for origin$]` instead of `[=clear cache for origin=]`.

This is because they have, for some reason, chosen to define this algorithm as an "abstract operation" instead of a normal definition.

prerendering.bs
```diff
@@ -641,6 +645,12 @@ This stores which client hints each origin has opted into receiving, until it ca
   1. Otherwise, [=map/set=] [=Accept-CH cache=][|origin|] to |hintSet|.
 </div>
 
+<h3 id="interaction-with-clear-site-data-clear-cache">Interaction with Clear-Site-Data Clear Cache</h3>
+
+Modify [=Clear cache for origin=] to add the following step at the end.
+1. [=list/For each=] [=navigable/top-level traversable=] |traversable| in the user agent's [=top-level traversable set=]: 
+  1. If |traversable| is a [=prerendering traversable=], then [=destroy a top-level traversable|destroy=] |traversable|.
```
domenic, line 652: This needs to be scoped to _origin_. You can do that by checking if `|traversable|'s [=navigable/active document=]'s [=Document/origin=] is [=same origin=] with |origin|`.

domenic:
> ipr — PR has contribution issues. The following users' affiliation could not be determined: @srwei. 

For this, you can work to link your GitHub account and your W3C account, and ensure your W3C account [joins the WICG](https://wicg.io/).

prerendering.bs
srwei: Agreed! Updated to include scoping to origin.

domenic (approved):

prerendering.bs
```diff
@@ -641,6 +642,12 @@ This stores which client hints each origin has opted into receiving, until it ca
   1. Otherwise, [=map/set=] [=Accept-CH cache=][|origin|] to |hintSet|.
 </div>
 
+<h3 id="interaction-with-clear-site-data-clear-cache">Interaction with Clear-Site-Data Clear Cache</h3>
+
+Modify [$clear cache for origin$] to add the following step at the end.
+1. [=list/For each=] [=navigable/top-level traversable=] |traversable| in the user agent's [=top-level traversable set=]: 
+  1. If |traversable|'s [=navigable/active document=]'s [=Document/origin=] is [=same origin=] with |origin|, and |traversable| is a [=prerendering traversable=], then [=destroy a top-level traversable|destroy=] |traversable|.
```
domenic, line 649: Use `<var ignore>` here to bypass the build error. This is unfortunately necessary very often when writing these kind of patch specs.

prerendering.bs
```diff
@@ -641,6 +642,12 @@ This stores which client hints each origin has opted into receiving, until it ca
   1. Otherwise, [=map/set=] [=Accept-CH cache=][|origin|] to |hintSet|.
 </div>
 
+<h3 id="interaction-with-clear-site-data-clear-cache">Interaction with Clear-Site-Data Clear Cache</h3>
+
+Modify [$clear cache for origin$] to add the following step at the end.
```
domenic, line 647: Wrap this whole thing in `<div algorithm="clear cache for origin patch">` `</div>`

domenic (commented):
This looked good and then I looked at the build result and realized we had some minor updates to make. But the normative content is solid and we can merge as soon as those minor fixes are done.

prerendering.bs
domenic: You can find all publicly-exported definitions in https://dontcallmedom.github.io/webdex/, so e.g. I was able to find https://dontcallmedom.github.io/webdex/c.html#clear%20cache%20for%20origin%40%40clear-site-data%25%25abstract-op

We have a large anchors block because we are specifically poking at a lot of non-public definitions, due to our nature as a spec full of patches. "Normal" specs just get magic autolinking via the webdex database.

prerendering.bs
srwei: Looking from the other usages in the file for wrapping for patches, I assumed you meant the block in the other comment? Let me know if I need to change it, thanks!

prerendering.bs
domenic: It was actually the one in the algorithm.

I pushed a commit which fixes the build errors. Note that you can run the build locally using `make prerendering.html`. But it's all good now, so I'll merge!

PR #366, opened 2025-03-11
Add Clear-Site-Data Clear Cache interaction with prefetch

srwei:
Corresponding CL: https://chromium-review.googlesource.com/c/chromium/src/+/6311760

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -869,7 +869,7 @@ The header field is an [[RFC9651]] Structured Header whose value must be a [=str
   Modify [$clear cache for origin$] to add the following step at the end.
 
   1. [=list/For each=] [=prefetch record=] |prefetchRecord| in the user agent's |activeDocument|'s [=Document/prefetch records=]:
-  1. If |prefetchRecord|'s [=document state/initiator origin=] is [=same origin=] with |activeDocument|'s [=Document/origin=], then [=prefetch record/cancel and discard=] |prefetchRecord| given |document|.
+  1. If |prefetchRecord|'s [=document state/initiator origin=] is [=same origin=] with |activeDocument|'s [=Document/origin=], then [=prefetch record/cancel and discard=] |prefetchRecord| given <var ignore>origin</var>.
 </div>
 
 
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -100,6 +100,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: determine the origin; url: determining-the-origin
       text: navigable container; url: navigable-container
       text: parent; url: nav-parent
+      text: top-level traversable set; url: top-level-traversable-set
     urlPrefix: dom.html
       text: cross-origin opener policy; for: Document; url: concept-document-coop
     urlPrefix: semantics.html
@@ -713,6 +714,17 @@ This section contains patches to [[NAVIGATION-TIMING]].
     Add an additional parameter to [=create the navigation timing entry=], which is a [=string=] <var ignore>deliveryType</var>, and pass it as an additional argument to [=setup the resource timing entry=].
 </div>
 
+<h2 id="interaction-with-other-specs">Interaction with other specifications and concepts</h2>
+
+<div algorithm="clear prefetch cache for origin patch">
+  Modify [$clear cache for origin$] to add the following step at the end.
+
+  1. [=list/For each=] [=navigable/top-level traversable=] |traversable| in the user agent's [=top-level traversable set=]:
+    1. Let |navigables| be the [=list=] of [=Document/inclusive descendant navigables=] from |traversable|'s [=navigable/active document=].
+    2. [=list/For each=] [=navigable/active document=] |activeDocument| in |navigables|:
+      1. [=prefetch record/cancel and discard=] |activeDocument|'s [=Document/prefetch records=].
+</div>
+
 <h2 id="prefetch-algorithms">Prefetch algorithms</h2>
 
 <p class="issue">Check Service Worker integration</p>
@@ -862,16 +874,6 @@ The following parameters are defined for the `prefetch` token:
 The \`<code><dfn http-header>Sec-Speculation-Tags</dfn></code>\` HTTP request header specifies the web developer-provided tags associated with the speculative navigation request. It can also be used to distinguish speculative navigation requests from other speculative requests, since [:Sec-Purpose:] can be sent by both categories of request.
 
 The header field is an [[RFC9651]] Structured Header whose value must be a [=structured header/List=]. The list can contain either [=structured header/Token=] or [=structured header/String=] values. String values represent developer-provided tags, whereas token values represent predefined tags. As of now, the only predefined tag is `null`, which indicates a speculative navigation request with no developer-defined tag.
-
-<h2 id="interaction-with-other-specs">Interaction with other specifications and concepts</h2>
-
-<div algorithm="clear prefetch cache for origin patch">
-  Modify [$clear cache for origin$] to add the following step at the end.
-
-  1. [=list/For each=] [=prefetch record=] |prefetchRecord| in the user agent's |activeDocument|'s [=Document/prefetch records=]:
-  1. If |prefetchRecord|'s [=document state/initiator origin=] is [=same origin=] with |activeDocument|'s [=Document/origin=], then [=prefetch record/cancel and discard=] |prefetchRecord| given <var ignore>origin</var>.
-</div>
-
 
 <h2 id="security-considerations">Security considerations</h2>
 
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -723,7 +723,7 @@ This section contains patches to [[NAVIGATION-TIMING]].
     2. [=list/For each=] |navigable| in |navigables|:
       1. Let |activeDocument| be the |navigable|'s [=navigable/active document=].
       2. [=list/For each=] |prefetchRecord| in |activeDocument|'s [=Document/prefetch records=]:
-        1. [=prefetch record/cancel and discard=] the |prefetchRecord|.
+        1. [=prefetch record/Cancel and discard=] |prefetchRecord| given |activeDocument|.
 </div>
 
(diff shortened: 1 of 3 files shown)


prefetch.bs
```diff
@@ -863,6 +863,16 @@ The \`<code><dfn http-header>Sec-Speculation-Tags</dfn></code>\` HTTP request he
 
 The header field is an [[RFC9651]] Structured Header whose value must be a [=structured header/List=]. The list can contain either [=structured header/Token=] or [=structured header/String=] values. String values represent developer-provided tags, whereas token values represent predefined tags. As of now, the only predefined tag is `null`, which indicates a speculative navigation request with no developer-defined tag.
 
+<h2 id="interaction-with-other-specs">Interaction with other specifications and concepts</h2>
```
domenic, line 866: Let's make this section "Clear Site Data patches" and put it below "Navigation Timing patches".

prefetch.bs
```diff
@@ -863,6 +863,16 @@ The \`<code><dfn http-header>Sec-Speculation-Tags</dfn></code>\` HTTP request he
...
+
+<div algorithm="clear prefetch cache for origin patch">
+  Modify [$clear cache for origin$] to add the following step at the end.
+
+  1. [=list/For each=] [=prefetch record=] |prefetchRecord| in the user agent's |activeDocument|'s [=Document/prefetch records=]:
```
domenic, line 871: |activeDocument| is not defined anywhere, and it doesn't sound like something the user agent has.

I think that, like https://wicg.github.io/nav-speculation/prerendering.html#interaction-with-clear-site-data-clear-cache , we should start from the user agent's top-level traversable set. From there we can get the [inclusive descendant navigables](https://html.spec.whatwg.org/#inclusive-descendant-navigables) and look at each of their `[=navigable/active document=]`s.

prefetch.bs
```diff
@@ -863,6 +863,16 @@ The \`<code><dfn http-header>Sec-Speculation-Tags</dfn></code>\` HTTP request he
...
+  1. If |prefetchRecord|'s [=document state/initiator origin=] is [=same origin=] with |activeDocument|'s [=Document/origin=], then [=prefetch record/cancel and discard=] |prefetchRecord| given <var ignore>origin</var>.
```
domenic, line 872: This should be indented so that it's nested in the output. (I realize we didn't catch this on prerendering... I'll push a quick commit fixing it.)

domenic, line 872: "initiator origin" is a property of a document state (as indicated by the syntax `[=document state/initiator origin=]`). It is not a property of prefetch records.

I think no such test is necessary here (although it might be in the implementation) since in the spec the prefetch records live on documents and so origin partitioning is automatic.

domenic (changes requested):

prefetch.bs
srwei: Taking a look at how [inclusive descendant navigables ](https://html.spec.whatwg.org/#inclusive-descendant-navigables) is described as being "inclusive ancestor navigables of a Document", I'm a bit confused when you mean getting all the active documents from each inclusive descendant navigable.  My understanding was that for each traversable, there is one active document, with each active document potentially having multiple inclusive descendent navigables (line above the link)

prefetch.bs
srwei: I think I see what you mean here. If we [cancel and discard](https://wicg.github.io/nav-speculation/prefetch.html#prefetch-record-cancel-and-discardl) all prefetchRecords on a particular Document (in this case, the activeDocument where Clear-Site-Header header is coming from), then we know that their referring origin is same as Document's origin, since as you mentioned, there already is origin partitioning. Does that sound right to you?

prefetch.bs
domenic: Yes, your understanding is basically correct. So the route to get all documents in the entire user agent would be something like:

```js
const topLevelTraversables = userAgent.topLevelTraversableSet();
const allNavigables = topLevelTraversables.flatMap(t => inclusiveDescendantNavigables(t.activeDocument));
const allDocuments = allNavigables.map(n => n.activeDocument);
```

(Expressing this in specs probably looks like a version of this that only uses `for`-`of` instead of `flatMap()`/`map()`.)

prefetch.bs
domenic: Yep, that sounds right to me!

srwei: Thank you!

prefetch.bs
srwei: Thanks! Just pushed with wording that was based on other HTML specs. Not sure if it would be better to have 3 for-loop statements here, so defined a list instead for one of them as it seems easier to read for me.

prefetch.bs
```diff
@@ -713,6 +714,18 @@ This section contains patches to [[NAVIGATION-TIMING]].
     Add an additional parameter to [=create the navigation timing entry=], which is a [=string=] <var ignore>deliveryType</var>, and pass it as an additional argument to [=setup the resource timing entry=].
 </div>
 
+<h2 id="interaction-with-other-specs">Interaction with other specifications and concepts</h2>
+
+<div algorithm="clear prefetch cache for origin patch">
+  Modify [$clear cache for origin$] to add the following step at the end.
+
+  1. [=list/For each=] [=navigable/top-level traversable=] |traversable| in the user agent's [=top-level traversable set=]:
+    1. Let |navigables| be the [=list=] of [=Document/inclusive descendant navigables=] from |traversable|'s [=navigable/active document=].
```
domenic, line 723:
```suggestion
    1. Let |navigables| be the [=Document/inclusive descendant navigables=] of |traversable|'s [=navigable/active document=].
```

prefetch.bs
```diff
@@ -713,6 +714,18 @@ This section contains patches to [[NAVIGATION-TIMING]].
...
+    2. Let |documents| be the [=list=] of |navigables|' [=navigable/active document=]s.
+    3. [=list/For each=] [=navigable/active document=] |activeDocument| in |documents|:
```
domenic, line 725: It's probably cleaner to loop over _navigables_ and then inside the loop look at `|navigable|'s [=navigable/active document=]`.

prefetch.bs
```diff
@@ -713,6 +714,18 @@ This section contains patches to [[NAVIGATION-TIMING]].
...
+      1. [=prefetch record/cancel and discard=] |activeDocument|'s [=Document/prefetch records=].
```
domenic, line 726: You need to loop over each prefetch record. "Cancel and discard" takes a prefetch record + a document, but `|activeDocument|'s [=Document/prefetch records=]` is a list. Make sure to match up the type signatures at the call site and declaration site.

prefetch.bs
```diff
@@ -713,6 +714,18 @@ This section contains patches to [[NAVIGATION-TIMING]].
     Add an additional parameter to [=create the navigation timing entry=], which is a [=string=] <var ignore>deliveryType</var>, and pass it as an additional argument to [=setup the resource timing entry=].
 </div>
 
+<h2 id="interaction-with-other-specs">Interaction with other specifications and concepts</h2>
+
+<div algorithm="clear prefetch cache for origin patch">
+  Modify [$clear cache for origin$] to add the following step at the end.
+
+  1. [=list/For each=] [=navigable/top-level traversable=] |traversable| in the user agent's [=top-level traversable set=]:
```
domenic, line 722:
```suggestion
  1. [=list/For each=] [=navigable/top-level traversable=] |traversable| of the user agent's [=top-level traversable set=]:
```

(per https://infra.spec.whatwg.org/#list-iterate)

prefetch.bs
```diff
@@ -100,6 +100,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: determine the origin; url: determining-the-origin
       text: navigable container; url: navigable-container
       text: parent; url: nav-parent
+      text: top-level traversable set; url: top-level-traversable-set
```
domenic, line 103: You can delete this as long as you change `[=top-level traversable set=] ` to `[=user agent/top-level traversable set=]`.

(You can tell by looking at https://dontcallmedom.github.io/webdex/t.html#top-level%20traversable%20set%40%40user%20agent%40dfn which says `(concept for user agent)`.)

domenic (commented):
This approach looks good, just some suggestions on how to word it better to fit the usual patterns for spec language.

prefetch.bs
srwei: Yeah I agree, I was trying to prevent having too many loops, but this way is more readable. Realizing the spec is written very much how code would function, but just in spec related syntax.

prefetch.bs
srwei: Thanks, updated the usage in prerendering as well to fit the spec for list-iterate.

prefetch.bs
srwei: Ah I keep forgetting to check this github before making additions to include external specs. Will try to keep this in mind, thanks. Also made the corresponding change in prerendering.bs.

prefetch.bs
```diff
@@ -713,6 +713,19 @@ This section contains patches to [[NAVIGATION-TIMING]].
...
+
+<div algorithm="clear prefetch cache for origin patch">
+  Modify [$clear cache for origin$] to add the following step at the end.
+
+  1. [=list/For each=] [=navigable/top-level traversable=] |traversable| of the user agent's [=user agent/top-level traversable set=]:
+    1. Let |navigables| be the [=Document/inclusive descendant navigables=] of |traversable|'s [=navigable/active document=].
+    2. [=list/For each=] |navigable| in |navigables|:
+      1. Let |activeDocument| be the |navigable|'s [=navigable/active document=].
+      2. [=list/For each=] |prefetchRecord| in |activeDocument|'s [=Document/prefetch records=]:
+        1. [=prefetch record/cancel and discard=] the |prefetchRecord|.
```
domenic, line 726:
```suggestion
        1. [=prefetch record/Cancel and discard=] |prefetchRecord| given |activeDocument|.
```

domenic (approved):
LGTM with one last call site mismatch. Thanks so much!

PR #369, opened 2025-03-19
Clarify that tags do not contain control characters

domenic:
/cc @robertlin-chromium @nhiroki

The Structured Fields specification already required this but the specification was not properly failing at parse time.


nhiroki (approved):
LGTM, thanks!

PR #376, opened 2025-05-01
Add "prefetchCache" and "prerenderCache" Clear Site Data support

### LINKED ISSUE #357: Specify clear-site-data integration, and special keywords for prefetch and prerender
We should specify (and implement) that clear-site-data clears all active prerenders and prefetches, when you use the `"cache"` value. This would be done by patching the https://w3c.github.io/webappsec-clear-site-data/#clear-cache spec (probably via more monkeypatches in the prefetch / prerender specs).

Additionally, we should let origins specifically clear these caches in a more targeted way. I suggest introducing new values, `"prefetchCache"` and `"prerenderCache"`, which perform this targeted clearing. This helps with the use cases discussed in #352; see especially https://github.com/WICG/nav-speculation/issues/352#issuecomment-2575373886.

domenic:
Closes #357. See the explainer at https://github.com/WICG/nav-speculation/blob/main/clear-site-data-integration.md.

This also fixes a bug in the previously-specified prefetch integration where the origin scoping was missing.

@srwei please review!

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -713,16 +713,34 @@ This section contains patches to [[NAVIGATION-TIMING]].
     Add an additional parameter to [=create the navigation timing entry=], which is a [=string=] <var ignore>deliveryType</var>, and pass it as an additional argument to [=setup the resource timing entry=].
 </div>
 
-<h2 id="interaction-with-other-specs">Interaction with other specifications and concepts</h2>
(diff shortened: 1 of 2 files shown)


srwei (approved):
LGTM, those were the 3 places to update. Thanks for helping with this! Can use the nav-speculation spec for the wide reviews once merged!

Issue #350, opened 2025-01-02
No-Vary-Search can cause reserved client creation URL mismatch


hiroshige-g:
When the prefetch is served using No-Vary-Search matching (which is not yet specified yet?), the navigation param's reserved client's creation URL doesn't match with the navigation URL.

For example,

- Prefetch `/page?param=1`. The server responds with `No-Vary-Search: "param"` response header.
- Navigate to `/page?param=2`. The prefetched response is served due to the No-Vary-Search header.

Then the resulting Window's settings object's creation URL is `/page?param=1`, because in the current spec it is derived from the prefetch's navigation param's reserved client.

The observable behavior change includes controlling ServiceWorker: if we register two service workers (that just fall back to the network) with scopes `/page?param=1` and `/page?param=2`, respectively, which service worker should be the controller of the resulting page?
In the current spec, the controller would be the service worker with scope `/page?param=1` (which is set during prefetch), and remains the same even after e.g. `claim()` (because settings object's creation URL itself is also `/page?param=1`), while it would be the service worker with scope `/page?param=2` without the prefetch.

Candidate soloutions (just brainstorming, not yet sure at all):

- Just allow this mismatch. I feel the behavioral impact in the real world is more like theoretical (page authers can register a service worker controlling all URLs under No-Vary-Search matching and actually very likely do so). I'm not fully sure how the creation URL mismatch can affect things other than ServiceWorkers though.
- Fix the Window's settings object's creation URL only -- maybe relatively easy? and saves a little more consistency and the controller would be the service worker with scope `/page?param=2` after `claim()`.
- Also fix ServiceWorker controller (even before `claim()`)? But this needs going through ServiceWorker registration lookup etc. again, and I have to plumb a new path to [Match Service Worker Registration](https://w3c.github.io/ServiceWorker/#match-service-worker-registration), and I'm not sure this isn't something that can be done very immediately (I'll try another iteration later).

hiroshige-g:
@nhiroki @domenic FYI

domenic:
Sorry for the delay in getting back to this.

> When the prefetch is served using No-Vary-Search matching (which is not yet specified yet?),

I think it is specified: https://wicg.github.io/nav-speculation/prefetch.html#prefetch-record-matches-a-url

> Then the resulting Window's settings object's creation URL is `/page?param=1`, because in the current spec it is derived from the prefetch's navigation param's reserved client.

[...]

Issue #347, opened 2024-12-02
Prefetch requests keeping the service worker alive and preventing it from revving its version


domenic:
Another issue @hiroshige-g explained to me is that service worker lifetime and version cycle is tied to the existence of clients. In particular, a service worker cannot upgrade to a new version while clients are alive.

Because we store the reserved client in the prefetch record, this means as long as there are any non-discarded prefetches, a site's service worker will not upgrade.

It'd be better if we allowed service worker upgrades, and evicted the associated prefetch records.

How exactly to specify this is not yet clear to me, as I haven't read up on how service worker lifetime and upgrades work. I worry that it might come down to https://w3c.github.io/ServiceWorker/#control-and-use which states 

> The rest of the section is non-normative.

> The behavior in this section is not fully specified yet and will be specified in HTML Standard. The work is tracked by the [issue](https://github.com/w3c/ServiceWorker/issues/765) and the [pull request](https://github.com/whatwg/html/pull/2809).

which links to a PR from 2017 that we never managed to land :(

domenic:
Detective work for today: we need to look at https://w3c.github.io/ServiceWorker/#installation-algorithm . At the bottom it says

> Invoke [Try Activate](https://w3c.github.io/ServiceWorker/#try-activate) with registration.
>
> Note: If [Try Activate](https://w3c.github.io/ServiceWorker/#try-activate) does not trigger [Activate](https://w3c.github.io/ServiceWorker/#activate) here, [Activate](https://w3c.github.io/ServiceWorker/#activate) is tried again when the last client controlled by the existing [active worker](https://w3c.github.io/ServiceWorker/#dfn-active-worker) is [unloaded](https://w3c.github.io/ServiceWorker/#handle-service-worker-client-unload), [skipWaiting()](https://w3c.github.io/ServiceWorker/#dom-serviceworkerglobalscope-skipwaiting) is asynchronously called, or the [extend lifetime promises](https://w3c.github.io/ServiceWorker/#extendableevent-extend-lifetime-promises) for the existing [active worker](https://w3c.github.io/ServiceWorker/#dfn-active-worker) settle.

so those are the next places to look. The problem described is, I guess, [Try Activate](https://w3c.github.io/ServiceWorker/#try-activate) will not work as we'd hope.

The problematic clause seems to be

> no [service worker client](https://w3c.github.io/ServiceWorker/#dfn-service-worker-client) is [using](https://w3c.github.io/ServiceWorker/#dfn-use) registration

[...]

Issue #351, opened 2025-01-02
ServiceWorker controller mismatch when prefetch is served


hiroshige-g:
This is a kind of meta issue to track the mismatch between:

1. Controlling ServiceWorker at the time of prefetch, which is set as prefetch record's navigation param's reserved client's active service worker.
2. Controlling ServiceWorker of the navigated Document when served by prefetch (if no fixes are made).
3. Controlling ServiceWorker of the navigated Document that would be used without prefetch.
4. Controlling ServiceWorker of the navigated Document after `claim()` when served by prefetch.

Naively I'd like to expect [2][3][4] should be the same but they can be different for several reasons:

- After prefetch before navigation, a newer ServiceWorker version is added in the ServiceWorker registration used for [1]: [1]/[2] is the old version, [3]/[4] is the new version. If we evict prefetch records on newer version (https://github.com/WICG/nav-speculation/issues/347), then this is no longer a problem.
- After prefetch before navigation, a new ServiceWorker Registration covering the prefetched URL is added: [1]/[2] is the old registration, [3]/[4] is the new registration.
- No-Vary-Search (https://github.com/WICG/nav-speculation/issues/350): [1][2][4] is the controller based on prefetch URL, [3] is the controller based on navigation URL.

I and @nhiroki discussed an option to compute the service worker controller at the time of navigation/prefetch record matching (==[3]) and compare it against [1], to avoid this kind of mismatch.
However, I found this might be harder than expected. This is probably doable (not sure in the short-term though) and I'll try another iteration later.

hiroshige-g:
After some thoughts, now I think it's fine that [1] and [3] are different: Just like non-prefetched pages' controllers aren't updated (without Clients.claim()) after navigation, prefetched pages' controllers aren't updated after prefetch. The timespan would be longer, but at least the difference between [1] and [3] is not breaking existing invariants.

So my updated expectation (== also the current Chromium implementation) is: [1] == [2], which can be different from [3].
The cases I listed above satisfies this updated expectation.

The remaining open part is whether [3] == [4] or not, but it's tracked by #350.

If this sounds good, then we can close this issue as Fixed (while #350 is still open).
WDYT?

PR #378, opened 2025-05-14
Fix request clients for prefetch + service workers

### LINKED ISSUE #346: Prefetch response cloning and reserved clients
@hiroshige-g kindly explained to me some issues with the current spec, which are becoming apparent now that Chromium is working on implementing service worker integration.

The issue stems from the fact that we keep responses in the prefetch records, and then allow reuse of them (as of https://github.com/WICG/nav-speculation/commit/5d5ebce98d7d03f389830a4388532a377c3309f0) using this line:

> Optionally, set response to a [clone](https://fetch.spec.whatwg.org/#concept-response-clone) of response.

in [create navigation params from a prefetch record](https://wicg.github.io/nav-speculation/prefetch.html#create-navigation-params-from-a-prefetch-record)

The problem is that we didn't clone all the relevant state. In particular, the "reserved client", which roughly represents the destination Window/Document the navigation request is going to be used for.

This leads to the following problematic situation. Let's say /A prefetches /B. The user navigates to /B, but document A (and its prefetch cache) stays in bfcache. The user navigates back to /A, restoring from bfcache. Then the user navigates to /B again. Assume our implementation is the sort that allows reuse in the prefetch cache, so it's reusing the same prefetch record for both navigations to /B.

Then what happens is:

- The `FetchEvent`'s `resultingClientId` for the original prefetch is `id1`
  - What happens if the web developer calls `clients.get(id1)` at this point? There's no actual `Window` represented by `id1`, so this doesn't make much sense...
- The `FetchEvent`'s `resultingClientId` for the first navigation to /B is `id1`
  - After this navigation, now `clients.get(id1)` kind of makes sense.
- The `FetchEvent`'s `resultingClientId` for the second navigation to /B is `id1`
  - But now what does `clients.get(id1)` mean? Did two separate `Window`s share a client ID?!

What we should do instead, is expose the empty string to the prefetch `FetchEvent` (similar to what is done for `<link rel=prefetch>`), and then expose separate `id1` and `id2`s for the navigation FetchEvents.

To accomplish this, we suggest the strategy that:

- The navigation param's reserved environment always be set to a clone of the request's reserved client, with the clone updated to have a new ID.
- The `resultingClientId` value (set [here](https://w3c.github.io/ServiceWorker/#create-fetch-event-and-dispatch-algorithm:~:text=If%20request%20is%20a%20non%2Dsubresource%20request%20and%20request%E2%80%99s%20destination%20is%20not%20%22report%22%2C%20initialize%20e%E2%80%99s%20resultingClientId%20attribute%20to%20reservedClient%E2%80%99s%20id%2C%20and%20to%20the%20empty%20string%20otherwise.)) is masked to the empty string for the prefetch requests, either by detecting that it's a prefetch request, or by marking the original non-clone reserved client in some special way that the service worker spec can detect.

domenic:
Fixes #346. In particular:

* Fixed prefetch activation to use the correct reserved clients. Previously, the prefetch activation parts of the specification (in "create a navigation params from a prefetch record") would assume that the request they were being passed had a reserved client, when in reality they did not, because "create a navigation request" did not set a reserved client. We change them to consult the reserved clients from the redirect chain, which is more correct anyway.

* Properly creates a new reserved environment during each prefetch activation, by creating a new environment with correct values, including the active service worker. This is necessary to avoid prefetch record reuse from causing the creation of multiple clients with the same ID.

* Makes the resultingClientId on the FetchEvent that occurs during prefetch be the empty string. The client ID is not known at the time of the prefetch, so this is the only possibility that makes sense.

@hiroshige-g PTAL!

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -363,7 +363,7 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
     1. [=list/For each=] |exchangeRecord| in |record|'s [=prefetch record/redirect chain=]:
         1. Let |redirectChainRequest| be |exchangeRecord|'s [=exchange record/request=].
         1. Let |redirectChainResponse| be |exchangeRecord|'s [=exchange record/response=].
-        1. [=list/Append=] |redirectChainResponse|'s [=response/URL=] to |urlList|.
+        1. [=list/Append=] |redirectChainRequest|'s [=request/URL=] to |urlList|.
         1. Set |responsePolicyContainer| to the result of [=creating a policy container from a fetch response=] given |redirectChainResponse| and |redirectChainRequest|'s [=request/reserved client=].
         1. Set |finalSandboxFlags| to the [=set/union=] of |targetSnapshotParams|'s [=target snapshot params/sandboxing flags=] and |responsePolicyContainer|'s [=policy container/CSP list=]'s [=CSP-derived sandboxing flags=].
         1. Set |responseOrigin| to the result of [=determining the origin=] given |redirectChainResponse|'s [=response/URL=], |finalSandboxFlags|, |documentState|'s [=document state/initiator origin=], and null.
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -748,16 +748,6 @@ This section contains patches to [[NAVIGATION-TIMING]].
 
 This section contains patches to [[SERVICE-WORKERS]].
 
-<div algorithm="Service worker control">
-  Modify the definition of <a spec="SERVICE-WORKERS">controlled</a> as follows:
-
-  <blockquote>
-    When a [=service worker client=] has a non-null [=environment/active service worker=]<ins> whose [=environment/is navigational prefetch client=] is false</ins>, it is said to be <dfn noexport>controlled</dfn> by that [=environment/active service worker=].
-  </blockquote>
-</div>
-
-<hr>
-
 <div algorithm="Create Fetch Event and Dispatch">
   Modify <a spec="SERVICE-WORKERS">Create Fetch Event and Dispatch</a>'s step which sets {{FetchEvent/resultingClientId}} as follows:
 
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -358,43 +358,33 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
     1. Let |coopEnforcementResult| be the result of [=creating a cross-origin opener policy enforcement result for navigation=] given |navigable|'s [=navigable/active document=] and |documentState|'s [=document state/initiator origin=].
     1. Let |finalSandboxFlags| be an empty [=sandboxing flag set=].
     1. Let |responsePolicyContainer| be null.
+    1. Let |activeServiceWorker| be null.
     1. Let |urlList| be an empty [=list=].
     1. [=list/For each=] |exchangeRecord| in |record|'s [=prefetch record/redirect chain=]:
-        1. Let |response| be |exchangeRecord|'s [=exchange record/response=].
-        1. [=list/Append=] |response|'s [=response/URL=] to |urlList|.
-        1. Set |responsePolicyContainer| to the result of [=creating a policy container from a fetch response=] given |response| and |request|'s [=request/reserved client=].
+        1. Let |redirectChainRequest| be |exchangeRecord|'s [=exchange record/request=].
+        1. Let |redirectChainResponse| be |exchangeRecord|'s [=exchange record/response=].
+        1. [=list/Append=] |redirectChainResponse|'s [=response/URL=] to |urlList|.
+        1. Set |responsePolicyContainer| to the result of [=creating a policy container from a fetch response=] given |redirectChainResponse| and |redirectChainRequest|'s [=request/reserved client=].
         1. Set |finalSandboxFlags| to the [=set/union=] of |targetSnapshotParams|'s [=target snapshot params/sandboxing flags=] and |responsePolicyContainer|'s [=policy container/CSP list=]'s [=CSP-derived sandboxing flags=].
-        1. Set |responseOrigin| to the result of [=determining the origin=] given |response|'s [=response/URL=], |finalSandboxFlags|, |documentState|'s [=document state/initiator origin=], and null.
+        1. Set |responseOrigin| to the result of [=determining the origin=] given |redirectChainResponse|'s [=response/URL=], |finalSandboxFlags|, |documentState|'s [=document state/initiator origin=], and null.
+        1. Set |activeServiceWorker| to |redirectChainRequest|'s [=request/reserved client=]'s [=environment/active service worker=].
         1. If |navigable| is a [=top-level traversable=], then:
-            1. Set |responseCOOP| to the result of [=obtaining a cross-origin opener policy=] given |response| and |request|'s [=request/reserved client=].
+            1. Set |responseCOOP| to the result of [=obtaining a cross-origin opener policy=] given |redirectChainResponse| and |redirectChainRequest|'s [=request/reserved client=].
             1. [=Assert=]: If |finalSandboxFlags| is not empty, then |responseCOOP|'s [=cross-origin opener policy/value=] is "`unsafe-none`".
 
                 <p class="note">This is guaranteed since the sandboxing flags of the document cannot change to become non-empty since this was prefetched, and the check was not done for a different window. If this changes, this will need to be able to handle failure of this check.</p>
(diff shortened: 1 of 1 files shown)


domenic:
I think this also fixes https://github.com/WICG/nav-speculation/issues/350.

With the new flow, we create a new reserved client [here](https://storage.googleapis.com/spec-previews/WICG/nav-speculation/pull/378/head/prefetch.html#:~:text=Set%20request%E2%80%99s%20reserved%20client%20to%20the%20result%20of%20creating%20a%20reserved%20client%20given%20navigable%2C%20request%E2%80%99s%20current%20URL%2C%20and%20null.), based on _request_'s current URL. That should be the correct activation URL (e.g., `/page?param=2` in #350). This is then carried through and cloned [shortly after](https://storage.googleapis.com/spec-previews/WICG/nav-speculation/pull/378/head/prefetch.html#prefetch-activation-client-creation) into the final reserved environment used.

domenic:
I've added more commits so that this fixes #347 by choosing solution B from https://github.com/WICG/nav-speculation/issues/347#issuecomment-2893577551 . If we want to go with solution C, we can track that in #351.

hiroshige-g:
> Fixes #347: the "clients" created during prefetch time no longer count as controlling pages, in a way that prevents service worker updates. We may still want to consider evicting prefetch records on service worker updates to tackle #351, but that can be done as an additional step on top of this.

Some early brainstorming...

Currently there is an invariant of "client's active service worker (if any) controls the client" in the spec.
Equivalently in Chromium impl, it's "`ServiceWorkerClient::controller()->controllee_map()` contains `this`", and the concept of "active service worker" might be more tightly coupled with "controlled" as its name suggests.

This proposal separates these two concepts, so I'm not sure whether this breaks (possibly implicit) assumptions/invariants in the spec and impl.
(Maybe OK in the spec, as I didn't immediately spot specific affected points? Much less sure about the Chromium impl)
If it's worth separating the two concepts (even outside the context of prefetch), then it makes more sense (but I'm also not familiar with this).

@yoshisatoyanagisawa WDYT?

prefetch.bs
```diff
@@ -539,6 +559,9 @@ Modify the [=snapshot source snapshot params=] algorithm to set the return value
             1. If <var ignore>documentResource</var> is null:
               1. Let |prefetchRecord| be the result of [=finding a matching complete prefetch record=] given <var ignore>navigable</var>, <var ignore>sourceSnapshotParams</var>, <var ignore>entry</var>'s [=session history entry/URL=], and <var ignore>sourceSnapshotParams</var>'s [=source snapshot params/sandboxing flags=].
               1. If |prefetchRecord| is not null:
+                1. Set |request|'s [=request/reserved client=] to the result of [=creating a reserved client=] given <var ignore>navigable</var>, |request|'s [=request/current URL=], and null.
```
hiroshige-g, line 562: (Probably not specific to this PR)
`navigable` here when prefetching is the navigable of the initiator Document, which is not necessarily the same navigable used when actual navigation using the prefetched result (=target navigable, which can be different from the initiator Document when `window.open` etc.). Is this OK?

prefetch.bs
```diff
@@ -378,8 +378,23 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
...
+        : [=environment/creation URL=]
+        :: |request|'s [=request/reserved client=]'s [=environment/creation URL=]
+        : [=environment/top-level creation URL=]
+        :: |request|'s [=request/reserved client=]'s [=environment/top-level creation URL=]
+        : [=environment/top-level origin=]
+        :: |request|'s [=request/reserved client=]'s [=environment/top-level origin=]
+        : [=environment/target browsing context=]
+        :: |request|'s [=request/reserved client=]'s [=environment/target browsing context=]
+        : [=environment/active service worker=]
+        :: |request|'s [=request/reserved client=]'s [=environment/active service worker=]
```
hiroshige-g, line 393: I think `request's active service worker` is never set, because the `request` doesn't go through actual fetching.
Probably we want to clone the prefetch record's last leg's request's active service worker instead?

prefetch.bs
```diff
@@ -378,8 +378,23 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
 
        <p class="note">An implementation might wish to do this if it believes that the prefetch will be consumed more than once. For example, if in the future the response is consumed by a prerender, that [=prerendering traversable=] might be [=destroy a top-level traversable|destroyed=] through various means. Normally that would mean the response is discarded, but if the implementation performs this step, then the prefetched response will still be available to serve a future navigation. [[PRERENDERING-REVAMPED]]
     1. If the user agent did not perform the previous optional step, then it must [=list/remove=] |record| from |navigable|'s [=navigable/active document=]'s [=Document/prefetch records=].
-
-
+    1. <span id="prefetch-activation-client-creation"></span>Let |reservedEnvironment| be a new [=environment=] with:
```
hiroshige-g, line 381: How about using `create a reserved client` here (and perhaps merging the creation of `request` as well) to unify the code paths?
As far as I briefly checked, `request's reserved client` is unmodified since its creation here (but I might be missing something).

prefetch.bs
domenic: I guess only the active service worker changes, so probably that's simpler...

prefetch.bs
domenic: Updated per offline discussion, PTAL!

prefetch.bs
```diff
@@ -358,28 +358,33 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
...
-            1. Set |coopEnforcementResult| to the result of [=enforcing a response's cross-origin opener policy=] given |navigable|'s [=active browsing context=], |response|'s [=response/URL=], |responseOrigin|, |responseCOOP|, |coopEnforcementResult|, and |exchangeRecord|'s [=exchange record/request=]'s [=request/referrer=].
+            1. Set |coopEnforcementResult| to the result of [=enforcing a response's cross-origin opener policy=] given |navigable|'s [=active browsing context=], |redirectChainResponse|'s [=response/URL=], |responseOrigin|, |responseCOOP|, |coopEnforcementResult|, and |exchangeRecord|'s [=exchange record/request=]'s [=request/referrer=].
     1. Set |request|'s [=request/URL list=] to |urlList|.
+    1. <span id="prefetch-activation-client-creation"></span>Set |request|'s [=request/reserved client=] to the result of [=creating a reserved client=] given |navigable| and |request|'s [=request/URL=].
```
hiroshige-g, line 378: `|request|'s [=request/URL=]` here after `Set |request|'s [=request/URL list=] to |urlList|.` just above points to the last redirect leg's URL (in this PR last redirect leg's response URL per https://github.com/WICG/nav-speculation/issues/386), and thus doesn't fix #350 -- it's just because the `|request|'s [=request/URL=]` reference timing is moved here in the recent commit.

But, if we reversing the order (i.e. doing this line BEFORE `URL list` update), we'll use the initial request URL, even it was redirected.

Maybe it's better to fix #350 separately.

prefetch.bs
```diff
@@ -358,28 +358,33 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
...
+        1. Let |redirectChainResponse| be |exchangeRecord|'s [=exchange record/response=].
+        1. [=list/Append=] |redirectChainResponse|'s [=response/URL=] to |urlList|.
+        1. Set |responsePolicyContainer| to the result of [=creating a policy container from a fetch response=] given |redirectChainResponse| and |redirectChainRequest|'s [=request/reserved client=].
         1. Set |finalSandboxFlags| to the [=set/union=] of |targetSnapshotParams|'s [=target snapshot params/sandboxing flags=] and |responsePolicyContainer|'s [=policy container/CSP list=]'s [=CSP-derived sandboxing flags=].
-        1. Set |responseOrigin| to the result of [=determining the origin=] given |response|'s [=response/URL=], |finalSandboxFlags|, |documentState|'s [=document state/initiator origin=], and null.
+        1. Set |responseOrigin| to the result of [=determining the origin=] given |redirectChainResponse|'s [=response/URL=], |finalSandboxFlags|, |documentState|'s [=document state/initiator origin=], and null.
+        1. Set |activeServiceWorker| to |redirectChainRequest|'s [=request/reserved client=]'s [=environment/active service worker=].
```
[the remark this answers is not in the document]
domenic: I like the current text of setting it each time through the loop as it treats _activeServiceWorker_ the same as _responsePolicyContainer_, _coopEnforcementResult_, and _finalSandboxFlags_.

I don't understand how this avoids #385 since it's behaviorally equivalent...

So, let me keep this as-is for now.

prefetch.bs
domenic: Ugh, you're right. That's frustrating. OK, let's fix it separately.

PR #382, opened 2025-05-23
Add Sec-Purpose header for subresources when prerendering

### LINKED ISSUE #371: Should we send Sec-Purpose on subresources or iframes in prerendered pages?
Chrome currently does this, but we do not include this in the spec.

Probably it is good for website operators, so we should spec this instead of thinking about changing Chrome.

domenic:
Closes #371.

I can't actually find any web platform tests for this...


domenic:
Note to self for Monday: in addition to writing tests, probably I should double-check the prose for the `Sec-Purpose` header definition as well to make sure it covers this.

domenic:
Tests added in https://chromium-review.googlesource.com/c/chromium/src/+/6589392

PR #387, opened 2025-06-05
Set prefetch URL and service worker correctly

### LINKED ISSUE #350: No-Vary-Search can cause reserved client creation URL mismatch

domenic:
In cases where the URL mismatches due to No-Vary-Search, we explicitly add an item to the URL list. Fixes #350.

Then, we need to ensure the active service worker is in sync with the URL of the created Document. To do this, we just manually look up the correct service worker as part of creating the reserved client.

@hiroshige-g please take a look!

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -139,6 +139,8 @@ spec: no-vary-search; urlPrefix: https://httpwg.org/http-extensions/draft-ietf-h
     text: URL search variance; url: name-data-model
     text: obtain a URL search variance; url: name-obtain-a-url-search-varianc
     text: equivalent modulo search variance; url: name-comparing
+  type: http-header
+    text: No-Vary-Search; url: name-http-header-field-definitio
 spec: resource-timing; urlPrefix: https://w3c.github.io/resource-timing/
   type: dfn; for: PerformanceResourceTiming; text: delivery type; url: dfn-delivery-type
 </pre>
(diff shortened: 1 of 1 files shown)


hiroshige-g:
LGTM, thanks!

nit: it would be better to mention in the PR title/commit
mespage to mention the "Only the No-Vary-Search of the first response in the redirect chain is used" part.

hiroshige-g:
Filed implementation bugs:

- https://issues.chromium.org/428561252 for No-Vary-Search + prefetch
- https://issues.chromium.org/428561253 for No-Vary-Search + prerender
- https://issues.chromium.org/428500224 for ServiceWorker controller + prefetch/prerender
- https://issues.chromium.org/428561255 for Document URL + prefetch/prerender

Issue #384, opened 2025-06-04
Different navigable is used in creating navigation params by fetching between prefetch and actual navigation


hiroshige-g:
In https://wicg.github.io/nav-speculation/prefetch.html#prefetch,
document’s [node navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#node-navigable) is used for calling [creating navigation params by fetching](https://wicg.github.io/nav-speculation/prefetch.html#create-navigation-params-by-fetching), i.e. the navigable of the initiator Document.
But this is not necessarily the same navigable used when actual navigation using the prefetched result (=target navigable, which can be different from the initiator Document when `window.open` etc.).

Is this OK? I'm just not sure and filing this for tracking.

domenic:
Let's see. What does this impact?

- Setting request's reserved client. Does this impact any fields of the environment struct in a bad way?
  - creation URL, top-level creation URL, and top-level origin should be fine
  - target browsing context is set incorrectly. Does this matter?
    - [Lots of specs](https://dontcallmedom.github.io/webdex/t.html#target%20browsing%20context%40%40environment%40dfn) use target browsing context. I am a bit suspicious as to whether they're doing so correctly.
    - The Service Worker spec usages should be OK since these clients are not execution-ready and so get filtered out
    - Some specs seem to check whether the target browsing context has a parent or not. So as long as we don't allow prefetches to be used to fulfill iframes then we're OK
    - Some other specs look like they're misusing it
    - Some specs seem to be checking if it's null
    - Conclusion: probably OK?
- Aborting prefetches if a navigation occurs in the initiator
  - This seems fine, although as noted in the inline issue maybe we should wait for navigations to commit instead
- COOP checks on top-level traversables
  - We re-do these at activation time, so we are mostly OK
  - But I think this means we are checking both the current navigable and the target navigable. Which means we might disallow some cases that could succeed.
- CORP checks on child navigables
  - Not applicable as long as we stick with only prefetching top-level
- Conflicting credentials exist checks
  - This might in theory also disallow some cases that could succeed, but I think it's probably OK if we look into the details because it's using the actual redirect chain partitions
- Storing it in the navigation params
  - I think it is not used after being stored, since we don't do anything after "create navigation params by fetching"


Overall conclusion:
- As long as we only allow prefetches for top-level traversables, it is probably OK.
- Maybe we are disallowing some cases that could succeed by checking both the initiator and the destination navigables, whereas strictly speaking we only need to check the destination navigable.
[...]

PR #394, opened 2025-06-25
Delay all COOP checks to activation

domenic:
They have been broken since 24b2405eeb9190d6604951ec84b4f592f3c3a358, where we allowed prefetches to be activated into different browsing contexts than their initiator browsing context. Attempting to do COOP checks at prefetch time is no good with that foundation.

This better aligns with Chromium's implementation, and simplifies the specification a good deal.

This helps with #384, although we should still be sure to add a clarifying note about the mismatch and how, after this fix, it's no longer a problem.


domenic:
@hiroshige-g, I don't know if you have bandwidth or interest to review this, since you are not as familiar with the COOP implementation. But I will give it a few days before merging in case you are interested.

PR #398, opened 2025-07-11
Rebase on top of HTML

domenic:
https://github.com/whatwg/html/pull/11426 upstreams most of the speculation rules specification to HTML. Remove the upstreamed portions, and add references to the PR and PR preview.


domenic:
I'll merge this. Even though the HTML spec PR is not yet merged, it has a few changes that are important, and is a better basis for work going forward. We can use the PR preview at https://whatpr.org/html/11426/speculative-loading.html as the rendered spec until it gets merged.

PR #403, opened 2025-08-05
Add Automation for Prefetch Status Updated 

srwei:


srwei:
@domenic 
I think the CDDL related portions should be good as we've been able to get it working locally, but the rest of the content is very much WIP. Would love to have your feedback and help.
One issue I ran into is basing the PreloadingStatuses off the prefetchRecord's state, but it's not really 1 to 1 here. Would it make more sense to just match what's in the spec for the Bidi module type and add the success status for activated prefetches?

prefetch.bs
```diff
@@ -963,6 +967,81 @@ The following parameters are defined for the `prefetch` token:
...
+   initiatingFrameId: text,  
+   url: text,
+   status: speculation.PreloadingStatus
+}
+</pre>
+
+<div algorithm="remote end event trigger for speculation.prefetchStatusUpdated">
+To <dfn>trigger a prefetch status updated event</dfn> given a [=navigable=] |navigable|, a string |url|, and a |prefetchRecord|:
+
+1. Let <var ignore>initiatingFrameId</var> be the [=top-level traversable=] |navigable|'s [=navigable id=].
```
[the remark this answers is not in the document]
srwei: Thanks, made me realize I didn't map the variables to the event params. Was meant to use the frameId defined here to set the initiatingFrameId in the event properties.

prefetch.bs
```diff
@@ -963,6 +967,81 @@ The following parameters are defined for the `prefetch` token:
...
+1. Let |url| be the |prefetchRecord|'s [=prefetch record/URL=].
```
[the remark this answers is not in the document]
srwei: Agreed.

prefetch.bs
```diff
@@ -963,6 +967,81 @@ The following parameters are defined for the `prefetch` token:
...
+1. Let |status| be null.
+1. Given a |prefetchRecord|'s [=prefetch record/state=]:
```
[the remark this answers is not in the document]
srwei: Inline all of the if/else? Worried if that might look a bit confusing.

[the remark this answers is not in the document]
domenic: It looks like this is still in the process of being fixed. Now that you are passing in the parameter, you don't need any of these steps.

prefetch.bs
```diff
@@ -290,12 +296,12 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
     1. Let |recordToUse| be |exactRecord| if |exactRecord| is not null, otherwise |inexactRecord|.
     1. If |recordToUse| is not null:
         1. Let |currentTime| be the [=current high resolution time=] for |navigable|'s [=navigable/active document=].
-        1. If |recordToUse|'s [=prefetch record/expiry time=] is less than |currentTime|, return null.
+        1. If |recordToUse|'s [=prefetch record/expiry time=] is less than |currentTime|, return null and [=trigger a prefetch status updated event=] given |navigable|, |recordToUse|, and "{^speculation.PreloadingStatus/failure^}" status
```
domenic, line 299: It's generally confusing to do "return null and do X", because after returning no more steps can happen. I'd suggest nested lists instead, i.e.

```
1. If ...:
    1. Trigger a prefetch status updated event given ...
    2. Return null.

(Here and below.)

File: prefetch.bs [hunk 5]
```diff
@@ -290,12 +296,12 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
...
         1. [=list/For each=] |exchangeRecord| of |recordToUse|'s [=prefetch record/redirect chain=]:
-            1. If [=conflicting credentials exist=] for |exchangeRecord|'s [=exchange record/response=] given |navigable| and |recordToUse|'s [=prefetch record/source partition key=], return null.
+            1. If [=conflicting credentials exist=] for |exchangeRecord|'s [=exchange record/response=] given |navigable| and |recordToUse|'s [=prefetch record/source partition key=], return null and [=trigger a prefetch status updated event=] given |navigable|, |recordToUse|, and "{^speculation.PreloadingStatus/failure^}" status
 
             <div class="note">This handles the case where there were no cross-partition credentials initially, but there are now. User agents could use a slightly coarser algorithm, such as monitoring whether the cookies for the URL have been modified at all, or storing a hash, timestamp or revision number.</div>
-
+        1. [=Trigger a prefetch status updated event=] given |navigable|, |recordToUse|, and "{^speculation.PreloadingStatus/success^}" status.
```
domenic, line 304: I suspect this is too early for success. Probably more around https://storage.googleapis.com/spec-previews/WICG/nav-speculation/pull/403/head/prefetch.html#ref-for-wait-for-a-matching-prefetch-record%E2%91%A0:~:text=Set%20prefetched%20to%20true. ?

prefetch.bs
```diff
@@ -963,6 +970,80 @@ The following parameters are defined for the `prefetch` token:
...
+{^Remote end definition^} and {^local end definition^}
+
+<pre class="cddl" data-cddl-module="remote-cddl,local-cddl">
+speculation.PreloadingStatus = "pending" / "ready" / "success" /
+                   "failure" / "not-supported"
+</pre>
+
+The <code>speculation.PreloadingStatus</code> type represents the different types of preloading statuses. This status is shared by prefetches and prerenders.
+
+"<dfn cddl-value>pending</dfn>" status is defined when a [=prefetch record=] |prefetchRecord| begins to [=start a referrer-initiated navigational prefetch=].
```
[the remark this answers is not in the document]
srwei: Yeah, that makes sense. I've kept the actual param as speculation.PreloadingStatus as the plan is to reuse for prerendering in the future.

prefetch.bs
```diff
@@ -161,6 +161,10 @@ spec: no-vary-search; urlPrefix: https://httpwg.org/http-extensions/draft-ietf-h
     text: No-Vary-Search; url: name-http-header-field-definitio
 spec: resource-timing; urlPrefix: https://w3c.github.io/resource-timing/
   type: dfn; for: PerformanceResourceTiming; text: delivery type; url: dfn-delivery-type
+spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
+  type: dfn
+    text: error; url: dfn-errors
+    text: local end; url: dfn-local-ends
```
domenic, line 167: These two lines should not be necessary; can you try deleting them and see what happens? I guess "error" is pretty generic, but disambiguation should be done via link-defaults not anchors.

prefetch.bs
```diff
@@ -290,12 +296,16 @@ The user agent may [=prefetch record/cancel and discard=] records from the [=Doc
...
+            1. Return null.
         1. [=list/For each=] |exchangeRecord| of |recordToUse|'s [=prefetch record/redirect chain=]:
-            1. If [=conflicting credentials exist=] for |exchangeRecord|'s [=exchange record/response=] given |navigable| and |recordToUse|'s [=prefetch record/source partition key=], return null.
+            1. If [=conflicting credentials exist=] for |exchangeRecord|'s [=exchange record/response=] given |navigable| and |recordToUse|'s [=prefetch record/source partition key=]:
+                1. [=Trigger a prefetch status updated event=] given |navigable|, |recordToUse|, and "{^speculation.PreloadingStatus/failure^}" status.
+                1. Return null.
 
             <div class="note">This handles the case where there were no cross-partition credentials initially, but there are now. User agents could use a slightly coarser algorithm, such as monitoring whether the cookies for the URL have been modified at all, or storing a hash, timestamp or revision number.</div>
-
+
```
domenic, line 308: Please remove trailing whitespace :)

prefetch.bs
```diff
@@ -856,6 +867,7 @@ The <dfn>list of sufficiently strict speculative navigation referrer policies</d
         1. If |prefetchRecord|'s [=prefetch record/had conflicting credentials=] is true, then set |navigationParams| to null.
 
             <div class="note">This means that if any cross-partition origin along the redirect chain had credentials (and did not override this behavior using [:Supports-Loading-Mode:]), the prefetch is discarded. This reduces the chance of the user observing a logged-out page when they are logged in.</div>
+        1. [=Trigger a prefetch status updated event=] given |document|'s [=node navigable=], |prefetchRecord|, and "{^speculation.PreloadingStatus/pending^}" status.
```
domenic, line 870: This seems a bit too late to me, as it waits until after the fetch.

Probably we want to do this before creating the navigation params by fetching?

prefetch.bs
```diff
@@ -963,6 +967,81 @@ The following parameters are defined for the `prefetch` token:
     This specification conforms to this advice; the [=start a referrer-initiated navigational prefetch=] algorithm does not emit non-boolean values.
   </div>
 
+<h2 id="automated-testing">Automated testing</h2>
+
+For the purposes of user-agent automation and application testing, this document defines extensions to the [[WebDriver-BiDi]] specification.
+
+<div class="issue">CDDL snippets use the "text" type instead of "browsingContext.BrowsingContext" to allow indepedent programmatic processing of CDDL snippets. Currently, other modules cannot be referenced.</div>
```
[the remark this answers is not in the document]
domenic: Thanks! I filed https://github.com/w3c/webdriver-bidi/issues/994 to see if we can figure out why it's not exported, but, for now there's no action.

prefetch.bs
```diff
@@ -963,6 +975,74 @@ The following parameters are defined for the `prefetch` token:
...
+
+"<dfn for="speculation.PreloadingStatus" cddl-value>pending</dfn>" status is defined when a [=prefetch record=] |prefetchRecord| begins to [=start a referrer-initiated navigational prefetch=].
+
+"<dfn for="speculation.PreloadingStatus" cddl-value>ready</dfn>" status is defined when a [=prefetch record=] |prefetchRecord| is [=prefetch record/complete=].
+
+"<dfn for="speculation.PreloadingStatus" cddl-value>success</dfn>" status is defined when a [=prefetch record=] |prefetchRecord| is activated, such as from the user agent's navigation.
+
+"<dfn for="speculation.PreloadingStatus" cddl-value>failure</dfn>" status is defined when a [=prefetch record=] |prefetchRecord| has failed, such as when the |prefetchRecord| expires.
+
+"<dfn for="speculation.PreloadingStatus" cddl-value>not-supported</dfn>" status is defined when a [=prefetch record=] |prefetchRecord| is ineligble to [=start a referrer-initiated navigational prefetch=].
```
domenic, line 1007: Nothing in this spec ever sets this status. Is that still work to be done?

prefetch.bs
```diff
@@ -963,6 +975,74 @@ The following parameters are defined for the `prefetch` token:
...
+
+speculation.PrefetchStatusUpdatedParameters = {
+   context: text,  
+   url: text,
+   status: speculation.PreloadingStatus
+}
+</pre>
+
+<div algorithm="remote end event trigger for speculation.prefetchStatusUpdated">
+To <dfn>trigger a prefetch status updated event</dfn> given a [=navigable=] |navigable|, |prefetchRecord|, and a {^speculation.PreloadingStatus^} |preloadingStatus|:
```
domenic, line 1033:
```suggestion
To <dfn>trigger a prefetch status updated event</dfn> given a [=navigable=] |navigable|, a [=prefetch record=] |prefetchRecord|, and a {^speculation.PreloadingStatus^} |preloadingStatus|:
```

prefetch.bs
```diff
@@ -963,6 +975,74 @@ The following parameters are defined for the `prefetch` token:
...
+
+1. Let |frameId| be |navigable|'s [=navigable/top-level traversable=]'s [=navigable id=].
+1. Let |url| be the |prefetchRecord|'s [=prefetch record/URL=].
+1. Let |status| be the |preloadingStatus|.
```
domenic, line 1037: Aliasing |preloadingStatus| to |status| doesn't seem useful.

Generally since |frameId| and |url| are only used once, I'm not sure declaring them as variables is that helpful. (Especially |url|.)

prefetch.bs
srwei: I'm unable to delete the two text fields because it errors with all anchors needing exactly one text. But I think you're right in that we don't need anchors as the link-defaults currently work for references, like in "Emit an event".

prefetch.bs
srwei: I agree. I've moved it to after we start the time to expire timer, which is about where it is in our implementation as well.

prefetch.bs
srwei: Looking through the implementation, I don't see any places currently in the spec where this would match. It's mainly related to prerendering logic, like upgrading to prerender or having a duplicate preload attempt. I'm thinking we could remove this value from the spec, how does that sound?

domenic: Yes, sounds good, let's remove it!

domenic (approved):
LGTM!

Should we merge this, or do you want to give it a bit more time to ensure that the implementation is settled and aligned with the spec first?

whimboo (commented):
I have some feedback for CDDL specific additions in this PR. Maybe you can take a look at those and check if it would be fine to adjust in this spec. Thanks.

prefetch.bs
```diff
@@ -963,6 +972,69 @@ The following parameters are defined for the `prefetch` token:
...
+
+"<dfn for="speculation.PreloadingStatus" cddl-value>ready</dfn>" status is defined when a [=prefetch record=] |prefetchRecord| is [=prefetch record/complete=].
+
+"<dfn for="speculation.PreloadingStatus" cddl-value>success</dfn>" status is defined when a [=prefetch record=] |prefetchRecord| is activated, such as from the user agent's navigation.
+
+"<dfn for="speculation.PreloadingStatus" cddl-value>failure</dfn>" status is defined when a [=prefetch record=] |prefetchRecord| has failed, such as when the |prefetchRecord| expires.
+
+<h3 id="speculation-module-events">Events</h3>
+
+<pre highlight="cddl" class="cddl" data-cddl-module="remote-cddl,local-cddl">
```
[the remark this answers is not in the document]
srwei: Yup! let me know how this looks. Shortened the definitions a bit for better readability.

prefetch.bs
```diff
@@ -963,6 +972,69 @@ The following parameters are defined for the `prefetch` token:
...
+
+<div class="issue">CDDL snippets use the "text" type instead of "browsingContext.BrowsingContext" to allow indepedent programmatic processing of CDDL snippets. Currently, other modules cannot be referenced.</div>
+
+<h3 id="speculation-module">The speculation module</h3>
+
+The speculation module contains commands for managing the remote end behavior for prefetches, prerenders, and speculation rules.
+
+<h3 id="speculation-module-definitions">Definitions</h3>
+
+{^Remote end definition^} and {^local end definition^}
```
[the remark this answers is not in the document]
srwei: Yeah you're right, changed all data-cddl-modules to only include local definitions.

prefetch.bs
```diff
@@ -963,6 +972,69 @@ The following parameters are defined for the `prefetch` token:
...
 
+<h2 id="automated-testing">Automated testing</h2>
+
+For the purposes of user-agent automation and application testing, this document defines extensions to the [[WebDriver-BiDi]] specification.
+
+<div class="issue">CDDL snippets use the "text" type instead of "browsingContext.BrowsingContext" to allow indepedent programmatic processing of CDDL snippets. Currently, other modules cannot be referenced.</div>
+
+<h3 id="speculation-module">The speculation module</h3>
+
+The speculation module contains commands for managing the remote end behavior for prefetches, prerenders, and speculation rules.
```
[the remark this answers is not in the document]
srwei: Yup, will do.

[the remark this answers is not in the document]
whimboo: Just a note that this still needs to be done. Otherwise this PR looks fine to me. Thanks!

PR #405, opened 2025-09-02
Introduce Supports-Loading-Mode:prerender-cross-origin-frames

yoichio:
Explainer: https://github.com/WICG/nav-speculation/blob/main/prerendering-cross-origin-iframes.md


yoichio:
L544 |initiatorOrigin| is undefined in [=create and initialize a Document object=], how to get it?

domenic:
Did you mean to close this?

I don't think you need an initiatorOrigin check. You should still look at the `Supports-Loading-Mode` value even if page A and page B are same-origin.

PR #409, opened 2025-09-11
Do not keep service workers alive via prefetch clients

### LINKED ISSUE #347: Prefetch requests keeping the service worker alive and preventing it from revving its version

domenic:
Fixes #347: the "clients" created during prefetch time no longer count as controlling pages, in a way that prevents service worker updates. We may still want to consider evicting prefetch records on service worker updates (as discussed in #351), but that can be done as an additional step on top of this.

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -121,6 +121,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
         text: anonymization policy; url: sl-candidate-anonymization-policy
       for: speculative load candidate
         text: URL; url: sl-candidate-url
+    urlPrefix: webappapis.html
+      text: navigation and traversal task source
 spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
   type: dfn
     text: network partition key; url: network-partition-key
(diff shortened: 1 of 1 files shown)


hiroshige-g:
Given that https://github.com/WICG/nav-speculation/issues/347#issuecomment-3294335361 https://github.com/WICG/nav-speculation/issues/347#issuecomment-3294367358, I feel I'm not yet fully confident that this is implementable right now so prefer to keep this unresolved (until we get bandwidth/priority to investigate/implement this).

Also @yoshisatoyanagisawa, WDYT?

PR #412, opened 2025-09-22
add initial version for prerender-until-script

Clqsin45:


prerendering.bs
```diff
@@ -425,6 +430,13 @@ Every [=navigable=] has a <dfn for="navigable">loading mode</dfn>, which is one
 
 By default, a [=navigable=]'s [=navigable/loading mode=] is "`default`". A navigable whose [=navigable/loading mode=] is "`prerender`" is known as a <dfn>prerendering navigable</dfn>. A [=prerendering navigable=] that is also a [=top-level traversable=] is known as a <dfn>prerendering traversable</dfn>.
 
+Every [=navigable=] has a <dfn for="navigable">scripting mode</dfn>, which is one of the following:
```
hiroshige-g, line 433: My naive question would be: Is this flag associated with `navigable` or something else (e.g. `Document`)?

So, clarifying the context/reasoning for the placement of this flag (no matter of the specific placement) is helpful.

prerendering.bs
```diff
@@ -950,6 +970,12 @@ To <dfn export>get the supported loading modes</dfn> for a [=response=] |respons
 
 Various behaviors are disallowed in [=prerendering navigables=] because they would be intrusive to the user, since the prerendered content is not being actively interacted with.
 
+<h3 id="pausing-script-execution">Pausing script execution</h3>
+
+If a [=navigable=]'s [=navigable/scripting mode=] is "`paused-until-activation`", the user agent must not execute script in the navigable's [=navigable/active document=]. This includes, but is not limited to executing `<script>` elements.
```
hiroshige-g, line 975: This PR is more like the speculation-rules-side plumbing and doesn't define the behavior of "pausing script execution".

So, at least we should clearly label the to-be-defined concepts and centralize the undefined things (probably to the section).
Namely,  "Resume script execution" at Line 633.

prerendering.bs
```diff
@@ -974,7 +1000,11 @@ Modify the <a spec=HTML>download the hyperlink</a> algorithm to ensure that down
 
 <h3 id="delay-async-apis">Delaying async API results</h3>
 
-Many specifications need to be patched so that, if a given algorithm invoked in a [=prerendering navigable=], most of its work is deferred until the navigable's [=navigable/top-level traversable=] is [=prerendering traversable/activated=]. This is tricky to do uniformly, as many of these specifications do not have great hygeine around <a href="https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-for-spec-authors">using the event loop</a>. Nevertheless, the following sections give our best attempt.
+Many specifications need to be patched so that, if a given algorithm invoked in a [=prerendering navigable=], most of its work is deferred until the navigable's [=navigable/top-level traversable=] is [=prerendering traversable/activated=].
+
+<p class="note">Note that if the [=navigable/scripting mode=] is "`paused-until-activation`", the scripts that would invoke these algorithms are not executed in the first place until activation.</p>
```
hiroshige-g, line 1005: nit: I feel this note should be better placed after Line 1007, because I was confused as if "This is tricky to do uniformly, ... hygeine around using the event loop" would refer to the paused-until-activation.
(The document structure `<p class=note>` should help disambiguate this, but this confused me because the same thing also applies to paused-until-activation in some sense)

domenic (commented):
Nice work, this seems like a great start! I am also happy to see @hiroshige-g helping out!!

prerendering.bs
Clqsin45: Thank you for the discussion! I added a short note trying to cover the points. PTAL :)

prerendering.bs
```diff
@@ -282,43 +284,14 @@ A <dfn>prerender candidate</dfn> is a [=speculative load candidate=] with the fo
...
-            to |prerenderCandidates|.
-        1. If |rule|'s [=speculation rule/predicate=] is not null, then:
-          1. Let |links| be the result of [=finding matching links=] given |document| and |rule|'s [=speculation rule/predicate=].
-          1. [=list/For each=] |link| of |links|:
-            1. Let |target| be |rule|'s [=speculation rule/target navigable name hint=].
-            1. If |target| is null, set it to the result of [=getting an element's target=] given |link|.
-            1. Let |referrerPolicy| be the result of [=computing a speculative load referrer policy=] given |rule| and |link|.
-            1. [=list/Append=] a [=prerender candidate=] with
-
+    1. [=list/For each=] (|rules|, |scriptsPaused|) of « (|ruleSet|'s [=speculation rule set/prerender rules=], false), (|ruleSet|'s [=speculation rule set/prerender_until_script rules=], true) »:
```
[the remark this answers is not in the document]
Clqsin45: :D

prerendering.bs
```diff
@@ -269,9 +270,10 @@ Extend the [=speculation rule=] [=struct=] with one additional [=struct/item=]:
 
 <h3 id="speculation-rules-processing">Processing model</h3>
 
-A <dfn>prerender candidate</dfn> is a [=speculative load candidate=] with the following additional [=struct/item=]:
+A <dfn>prerender candidate</dfn> is a [=speculative load candidate=] with the following additional [=struct/items=]:
 
 * <dfn for="prerender candidate">target navigable name hint</dfn>, a [=valid navigable target name or keyword=] or null
+* <dfn for="prerender candidate">scripts paused</dfn>, a [=boolean=], initially false
```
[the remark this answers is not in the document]
Clqsin45: Ah exactly! Revised..

I also change the boolean's name to shouldPauseScripts, but wondering pauseScripts sounds better or not..

Clqsin45:
Thank you for your comments!! ❤️ Trying to resolve as many as I can, and I will make a follow-up PR soon.

prerendering.bs
```diff
@@ -974,7 +1008,11 @@ Modify the <a spec=HTML>download the hyperlink</a> algorithm to ensure that down
 
 <h3 id="delay-async-apis">Delaying async API results</h3>
 
-Many specifications need to be patched so that, if a given algorithm invoked in a [=prerendering navigable=], most of its work is deferred until the navigable's [=navigable/top-level traversable=] is [=prerendering traversable/activated=]. This is tricky to do uniformly, as many of these specifications do not have great hygeine around <a href="https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-for-spec-authors">using the event loop</a>. Nevertheless, the following sections give our best attempt.
+Many specifications need to be patched so that, if a given algorithm invoked in a [=prerendering navigable=], most of its work is deferred until the navigable's [=navigable/top-level traversable=] is [=prerendering traversable/activated=].
+
+This is tricky to do uniformly, as many of these specifications do not have great hygeine around <a href="https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-for-spec-authors">using the event loop</a>. Nevertheless, the following sections give our best attempt.
```
domenic, line 1013: Nit: you could re-collapse these two sentences back into one paragraph.

domenic (approved):

prerendering.bs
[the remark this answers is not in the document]
domenic: I can see arguments for both. "should pause scripts" sounds a bit clearer that it's about a future behavior, but it's a bit confusing since "should" has special meaning in specs that doesn't quite match what we're using here...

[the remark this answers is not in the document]
hiroshige-g: nit: I basically prefer "block scripts" / "unblock scripts" over "pause/resume scripts":

- To align with https://html.spec.whatwg.org/multipage/#has-no-style-sheet-that-is-blocking-scripts
- To avoid confusion: we don't pause the parser directly, we don't pause scripts in the middle of a script execution.

But this naming will be also affected by the detailed spec defintion we'll make later, so not blocking.

prerendering.bs
Clqsin45: Thank you for the information!! So, I think we're going to discuss the blocking/unblocking implementation during TPAC. I will draft changes to scripting first and finalize it after TPAC.

prerendering.bs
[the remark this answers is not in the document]
Clqsin45: Thank you all! So then maybe let me follow the naming style of https://html.spec.whatwg.org/multipage/#has-no-style-sheet-that-is-blocking-scripts, and update it after the scripting spec is ready.

How do the current names look to you?

prerendering.bs
Clqsin45: Done.

prerendering.bs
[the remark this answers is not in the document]
domenic: LGTM!

Issue #305, opened 2024-03-25
Add an additional action - preparse (aka NSP)


tvereecke:
We propose to add a third action called “preparse” (open for alternative names), an action similar to No State Prefetch (NSP), but triggered via the Speculation Rules API.

This new third action allows for more performance gains than a simple prefetch and would be a significant risk reduction compared to a full prerender (due to 1st party and 3rd party JS execution)

We have found that some of our customers are reluctant to enable Speculation Rules because the risk to their application of enabling pre-rendering is too great

Here are 3 (related) statements:

> "Prerender is too risky because we don't know how it will affect analytics or features like product recommendations or recently viewed products"
> "For Prerender to work, the impact on the app needs to be fully understood and many different teams need to work together"
> "If we do it wrong, it can break our application. (e.g. analytics, personalization, price controls, targeting, ...)"

Existing alternatives with 2 different actions are suboptimal:
* Falling back to Prefetch is unlikely to deliver sufficient performance gains.
*  Using Eagerness=conservative reduces the risk significantly, but due to the small time window the performance benefits are usually limited.

This new action also allows for progressive speculation rules, where the result of an eagerly triggered action (a lot of time) can be used for the less eager action.
* Preparse with eagerness "moderate" and Prerender with eagerness "conservative."
* Prefetch with eagerness "moderate" and Preparse with eagerness "conservative".
* Preparse with eagerness "eager" and only Prerender with eagerness "conservative".
* ...

Thank you for considering this option

jeremyroman:
(In fact Chromium actually still has some code to support `"prefetch_with_subresources"` which invokes NSP.)

Lots of interesting questions to evaluate how useful this is (e.g., how large are the subresources that are not already in cache, assuming we're doing this same-origin; would we want to support this cross-origin or not).

On the spec side, probably the most interesting piece here would be formalizing more about how preloaded subresources are found (somewhere there is discussion on a GitHub issue about invoking the HTML parser in a way that will scan through `<noscript>` etc properly, in the context of `<meta>`, which I can dig up). It's simplest conceptually if everything is in `Link` header fields, but would that be enough?

krzksz:
Huge +1 to this one as we're facing similar challenges on our platform when it comes to Speculation Rules. We want to roll them out to everyone in a safe and impactful way. Unfortunately, the less conservative we get the higher the risk that we'll break merchants' scripts and analytics.

[...]

PR #416, opened 2025-09-24
add first version of prerender-until-script explainer

Clqsin45:


prerender-until-script.md
```diff
@@ -0,0 +1,123 @@
...
+  ]
+}
+</script>
+
+<a href="/heavy-script-page.html">Link to a page with heavy scripts</a>
+```
+
+```html
+<!-- heavy-script-page.html -->
+<script defer scr="deferred-script.js" id="deferred-script"></script>
```
domenic, line 57:
```suggestion
<script defer src="deferred-script.js" id="deferred-script"></script>
```

prerender-until-script.md
```diff
@@ -0,0 +1,123 @@
...
+  // Heavy script... 
+}
+</script>
+<img src="image2.jpg"/>
+
+```
+
+In this example, `/heavy-script-page.html` will be prerendered, but its scripts will not run until the user clicks the link and the page is activated.
+
+More specifically, the user agent processes the page as follows:
```
domenic, line 71: This is very good and helpful.

prerender-until-script.md
```diff
@@ -0,0 +1,123 @@
...
+
+1.  The `deferred-script` script is parsed and added to the queue of scripts that will be executed after page parsing is complete. Parsing then resumes.
+2.  The `async-script` script is parsed and added to the queue of scripts that will be executed after prerender activation. Parsing then resumes.
+3.  The `first-image` element is parsed, `image1.jpg` is fetched and loaded, and parsing resumes.
+4.  When the `first-blocking` script is parsed, the parser pauses and waits for prerender activation.
+5.  While waiting for activation, the user agent may optionally prefetch `image2.jpg`.
+6.  The prerendered document is activated.
+7.  The parsing of `first-blocking` script resumes, and the script is executed.
```
domenic, line 79:
```suggestion
7.  The parsing of `first-blocking` script resumes, and the `first-blocking` script is executed.
```

domenic, line 79: Now continue the story and tell me about when and in what order `deferred-script` and `async-script` are executed!

prerender-until-script.md
```diff
@@ -0,0 +1,123 @@
...
+ 
+## Detailed Design Discussion
```
domenic, line 81: This is the section where I think you should discuss some of the complexities around what pausing means, inline event handlers, etc.

In particular I was interested to see your strong claim

> The JavaScript engine is effectively "frozen" until the document is activated. Upon activation, script execution resumes as if there was no interruption.

If that's our plan, that is great, as it means inline event handlers are also probably solved! But I thought maybe we were only deferring `<script>`s? Contrasting these different approaches would be good.

prerender-until-script.md
```diff
@@ -0,0 +1,123 @@
...
+
+Another alternative is for developers to modify their pages to be more "prerender-friendly," for example, by delaying script execution manually using the `prerenderingchange` event. While this is a valid approach, `prerender_until_script` provides a simpler, declarative way to achieve a similar result without requiring significant code changes. It also provides a stronger guarantee that *no* scripts will be executed, which can be difficult to achieve manually.
+
+### Alternative Naming
+
+The initial proposal for this feature used the name `preparse`. Other names, such as `prescan` and `prefetch_with_subresources`, were also considered. The name `prerender_until_script` was chosen to make it clear that this is a form of prerendering, but with a specific modification to the script execution lifecycle. This naming helps to place the feature within the existing mental model of the Speculation Rules API.
+
+## Stakeholder Feedback
+
+*This section is a placeholder for feedback from browser vendors, web developers, and other stakeholders.*
```
domenic, line 117: Move the link to #305 into this section and briefly that we have at least two developers who are very positive on this.

Maybe also just have a general mention on how many times we've heard about people being unable to go beyond prefetch, all the way to prerender.

domenic (commented):
Really good!

prerender-until-script.md
Clqsin45: Ah right, actually we only pauses script elements, and I should be careful about it.

I changed it to:
1. For sure, we're going to defer script element execution. 
2. We're seeking for community feedback on inline event handlers. 

How does it sound to you?

prerender-until-script.md
Clqsin45: oops, thank you for catching it.

prerender-until-script.md
Clqsin45: Added! :D

prerender-until-script.md
Clqsin45: Thank you!

prerender-until-script.md
domenic: It looks quite nice now. However one detailed point on making it clearer: it seems helpful to distinguish between "entry point" script executions and other script execution. E.g. you say

> However, web pages execute JavaScript in many ways beyond the `<script>` tag. This includes inline event handlers (e.g., `onclick`), timers (`setTimeout`), and promise resolutions.

and then later mention that inline event handlers are particularly important, but do not go into detail on why.

The way I would explain this would be to discuss something about how, if we are able to perfectly cut off all scripting entry points, then that is equivalent to cutting off all scripts. And, the two script entry points that we are aware of at the current time are `<script>` elements and inline event handlers.

If either of those two entry points execute script, then the scripts they execute might spawn other scripts, e.g. via timers and promises and adding non-inline event handlers.

But if we cut off both of those entry points, then I believe we do not need to worry about timers etc., since there will be no way for the web developer to start timers or attach promise handlers or whatever.

(It's possible there are more entry points we're missing, but I cannot immediately think of any...)

domenic (approved):
LGTM although there is room for a bit more refinement and I'm happy to do another review pass.

prerender-until-script.md
Clqsin45: Thank you!! 
Actually, I was under the impression that I had seen a setTimeout attribute.... But it seems I was wrong.

I rewrote the this part, including:
- removed setTimeout from the consideration.
- listed the entry points for further discussion (thanks so much for this!).
- added a warning that using inline event handlers is not a good practice.

How does this look to you?

domenic: This looks really excellent! Let me merge it :)

Issue #322, opened 2024-06-11
Way of allowing simple form-based navigations for Search


tunetheweb:
Currently speculative navigation only works for links. This seems sensible as we only want to speculate safe, idempotent results.

However Search forms on sites are incredibly common and is usually a safe GET request too. They are also often slower pages (especially for client-rendered search results) that would benefit from speculating in advance. However at present Chrome does not see these as navigations that can reuse prefetch nor prerendered speculations.

[Some have had success with prerendering](https://x.com/TimVereecke/status/1800100536647311579) search results, even though the prerendering search results page will not be used, due to the fact that the page (and it's assets like thumbnails) are then available in the HTTP cache for the subsequent form-based navigation.

While this pre-warming the cache does create performance benefits, it is not as good as allowing the the navigation to use the speculation directly. One reason for that is this benefit depends on the page being cacheable and sadly many are not. One of the benefits of the Speculation Rules API over `<link rel=prefetch>` is it does work for non-cacheable documents providing the speculation is activated (which, as I say, it is not with a form-based GET navigation).

Others are suggesting [other ways of hacking around this](https://github.com/WordPress/performance/issues/1291) by intercepting the form-based navigation and instead using JavaScript to speculate and then navigate, which does allow the speculation to be used, but at the cost of complexity and it feels a bit icky to have to do this.

Should we support this navigately? Maybe with a `<form>` attribute that can be added to the markup? For example:
```html
<form allow-speculation="true" method="get" action="/search">
  ...
</form>
```

This should be restricted to `get` methods IMHO.

jeremyroman:
If the form navigation itself has to be stopped, I don't see why that wouldn't work already -- we should still be able to serve a prefetched/prerendered response to a GET form submission. If we can't, that's probably straightforward to address. (Document rules won't pick it up, which I suppose we could if this were common, but that aspect doesn't seem high priority.)

Experimentally this [seems to work fine](https://sour-ethereal-punishment.glitch.me/) in Chrome for prefetch, so maybe this is just a prerendering bug? The given failure reason is "The prerender was not used because during activation time, different navigation parameters (e.g., HTTP headers) were calculated than during the original prerendering navigation request."

If you agree, maybe that can just turn into a Chromium bug against Internals>Preload>Prerender?

jeremyroman:
For clarity, I'm still happy to keep the idea of document rules handling this in the future around somewhere, of course!

tunetheweb:
[...]

PR #424, opened 2026-01-23
Add more details which `form_submission` is needed

robertlin-chromium:
Add more details to the reason why this field only applies to prerender.

### DIFF
--- a/prerendering-form-submission.md
+++ b/prerendering-form-submission.md
@@ -13,7 +13,7 @@ It should be noted the speculation will need to be triggered by the page in some
 ### More details
 
 Note that this field is only required for prerender. On the other hand, prefetch speculation rules can be used by form navigations without specifying this field.
-The key difference is that unlike prefetch, which only downloads HTML resources, prerendering actually starts the navigation. Because parameters are involved, using a standard (non-form) prerender for a form submission will result in breakage.
+The key difference is that unlike prefetch, which only downloads HTML resources, prerendering actually starts the navigation. Because internal navigation parameters in a browser are predetermined, using a standard (non-form) prerender for a form submission will result in breakage.
 
 ## Example
 
--- a/prerendering-form-submission.md
+++ b/prerendering-form-submission.md
@@ -13,7 +13,7 @@ It should be noted the speculation will need to be triggered by the page in some
 ### More details
 
 Note that this field is only required for prerender. On the other hand, prefetch speculation rules can be used by form navigations without specifying this field.
-The key difference is that unlike prefetch, which only downloads HTML, prerendering actually starts the navigation. Because parameters are involved, using a standard (non-form) prerender for a form submission will result in breakage."
+The key difference is that unlike prefetch, which only downloads HTML resources, prerendering actually starts the navigation. Because parameters are involved, using a standard (non-form) prerender for a form submission will result in breakage.
 
 ## Example
 
--- a/prerendering-form-submission.md
+++ b/prerendering-form-submission.md
@@ -13,7 +13,7 @@ It should be noted the speculation will need to be triggered by the page in some
 ### More details
 
 Note that this field is only required for prerender. On the other hand, prefetch speculation rules can be used by form navigations without specifying this field.
-The key difference is that unlike prefetch, which only downloads HTML resources, prerendering actually starts the navigation. Because internal navigation parameters in a browser are predetermined, using a standard (non-form) prerender for a form submission will result in breakage.
+The key difference is that unlike prefetch, which only downloads HTML resources, prerendering actually starts the navigation. Because internal navigation parameters in a browser are predetermined, using a standard (non-form) prerender for a form submission will result in problems mentioned in the section "Design considerations and alternatives considered".
 
 ## Example
 
--- a/prerendering-form-submission.md
+++ b/prerendering-form-submission.md
@@ -3,11 +3,8 @@
 Currently, form submissions cannot activate prerendered pages by design, due to internal browser limitations. In at least Chrome, ordinary form submission navigations have special state and run extra checks that ordinary prerenders don't experience.
 This means that a form submission can never activate a prerender, because the prerender was not prepared properly as a form submission.
 On the other hand, prefetch speculation rules, which only downloads HTML resources without starting a navigation with the internal state, can be used by form navigations without specifying this field.
-Examples include a simple search form which results in a “/search?q=XXX” GET request navigation, [support of which has been requested by web developers](https://issues.chromium.org/issues/346555939).
 
-## The proposal
-
-Introduce the `form_submission` field to speculation rules, allowing web developers to specify which speculative navigation is a form submission one. This will only be available for GET form-based speculations.
+We propose an addition to the Speculation Rules API, which allows web developers to specify form_submission in their speculation rules for prerenders, which directs the browser to prepare the prerender as a form submission, so that it can be activated by real form submission navigations. Examples include a simple search form which results in a “/search?q=XXX” GET request navigation, [support of which has been requested by web developers](https://issues.chromium.org/issues/346555939).
 
 It should be noted the speculation will need to be triggered by the page in some manner (e.g. by injecting the rule with JavaScript on hovering over the submit button). This proposal does not add functionality to trigger the speculation, but simply allows a previously-initiated speculation to be matched upon navigation.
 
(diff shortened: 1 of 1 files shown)


prerendering-form-submission.md
```diff
@@ -10,6 +10,11 @@ Introduce the `form_submission` field to speculation rules, allowing web develop
 
 It should be noted the speculation will need to be triggered by the page in some manner (e.g. by injecting the rule with JavaScript on hovering over the submit button). This proposal does not add functionality to trigger the speculation, but simply allows a previously-initiated speculation to be matched upon navigation.
 
+### More details
+
+Note that this field is only required for prerender. On the other hand, prefetch speculation rules can be used by form navigations without specifying this field.
+The key difference is that unlike prefetch, which only downloads HTML resources, prerendering actually starts the navigation. Because parameters are involved, using a standard (non-form) prerender for a form submission will result in breakage.
```
domfarolino: Do you mean "query" parameters? (If so, please specify). And why would that "result in breakage"? Is it because we ignore query parameters in prerender URL matching?

Because if we *don't* ignore query parameters, and we instead respect them, then you can imagine I prerender (non-form) the page `./prerender.html?name=foo&email=bar`, and then later submit a form with those exact parameters. As long as we match the full URL from the form submission against all live prerenders, if they match, why shouldn't my form navigation be able to activate an earlier non-form prerender that communicates the same exact data?

robertlin-chromium: Sorry for the confusion, I meant internal parameters in a browser used for a navigation. I've rephrased this part. PTAL

prerendering-form-submission.md
```diff
@@ -10,6 +10,11 @@ Introduce the `form_submission` field to speculation rules, allowing web develop
 
 It should be noted the speculation will need to be triggered by the page in some manner (e.g. by injecting the rule with JavaScript on hovering over the submit button). This proposal does not add functionality to trigger the speculation, but simply allows a previously-initiated speculation to be matched upon navigation.
 
+### More details
```
tunetheweb:
```suggestion
### Only needed for prerender
```

prerendering-form-submission.md
```diff
@@ -10,6 +10,11 @@ Introduce the `form_submission` field to speculation rules, allowing web develop
...
+
+Note that this field is only required for prerender. On the other hand, prefetch speculation rules can be used by form navigations without specifying this field.
+The key difference is that unlike prefetch, which only downloads HTML resources, prerendering actually starts the navigation. Because internal navigation parameters in a browser are predetermined, using a standard (non-form) prerender for a form submission will result in breakage.
```
tunetheweb: Is "breakage" accurate? Don't you mean "an unused and wasted prerender"?

tunetheweb (changes requested):
Reading this, this seems like a browser problem and doesn't make a clear case why we should work around implementation issues by adding this to the web platform.

I thought the CSP argument was the strongest argument, as the browser can't work around that, but it's not present here.

prerendering-form-submission.md
```diff
@@ -10,6 +10,11 @@ Introduce the `form_submission` field to speculation rules, allowing web develop
 
 It should be noted the speculation will need to be triggered by the page in some manner (e.g. by injecting the rule with JavaScript on hovering over the submit button). This proposal does not add functionality to trigger the speculation, but simply allows a previously-initiated speculation to be matched upon navigation.
 
+### More details
+
+Note that this field is only required for prerender. On the other hand, prefetch speculation rules can be used by form navigations without specifying this field.
```
domfarolino: I would move a reduced version of this whole paragraph up to the top, even before "The proposal". You want to clearly lay out the "problem" before jumping into solutions. If I were you, I'd make the intro of this entire document the following:

> Currently, form submissions cannot activate prerendered pages by design, due to internal browser limitations. In at least Chrome, ordinary form submission navigations have special state and run extra checks that ordinary prerenders don't experience. This means that a form submission can never activate a prerender, because the prerender was not prepared properly as a form submission.
>
> We propose an addition to the [Speculation Rules API](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API), which allows web developers to specify `form_submission` in their speculation rules for prerenders, which directs the browser to prepare the prerender as a form submission, so that it can be activated by real form submission navigations. Examples include [...]

Then get rid of this "more details" section.

prerendering-form-submission.md
robertlin-chromium: I've removed this section as suggested by the comments below.

prerendering-form-submission.md
robertlin-chromium: It was referring to the chrome internal implementation, but  "an unused and wasted prerender" is about all browsers in common. I've also mentioned the issue of wasting resources on unused prerender in the first section. PTAL

prerendering-form-submission.md
```diff
@@ -1,12 +1,11 @@
...
+This means that a form submission can never activate a prerender, because the prerender was not prepared properly as a form submission.
+In addition to the internal browser limitations, resources can be wasted on prerendering a page which is not eligible, such as CSP disallowing `form-action`.
+On the other hand, prefetch speculation rules, which only downloads HTML resources without starting a navigation, can be used by form navigations without specifying this field.
 
-
-## The proposal
-
-Introduce the `form_submission` field to speculation rules, allowing web developers to specify which speculative navigation is a form submission one. This will only be available for GET form-based speculations.
+We propose an addition to the Speculation Rules API, which allows web developers to specify form_submission in their speculation rules for prerenders, which directs the browser to prepare the prerender as a form submission, so that it can be activated by real form submission navigations. Examples include a simple search form which results in a “/search?q=XXX” GET request navigation, [support of which has been requested by web developers](https://issues.chromium.org/issues/346555939).
```
tunetheweb:
```suggestion
We propose an addition to the Speculation Rules API, which allows web developers to specify `form_submission` in their speculation rules for prerenders, which directs the browser to prepare the prerender as a form submission, so that it can be activated by real form submission navigations. Examples include a simple search form which results in a `/search?q=XXX` GET request navigation, [support of which has been requested by web developers](https://issues.chromium.org/issues/346555939).
```

tunetheweb (approved):
LGTM with some code formatting nits.

But will let @domfarolino be the ultimate decider on this.

domfarolino (approved):
> Reading this, this seems like a browser problem and doesn't make a clear case why we should work around implementation issues by adding this to the web platform.
>
> I thought the CSP argument was the strongest argument, as the browser can't work around that, but it's not present here.

This is honestly my impression of the proposal. The CSP / "wasteful prerendering" argument is odd to me. We're just giving developers a vehicle to express `form` intention so these prerenders get caught by restrictive `form-action` CSP directives, but I imagine a developer sophisticated enough to work out the the right speculation rules would also know when `form-action` is present, and could generate compatible prerenders. I presume they'd not prerender banned destinations because:

1. They know that form submissions cannot activate prerenders, since they've tested it.
2. They know their own CSP.

Maybe I'm expecting too much from developers. Either way, if the real reason Chrome doesn't do this is because of internal browser reasons, we should be upfront about that (in addition to the CSP stuff!) and see if other browsers have similar limitations.

----

@robertlin-chromium Does the spec currently stop form submissions from activating prerenders? I know Chrome does, but I can't find anything in the spec that prevents form submissions from activating prerenders.

tunetheweb:
> @robertlin-chromium Does the spec currently stop form submissions from activating prerenders? I know Chrome does, but I can't find anything in the spec that prevents form submissions from activating prerenders.

Other Domenic (with an ‘e’) previously said no:
https://github.com/WICG/nav-speculation/issues/322#issuecomment-2626449625
At least until we decided to add this `form_submission` param.

robertlin-chromium:
> > @robertlin-chromium Does the spec currently stop form submissions from activating prerenders? I know Chrome does, but I can't find anything in the spec that prevents form submissions from activating prerenders.
> 
> Other Domenic (with an ‘e’) previously said no: [#322 (comment)](https://github.com/WICG/nav-speculation/issues/322#issuecomment-2626449625) At least until we decided to add this `form_submission` param.

I think the existing spec doesn't mention anything about the limitation.

There was another shipped field (target_hint) that was also browser implementation specific.
The argument was to mention that this was browser specific, but it would take much engineering work to remove it at that moment. The plan for `target_hint` field was to introduce the field first and mentioned about that it might be removed in the future.
Original PR: https://github.com/w3ctag/design-reviews/issues/931
The parts mention about the plan of removing the field
https://github.com/WICG/nav-speculation/commit/d8ce9b5586bcb079a3942bce079d3d0ee0d1fd17
https://github.com/WICG/nav-speculation/pull/173/changes#diff-086d7831ccea2721156a3a66a719f58f0eb0501d4b0da344f07e7a0df46a3efeR151

domfarolino (approved):

PR #426, opened 2026-02-20
Add Form Submission spec to prerendering.bs

robertlin-chromium:


prerendering.bs
```diff
@@ -242,6 +243,7 @@ Extend the [=speculation rule set=] [=struct=] with two additional [=struct/item
 Extend the [=speculation rule=] [=struct=] with one additional [=struct/item=]:
```
tunetheweb:
```suggestion
Extend the [=speculation rule=] [=struct=] with two additional [=struct/items=]:
```

prerendering.bs
```diff
@@ -258,14 +262,19 @@ Extend the [=speculation rule=] [=struct=] with one additional [=struct/item=]:
...
       1. Return null.
     1. Set |targetHint| to |input|["`target_hint`"].
+  1. If |input|["`form_submission`"] [=map/exists=]:
+    1. If |input|["`form_submission`"] is not a boolean value:
+      1. The user agent may [=report a warning to the console=] indicating that the supplied form submission was invalid.
+      1. Return null.
+    1. Set |formSubmission| to boolean value of |input|["`form_submission`"].
 
-  and then updating the final step which returns a [=speculation rule=] to include setting the [=speculation rule/target navigable name hint=] to |targetHint|.
+  and then updating the final step which returns a [=speculation rule=] to include setting the [=speculation rule/target navigable name hint=] to |targetHint| and [=speculation rule/form submission=] = |formSubmission|.
```
tunetheweb: Not sure if `=` should be `to` for consistency? Or is the fact it's a boolean mean `=` is better?

prerendering.bs
tunetheweb, line 406: Dom we need anything added here for form similar to the name hint?

tunetheweb (commented):
@domfarolino knows much more about specs than me, but spotted a couple of things that might need changing.

prerendering.bs
robertlin-chromium: Using `to` is better for consistency, updated.

prerendering.bs
robertlin-chromium: Yes, prefetchRecord includes both prefetch and prerender, so it is needed. Just updated.

tunetheweb (approved):
LGTM

prerendering.bs
```diff
@@ -258,14 +262,19 @@ Extend the [=speculation rule=] [=struct=] with one additional [=struct/item=]:
 <div algorithm="parse a speculation rule">
   Modify [=parse a speculation rule=] by adding the following steps:
 
-  1. Let |targetHint| be null.
+  1. Let |targetHint| be null and |formSubmission| be false.
```
domfarolino: Can you give this variable declaration its own line?

prerendering.bs
```diff
@@ -317,6 +330,8 @@ A <dfn>prerender candidate</dfn> is a [=speculative load candidate=] with the fo
             1. [=list/For each=] |link| of |links|:
               1. Let |target| be |rule|'s [=speculation rule/target navigable name hint=].
               1. If |target| is null, set it to the result of [=getting an element's target=] given |link|.
+              1. Let |formSubmission| be |rule|'s [=speculation rule/form submission=].
+              1. If |formSubmission| is not specified, set it to be false.
```
domfarolino: It will always be false, since the struct item you added is never null (it just defaults to false). Given that, do we need to do anything here?

prerendering.bs
```diff
@@ -258,14 +262,19 @@ Extend the [=speculation rule=] [=struct=] with one additional [=struct/item=]:
 <div algorithm="parse a speculation rule">
```
domfarolino, line 262: I think we also need to [modify step 2](https://html.spec.whatwg.org/multipage/speculative-loading.html#parse-a-speculation-rule:~:text=If%20input%20has%20any%20key%20other%20than%20%22source%22%2C%20%22urls%22%2C%20%22where%22%2C%20%22relative_to%22%2C%20%22eagerness%22%2C%20%22referrer_policy%22%2C%20%22tag%22%2C%20%22requires%22%2C%20%22expects_no_vary_search%22%2C%20or%20%22target_hint%22%3A), and add `form_submission` to the list of keys that we "accept"?

domfarolino (commented):
Does this PR modify https://wicg.github.io/nav-speculation/prerendering.html#navigate-activation or elsewhere, in order to prevent existing form navigations from activating non-`form_submission` prerenders?

I ask this as a follow-up to https://github.com/WICG/nav-speculation/pull/424#issuecomment-3820798549. This PR doesn't really carry its weight if forms can always activate all prerenders 😅

prerendering.bs
robertlin-chromium: Got it, I just removed the line.

prerendering.bs
robertlin-chromium: I've just updated the activation matching algorithm. PTAL
As for the step 2, it seems a change is needed to another repo: https://github.com/whatwg/html/blob/main/source
I'm creating another PR addressing that.

domfarolino: Note that this algorithm in _this_ specification already modifies (by spec "monkeypatching") the HTML algorithm, so we do not need a PR against whatwg/html to fix this step. Just add:

```md
Modify [=parse a speculation rule=]'s step 2, by adding `form_submission` to the list of `[=map/keys=]` that this step accepts.
```

...

and then you can continue with the rest of this. Does that make sense?

PR #427, opened 2026-02-27
Fix form submission example

tunetheweb:
As noted in the TAG review: https://github.com/w3ctag/design-reviews/issues/1192#issuecomment-3962365845


robertlin-chromium (approved):
LGTM

Issue #368, opened 2025-03-17
Website-provided anonymizing proxy


zcorpan:
Currently, the options for `anonymous-client-ip-when-cross-origin` (ACIWCO) are for the UA to use a UA-defined privacy-preserving proxy (PPP) or to not prefetch at all.

> The user agent must use a [connection](https://fetch.spec.whatwg.org/#concept-connection) which anonymizes the client IP address (e.g., using a proxy) when fetching request, or set response to a [network error](https://fetch.spec.whatwg.org/#concept-network-error) and [break](https://infra.spec.whatwg.org/#iteration-break).

[source](https://wicg.github.io/nav-speculation/prefetch.html#create-navigation-params-by-fetching:~:text=The%20user%20agent%20must%20use%20a%20connection%20which%20anonymizes%20the%20client%20IP%20address%20%28e%2Eg%2E%2C%20using%20a%20proxy%29%20when%20fetching%20request%2C%20or%20set%20response%20to%20a%20network%20error%20and%20break%2E)

We wonder if it's possible for websites to provide their own PPP and specify it as part of the speculation rule. The privacy responsibility would then be burdened by the website and the provided PPP, if the UA uses it for the prefetch. The UA could then choose to still use its own PPP, use the website-provided PPP, or not prefetch.

domenic:
This makes sense to me, and seems like a good idea on the surface. Although I am not intimately familiar with the PPP that Chrome uses and whether there would be any caveats here. Let me get in touch with them and get them to fill me in.

domenic:
We spent some time discussing this internally at Google and all parties are aligned on it being a good direction. (Namely: the Chrome web platform team building the prefetching feature; the team running the PPP Chrome is using; and Google Search, which is the current major user of the proxy.)

However, the details will be tricky to get right and will take some time. The two main areas for discussion would be:

* How or whether we can specify an open standard for such proxies, what they must behave like, and how to ensure they're secure. For example, they must not allow the referrer site to change the prefetched content. As a first approximation, a passthrough CONNECT where the target URL must be HTTPS and the proxy must be hosted on the same domain as the referrer is a good starting point, but there are probably a lot more details.

* How the proxy operator can be open to a wider set of users, from multiple browsers, without just supplying a "free VPN" to anyone. Today Chrome + Google's PPP solves this in the same way that Chrome couples to other Google services, like sync, translation, and wallet. Coming up with some new system will be trickier.

[...]

PR #428, opened 2026-03-10
Add explainer and server requirements for Referrer-Provided Prefetch Proxies

rrliu:
This PR introduces an early design sketch for the **Referrer-Provided Prefetch Proxies** architecture.

It allows referring sites to operate their own privacy-preserving proxy servers for cross-origin prefetching. This shifts the operational cost to the party benefiting from faster navigations and aligns the proxy trust model without compromising client privacy.

### Documents added
1. **`referrer_provided_prefetch_proxies_draft.md`**: The main explainer outlining the problem, use cases, and proposed changes (new Speculation Rule requirements). Includes answers to the security and privacy self-review questionnaire.
2. **`referrer_provided_prefetch_proxy_server_requirements.md`**: Client-server protocol and server requirements (abuse prevention, privacy goals, geolocation).


rrliu:
This explainer intends to address issue #368.

referrer_provided_prefetch_proxies_draft.md
```diff
@@ -0,0 +1,249 @@
+# Explainer: Referrer-Provided Prefetch Proxies
+
+This proposal is an early design sketch by the Chrome Prefetch Proxy team to describe the problem below and solicit feedback on the proposed solution. It has not been approved to ship in Chrome.
```
tunetheweb, line 3: Let's add a one liner description to here, so people don't need to scroll down to even understand if they're interested in this.

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Introduction
+
+Web developers use **speculation rules** to instruct the browser to perform navigation actions like prefetching or prerendering before navigation starts, speeding up subsequent navigations. For cross-origin prefetches, the user’s IP address must be anonymized to prevent the origin from learning that a UA speculatively navigated to a page before the user expressed intent to navigate.
+
+IP anonymization [is understood to mean that the user agent should hide or mask the external IP address of the client](https://github.com/WICG/nav-speculation/blob/main/anonymous-client-ip.md). IP anonymization may be requested by a referring page to avoid leaking client PII when performing a cross-origin prefetch. This requirement is expressed as a speculation rule requirement: "anonymous-client-ip-when-cross-origin" (**ACIWCO**). Speculation rules can include arrays of speculation rule requirements and ACIWCO is the only one yet defined. When set in a rule, ACIWCO requires a **cross-origin prefetch IP anonymization policy** to be set for the prefetch candidate corresponding to the rule.
```
tunetheweb:
```suggestion
IP anonymization [is understood to mean that the user agent should hide or mask the external IP address of the client](https://github.com/WICG/nav-speculation/blob/main/anonymous-client-ip.md). IP anonymization may be requested by a referring page to avoid leaking client PII when performing a cross-origin prefetch. This requirement is expressed as a speculation rule requirement: `"anonymous-client-ip-when-cross-origin"` (**ACIWCO**). Speculation rules can include arrays of speculation rule requirements and ACIWCO is the only one yet defined. When set in a rule, ACIWCO requires a **cross-origin prefetch IP anonymization policy** to be set for the prefetch candidate corresponding to the rule.
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+Today, the Prefetch spec asks browsers to implement a cross-origin prefetch IP anonymization policy. Previous work has described [a browser-provided privacy-preserving prefetch proxy](https://github.com/buettner/private-prefetch-proxy). This isn’t ideal for a few reasons: costs are a significant factor for vendors; privacy concerns limit the use of browser-provided prefetch proxies; and referrers don’t share the cost of supporting their speculation rules.
```
tunetheweb: Link to spec. And let's move cost to be the secondary consideration and add a third reason.

```suggestion
Today, [the spec](https://html.spec.whatwg.org/multipage/speculative-loading.html#speculative-loading:~:text=If%20rule%27s%20requirements%20contains%20%22anonymous%2Dclient%2Dip%2Dwhen%2Dcross%2Dorigin%22%2C%20then%20set%20anonymizationPolicy%20to%20a%20cross%2Dorigin%20prefetch%20IP%20anonymization%20policy%20whose%20origin%20is%20document%27s%20origin.) asks browsers to implement a cross-origin prefetch IP anonymization policy.

Previous work has described [a browser-provided privacy-preserving prefetch proxy](https://github.com/buettner/private-prefetch-proxy). This isn’t ideal for a few reasons: privacy concerns limit the use of browser-provided prefetch proxies; costs are also a significant factor for vendors—referrers don’t share the cost of supporting their speculation rules.
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+This explainer describes another model, a referrer-provided prefetch proxy, in order to get public feedback on its feasibility for standardization. In the **referrer-provided prefetch proxy** architecture, the website initiating the prefetch (the referrer) operates its own privacy-preserving proxy server and specifies it in the page’s Speculation Rules script. This shifts the operational cost to the party that directly benefits from faster navigation and aligns the proxy trust model.
```
tunetheweb:
```suggestion
This explainer describes another model, a referrer-provided prefetch proxy, in order to get public feedback on its feasibility for standardization. In the **referrer-provided prefetch proxy** architecture, the website initiating the prefetch (the referrer) operates its own privacy-preserving proxy server and specifies it in the page’s speculation rules script. This aligns the proxy trust model, prevents introducing a third-party service, while also shifting the operational cost to the party that directly benefits from faster navigation.
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+These findings suggest that enabling cross-origin prefetch in a more scalable and accessible way could yield similar benefits for the broader web ecosystem.
+
+## Use cases
+
+A site owner wants to speed up cross-origin/site navigations on their pages. They want to get the same performance improvement on cross-origin navigation as they do on same-site navigation.
+
+Their current implementation options are limited:
+* Use Speculation Rules API or another means of speculatively loading resources without a privacy-preserving proxy. **This exposes client IP addresses to 3P origins.**  
+* Use Speculation Rules API with the ACIWCO requirement. The satisfaction of this requirement is left up to the UA implementation, which might not support ACIWCO. If the UA doesn’t support ACIWCO, it [must not execute any rules with that requirement](https://wicg.github.io/nav-speculation/prefetch.html#ref-for-prefetch-record-anonymization-policy%E2%91%A0).
```
tunetheweb:
```suggestion
* Use Speculation Rules API with the ACIWCO requirement. The satisfaction of this requirement is left up to the UA implementation, which might not support ACIWCO. If the UA doesn’t support ACIWCO, it [must not execute any rules with that requirement](https://wicg.github.io/nav-speculation/prefetch.html#ref-for-prefetch-record-anonymization-policy%E2%91%A0). Additionally, since this introduces a third-party service run by the browser-vendor, the UA may require additional opt-ins ([Chrome, the only browser currently running such a proxy [requires users to enable the "Extended preloading" mode in Chrome's [preload settings](https://support.google.com/chrome/answer/1385029?&co=GENIE.Platform%3DAndroid)](https://developer.chrome.com/blog/private-prefetch-proxy#referrers:~:text=to%20allow%20other%20sites%20to%20preload%20navigations%20through%20Google%20servers%2C%20users%20need%20to%20select%20the%20%22Extended%20preloading%22%20mode%20in%20Chrome%27s%20preload%20settings.)).
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+Sites may want to implement their own prefetch proxy if:
+* Browser-provided implementations don't exist. Referrer-provided prefetch proxies can support private cross-origin prefetch in every browser that implements the feature.
+* Sites may not want to send prefetch traffic to another company’s proxy infrastructure.
+* The browser-provided implementation requires a user opt-in for privacy reasons. Having the same party operate the referrer page and proxy prevents additional information leakage to the proxy operator, which may obviate the need for user opt-in.
+* The site wants more direct control over prefetch usage, cost, or infrastructure.
+* Sites may already have the infrastructure in place to operate a prefetch proxy.
+
+## Proposed solution: referrer-provided prefetch proxies
+
+We propose the **referrer-provided prefetch proxy** architecture, where the referring website operates its own privacy-preserving proxy server and specifies it in the Speculation Rules.
```
tunetheweb:
```suggestion
We propose the **referrer-provided prefetch proxy** architecture, where the referring website operates its own privacy-preserving proxy server and specifies it in the speculation rules.
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+### New speculation rule requirements
+
+We propose adding two new speculation rule requirements at the rule level:
+
+* “use-referrer-provided-prefetch-proxy”, which instructs the browser to use a specific privacy-preserving prefetch proxy  
+* “prefetch-proxy-authorization-tokens”, which instructs the browser to use authorization tokens provided in the rule when connecting to a specified privacy-preserving prefetch proxy.
```
tunetheweb:
```suggestion
* `"use-referrer-provided-prefetch-proxy"`, which instructs the browser to use a specific privacy-preserving prefetch proxy  
* `"prefetch-proxy-authorization-tokens"`, which instructs the browser to use authorization tokens provided in the rule when connecting to a specified privacy-preserving prefetch proxy.
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+When one of the above requirements is set, its respective key must be set with an appropriate value in the speculation rule JSON object:
+
+* “referrer\_provided\_prefetch\_proxy”: Valid URL string  
+* “proxy\_authorization\_tokens”: List of JavaScript strings
```
tunetheweb:
```suggestion
* `"referrer\_provided\_prefetch\_proxy"`: Valid URL string  
* `"proxy\_authorization\_tokens"`: List of JavaScript strings
```

referrer_provided_prefetch_proxies.md
tunetheweb: Why is this needed? Isn't the presence of the two below keys sufficient to know the rules needs these?

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+These keys are parsed into corresponding items in the speculation rule struct:
+
+* Referrer provided prefetch proxy URL, a URL  
+* Proxy authorization tokens, an ordered set of (ASCII strings?)
```
tunetheweb: Why question mark at the end? Is this still TBD? If so let's sayh that:


```suggestion
* Proxy authorization tokens, an ordered set of ASCII strings (to be confirmed)
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+Speculation rules with the new requirements are only well-defined when:
+
+* **All three requirements are set:** UA must use the referrer-provided prefetch proxy and authenticate with the provided authorization tokens. If this connection does not succeed, the UA may fall back to its own IP anonymization implementation.
```
tunetheweb: I think you should be clearer here what you mean by "three" and "requirements". Maybe code-formatting `requirement` is enough?


```suggestion
* **All three `requirements` are set:** UA must use the referrer-provided prefetch proxy and authenticate with the provided authorization tokens. If this connection does not succeed, the UA may fall back to its own IP anonymization implementation.  
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+These keys are parsed into corresponding items in the speculation rule struct:
+
+* Referrer provided prefetch proxy URL, a URL  
+* Proxy authorization tokens, an ordered set of (ASCII strings?)
+
+Speculation rules with the new requirements are only well-defined when:
+
+* **All three requirements are set:** UA must use the referrer-provided prefetch proxy and authenticate with the provided authorization tokens. If this connection does not succeed, the UA may fall back to its own IP anonymization implementation.  
+* **Only the two new requirements are set:** UA must use the referrer-provided prefetch proxy and authenticate with the provided authorization tokens. If this connection does not succeed, the UA stops execution of this speculation rule.
```
tunetheweb: Be clearer:

```suggestion
* **Only the two new `requirements` are set, without `"anonymous-client-ip-when-cross-origin"`:** UA must use the referrer-provided prefetch proxy and authenticate with the provided authorization tokens. If this connection does not succeed, the UA stops execution of this speculation rule.  
```

This also feels weird, because we still are anonymising the client ip when cross origin. But if we want to support with and without fallback I can't think of a better way...

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+* **All three requirements are set:** UA must use the referrer-provided prefetch proxy and authenticate with the provided authorization tokens. If this connection does not succeed, the UA may fall back to its own IP anonymization implementation.  
+* **Only the two new requirements are set:** UA must use the referrer-provided prefetch proxy and authenticate with the provided authorization tokens. If this connection does not succeed, the UA stops execution of this speculation rule.  
+* **Only ACIWCO is set:** No change from current spec – UA either provides IP anonymization for this prefetch, or it [must not execute the speculation rule](https://github.com/WICG/nav-speculation/blob/main/anonymous-client-ip.md).  
+* **No requirements are set:** UA will not use a proxy to prefetch this speculation rule candidate.
+
+After this change, the meaning of the existing **ACIWCO** requirement will change. The requirement is now associated with using the UA’s IP-anonymization implementation. If the two new requirements are set, **ACIWCO** will be used as a fallback, not as the first-choice proxy, if the connection to the referrer-provided proxy does not succeed.
```
tunetheweb: This isn't clearer.

```suggestion
After this change, the meaning of the existing **ACIWCO** `requirement` will change when the two additional new `requirement`s are set. At the minute, that us a directive to always use the UA’s IP-anonymization implementation, but with this change it will **ACIWCO** will be used as a fallback, not as the first-choice proxy, and only used if the connection to the referrer-provided proxy does not succeed.
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+A UA should parse the new keys by:
+
+1. Let *referrerProvidedPrefetchProxyURL* be an empty URL.  
+2. If input\[“referrer\_provided\_prefetch\_proxy”\] exists:  
+   1. If input\[“referrer\_provided\_prefetch\_proxy”\] is not a valid URL string:  
+      1. The user agent may report a warning to the console indicating that the referrer-provided URL was not understood.  
+      2. Return null.  
+   2. Set *referrerProvidedPrefetchProxyURL* to input\[“referrer\_provided\_prefetch\_proxy”\].  
+3. Let *proxyAuthorizationTokens* be an empty set.  
+4. If input\[“proxy\_authorization\_tokens”\] exists, append input\[“proxy\_authorization\_tokens”\] to *proxyAuthorizationTokens*.
```
tunetheweb: Remove smart quotes.

Also do you need all those escapes?

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+* Enable any referring site to offer privacy-preserving cross-origin prefetch to its users without dependence on browser-provided proxy infrastructure.  
+* Give origins control over prefetch traffic using existing mechanisms: .well-known/traffic-advice configuration, Sec-Purpose headers.  
+* Specify how referrer-provided proxies can be privacy-preserving.  
+* Specify how cross-origin/site prefetches can avoid leaking additional information about clients.
+
+## Non-goals
+
+* This proposal is scoped to cross-origin/site prefetch only. Same-origin/site prefetch and prerendering are out of scope.  
+* This proposal does not define how user agents or sites should implement a privacy-preserving prefetch architecture, only the privacy and security goals of such architectures.
```
tunetheweb: I'm confused as below it seems you are defining this?

https://github.com/rrliu/nav-speculation/blob/4bbb8390c3390de592ed6445ca11a8d2d5fdd52f/referrer_provided_prefetch_proxies.md#creating-an-ip-anonymized-connection

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+* Require prefetch servers to respect ./well-known/traffic-advice files
+
+**See [Referrer-provided Prefetch Proxy Server Requirements](referrer_provided_prefetch_proxy_server_requirements.md) for more details.**
+
+## Detailed design discussion
+
+### Specifying a standard for privacy-preserving prefetch proxies
+
+* **Should the referrer-provided proxy take precedence over the UA IP anonymization implementation when all three speculation rule requirements are listed in a rule?**  
+  Yes: the site is explicitly providing a resource for the UA. If cost is a concern, this prevents the UA from paying for IP anonymization itself.
```
tunetheweb: I don't see the need to mention cost here. I don't think that's the primary motiviation of implementing this spec!

```suggestion
  Yes: the site is explicitly providing a resource for the UA.
```

referrer_provided_prefetch_proxies_draft.md
```diff
@@ -0,0 +1,249 @@
...
+  An encrypted version of HTTP CONNECT (HTTPS, H/2, H/3). Connection should be encrypted from the UA to the proxy and from the proxy to the origin server.  
+* **Privacy standards for proxy implementations**  
+  Additional privacy requirements are discussed in the attached document on proxy server requirements.
+
+### Token handling algorithm for CONNECT requests
+
+* **How do site developers control which tokens are used by a UA in a specific request?**  
+  Proposal: it shouldn’t matter exactly which token a UA uses, the site should provide as many tokens as there are prefetch candidates.  
+* **What happens if a speculation rules JSON document doesn’t provide exactly as many tokens as there are prefetch candidates?**  
+  Proposal: UA will select and remove a token at random from the ordered set of tokens and use that token for a CONNECT request. If the set is empty, the “create navigation params” algorithm should fail and print a warning to the console.
```
tunetheweb, line 204: I thought it was one token for all requests.

A token per prefetch seems unworkable for document rules. And even for URL lists rules where a URL may be speculated multiple times.

referrer_provided_prefetch_proxies_draft.md
```diff
@@ -0,0 +1,249 @@
...
+  Proposal: UA will select and remove a token at random from the ordered set of tokens and use that token for a CONNECT request. If the set is empty, the “create navigation params” algorithm should fail and print a warning to the console.  
+* **What if a site developer wants to use the same token for every prefetch request?**  
+  Proposal: Call this something else to distinguish from the temporary token use case. Open to naming suggestions.
+
+## Considered alternatives
+
+### Status quo: browser-provided proxies only
+
+* Expensive for browsers to maintain and offer.  
+* Trade-off between privacy and cost: using a single proxy server reveals information about the target sites to the proxy; using a two-hop proxy chain hides more information but is more expensive.
```
tunetheweb, line 216:
```suggestion
* Using a single proxy server reveals information about the target sites to the proxy
* Using a two-hop proxy chain hides more information but is more complex.
* Keeping this as an opt-in, off-by-default for privacy reasons severely limits it's use.
* Expensive for browsers to maintain and offer.
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+We ruled out Oblivious HTTP as an alternative proxying protocol because OHTTP is message-oriented rather than connection-oriented and thus easier to MITM. Connection-oriented protocols establish a TLS connection from the UA to the origin and ensure that the proxy server can’t tamper with the HTTP request in the CONNECT tunnel.
+
+## Security and Privacy Considerations
+
+Prefetching web pages has the risk of exposing user information through speculative navigations. We consider the following parties in our privacy model: User agent, Referrer page, Private prefetch proxy, and Origin server.
+
+The goal of this privacy model is to define principles that result in the least information being available to the proxy and the origin server.
+
+First, no information about the client is leaked to the proxy *if the referrer and proxy are run by the same party*. The proxy server learns only the same information about the client that has been revealed when the client navigated to the referrer page. Therefore **we propose that the referrer-provided proxy must be same-site to the referrer page.**
```
tunetheweb: I'm not sure I agree with this wording.

Keeping something on the client, versus sending to a server (even if run by the same party) could still potentially leak the data. Otherwise, by that rational, the UA doesn't leak anything when using a UA-proxy.

```suggestion
First, **we propose that the referrer-provided proxy must be same-site to the referrer page.**. If the referrer and proxy, then the client has all the information so any data is not being shared further. Allowing non same-site proxies reintroduces many of the issues with the original UA implementation.
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+Second, **the user agent should not selectively send prefetch requests based on client PII** to avoid leaking information about PII. When the referrer and the proxy are run by the same party, that party can identify which prefetches are requested by the referring page and not sent to the proxy. This can leak PII that the user agent uses in the decision to prefetch, like the presence of a cookie for a particular site.
```
tunetheweb: Finding it difficult to read. Suggesting this reword.

```suggestion
Second, **the user agent should not, to the best of it's abilities, send prefetch requests potentially containing client PII**, to avoid leaking information about PII. For example, proxying requests with cookies and including those cookies could negate the privacy-preserving intentions of using the proxy. However, this complicates matters as when the referrer and the proxy are run by the same party, that party can identify which prefetches are requested via the proxy or not, and so leak the fact that cookies exist for that origin. For this reason, the user-agent should still send requests, but without cookies, and then discard the responses.
```

I think this what Search does right? https://developer.chrome.com/blog/search-speculation-rules#privacy-preserving-prefetches:~:text=Note%3A%20Chrome%20will%20still%20send%20the%20anonymous%20prefetch%20request%20(without%20cookies)%20so%20the%20site%20(Google%20Search%20in%20this%20case)%20is%20unable%20to%20know%20the%20user%20has%20cookies%20for%20a%20third%2Dparty%20site%2C%20but%20the%20prefetch%20response%20will%20be%20discarded.

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+
+**See the [Security and Privacy Questionnaire](referrer_provided_prefetch_proxy_security_privacy_questionnaire.md) for more information.**
+
+## Stakeholder Feedback / Opposition
+
+* **Firefox:** No public position. Firefox engineers wrote the initial public proposal for this feature in March 2025: [https://github.com/WICG/nav-speculation/issues/368](https://github.com/WICG/nav-speculation/issues/368).
```
tunetheweb:
```suggestion
* **Firefox:** No public position. Firefox engineers originally suggested this feature in March 2025: [https://github.com/WICG/nav-speculation/issues/368](https://github.com/WICG/nav-speculation/issues/368).  
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,249 @@
...
+* **Firefox:** No public position. Firefox engineers wrote the initial public proposal for this feature in March 2025: [https://github.com/WICG/nav-speculation/issues/368](https://github.com/WICG/nav-speculation/issues/368).  
+* **Safari:** No public position.  
+* **Edge:** No public position.
+
+## References & acknowledgements
+
+* [HTML Standard, “Speculative loading”](https://html.spec.whatwg.org/multipage/speculative-loading.html#speculative-loading)
+* [Prefetch draft CG report](https://wicg.github.io/nav-speculation/prefetch.html)
+* ["Private Prefetch Proxy Explained"](https://github.com/buettner/private-prefetch-proxy), [Michael Buettner](mailto:buettner@google.com)   
+* Thanks to [Dominic Farolino](mailto:domfarolino@google.com) for the spec mentorship\!
```
tunetheweb: Maybe not give out his email :-)


```suggestion
* Thanks to [Dominic Farolino](https://github.com/domfarolino) for the spec mentorship\!
```

referrer_provided_prefetch_proxy_security_privacy_questionnaire.md
```diff
@@ -0,0 +1,74 @@
+# Questions to Consider
```
tunetheweb: Does this need to be a separate doc?

tunetheweb (commented):
Only really reviewed the first explainer so far. Will look at the other two later.

referrer_provided_prefetch_proxy_security_privacy_questionnaire.md
domfarolino: Agreed, I'd probably stuff this in the bottom of `referrer_provided_prefetch_proxies.md`.

referrer_provided_prefetch_proxies_draft.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: I'd like the semantics to be explicit: the rule cannot be executed without using the referrer-provided proxy (with tokens if given). Defining them as requirements is explicit and puts them on the same level as ACIWCO.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: Done. Yeah, this case is why I wanted to clarify things in the following paragraph.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: Removed.

referrer_provided_prefetch_proxies_draft.md
rrliu: Added.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: Done. Sorry about that!

referrer_provided_prefetch_proxies_draft.md
rrliu: That's useful to know, we do want to cover those use cases if they're cross-origin. It sounds like the proxy operator should verify the token and verify that the target URL matches the rule? That would let us use one token for many URLs covered under a single rule.

Are there any other constraints about speculation rules that would help site & proxy operators design a token scheme?

referrer_provided_prefetch_proxies.md
rrliu: Updated to "if the referrer and proxy are same-site, the proxy operator has all the information the referrer received. No additional information is sent to the proxy that the referrer didn't already receive."

referrer_provided_prefetch_proxies.md
rrliu: I expanded on this a bit differently. The original point was that UAs shouldn't use client-side data to decide whether to prefetch, because that can indirectly leak data. I think your reword is saying that we shouldn't directly send that data in prefetch requests either, and I added that to this point.

referrer_provided_prefetch_proxies.md
rrliu: Removed. Automated MD conversion from a Google Doc put the escapes in, but I believe underscores are used for italics, so I'll keep them in.

referrer_provided_prefetch_proxies.md
rrliu: Removed this line. My meaning is that we don't want to mandate how proxy operators should implement their proxy outside of defining the browser-proxy protocol.

referrer_provided_prefetch_proxy_security_privacy_questionnaire.md
rrliu: Moved the questionnaire to the end of the main explainer.

referrer_provided_prefetch_proxies_draft.md
```diff
@@ -0,0 +1,328 @@
...
+
+Second, **the user agent should not selectively send prefetch requests based on client PII, or include PII in prefetch requests** to avoid leaking information about PII. When the referrer and the proxy are run by the same party, that party can identify which prefetches are requested by the referring page and not sent to the proxy. This can indirectly leak PII that the user agent uses in the decision to prefetch, like the presence of a cookie for a particular site. Including PII in prefetch requests, like cookies or service workers, may directly leak user identity to origin servers.
+
+See the **Security and Privacy Questionnaire** section below for more information.
+
+## Stakeholder Feedback / Opposition
+
+* **Firefox:** No public position. Firefox engineers originally suggested this feature in March 2025: [https://github.com/WICG/nav-speculation/issues/368](https://github.com/WICG/nav-speculation/issues/368).  
+* **Safari:** No public position.  
+* **Edge:** No public position.
```
domfarolino, line 246: Have you filed standards position issues?

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,328 @@
...
+
+First, **we propose that the referrer-provided proxy must be same-site to the referrer page.**. If the referrer and proxy are same-site, then the proxy operator has all the information that the referrer received. No additional user information is sent to the proxy that the referrer didn't already receive. Allowing non same-site proxies reintroduces many of the issues with the original UA implementation.
+
+Second, **the user agent should not selectively send prefetch requests based on client PII, or include PII in prefetch requests** to avoid leaking information about PII. When the referrer and the proxy are run by the same party, that party can identify which prefetches are requested by the referring page and not sent to the proxy. This can indirectly leak PII that the user agent uses in the decision to prefetch, like the presence of a cookie for a particular site. Including PII in prefetch requests, like cookies or service workers, may directly leak user identity to origin servers.
+
+See the **Security and Privacy Questionnaire** section below for more information.
```
domfarolino: Maybe link to the section so clicking it performs a fragment navigation?

referrer_provided_prefetch_proxies_draft.md
rrliu: No, I haven't.

Should I file them at this stage? [Blink > Launching Features > Getting Wider Review](https://www.chromium.org/blink/launching-features/wide-review/) suggested getting formal standards positions around OT / I2S time.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies_draft.md
domfarolino: Eh, I'd probably do it sooner than later honestly.

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,328 @@
...
+  - [OHTTP-based approach](#ohttp-based-approach)
+- [Security and Privacy Considerations](#security-and-privacy-considerations)
+- [Stakeholder Feedback / Opposition](#stakeholder-feedback--opposition)
+- [References & acknowledgements](#references--acknowledgements)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Introduction
+
+Web developers use **speculation rules** to instruct the browser to perform navigation actions like prefetching or prerendering before navigation starts, speeding up subsequent navigations. For cross-origin prefetches, the user’s IP address must be anonymized to prevent the origin from learning that a UA speculatively navigated to a page before the user expressed intent to navigate.
```
domfarolino: The IP address must be anonymized to prevent origin leakage. But it doesn't have to be anonymized for x-o prefetch to work in general, right? If there is a distinction here, let's clarify it.

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,328 @@
...
+
+IP anonymization [is understood to mean that the user agent should hide or mask the external IP address of the client](https://github.com/WICG/nav-speculation/blob/main/anonymous-client-ip.md). IP anonymization may be requested by a referring page to avoid leaking client PII when performing a cross-origin prefetch. This requirement is expressed as a speculation rule requirement: `"anonymous-client-ip-when-cross-origin"` (**ACIWCO**). Speculation rules can include arrays of speculation rule requirements and ACIWCO is the only one yet defined. When set in a rule, ACIWCO requires a **cross-origin prefetch IP anonymization policy** to be set for the prefetch candidate corresponding to the rule.
```
domfarolino: Let's link to https://html.spec.whatwg.org/#speculation-rule-requirement for the header value, so we at least get nearby it.

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,328 @@
...
+
+This explainer describes another model, a referrer-provided prefetch proxy, in order to get public feedback on its feasibility for standardization. In the **referrer-provided prefetch proxy** architecture, the website initiating the prefetch (the referrer) operates its own privacy-preserving proxy server and specifies it in the page’s speculation rules script. This aligns the proxy trust model, avoids introducing a third-party service, and shifts the operational cost to the party that directly benefits from faster navigation.
```
domfarolino:
```suggestion
This explainer introduces a referrer-provided prefetch proxy. In the **referrer-provided prefetch proxy** architecture, the website initiating the prefetch (the referrer) operates its own privacy-preserving proxy server and specifies it in the page’s speculation rules script. This aligns the proxy trust model, avoids introducing a third-party service, and shifts the operational cost to the party that directly benefits from faster navigation.
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,328 @@
...
+
+## Goals
+
+* Enable any referring site to offer privacy-preserving cross-origin prefetch to its users without dependence on browser-provided proxy infrastructure.
```
domfarolino:
```suggestion
* Enable any referring site to offer privacy-preserving cross-origin prefetch to its users without relying on browser-provided proxy infrastructure.  
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,328 @@
...
+## Goals
+
+* Enable any referring site to offer privacy-preserving cross-origin prefetch to its users without dependence on browser-provided proxy infrastructure.  
+* Give origins control over prefetch traffic using existing mechanisms: .well-known/traffic-advice configuration, Sec-Purpose headers.  
+* Specify how referrer-provided proxies can be privacy-preserving.  
+* Specify how cross-origin/site prefetches can avoid leaking additional information about clients.
+
+## Non-goals
+
+* This proposal is scoped to cross-origin/site prefetch only. Same-origin/site prefetch and prerendering are out of scope.
```
domfarolino:
```suggestion
* This proposal is scoped to cross-origin prefetch only. Same-origin prefetch and prerendering are out of scope.  
```

If it's scoped to the cross-origin case, then it applies to the cross-site case by construction. So since cross-origin is the tighter condition, let's limit discussion to it here and in all below instances. Does that make sense?

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,328 @@
...
+* Specify how cross-origin/site prefetches can avoid leaking additional information about clients.
+
+## Non-goals
+
+* This proposal is scoped to cross-origin/site prefetch only. Same-origin/site prefetch and prerendering are out of scope.  
+* This proposal does not deprecate the ability of user agents to operate IP anonymizing services for cross-origin/site prefetching. That will remain an implementation option.
+
+## User research
+
+Site owners have shown significant performance gains from using Speculation Rules for same-site and cross-origin navigations. For example:
```
domfarolino: What does "same-site and cross-origin navigations" refer to here? Does it refer to:
- both "same-site, cross-origin navigations AND cross-site cross-origin" navigations; or
- only one type of navigation: "same-site, cross-origin navigations"

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,328 @@
...
+      "proxy_authorization_tokens": ["EXAMPLE_TOKEN"]
+    }
+  ]
+}
+</script>
+```
+
+### Parsing model
+
+A UA should parse the new keys by:
```
domfarolino: "should" might get us in trouble, if read through the lens of https://infra.spec.whatwg.org/#conformance. Do you mean something like "will"?

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,328 @@
...
+   1. If input\["referrer\_provided\_prefetch\_proxy"\] is not a valid URL string:  
+      1. The user agent may report a warning to the console indicating that the referrer-provided URL was not understood.  
+      2. Return null.  
+   2. Set *referrerProvidedPrefetchProxyURL* to input\["referrer\_provided\_prefetch\_proxy"\].  
+3. Let *proxyAuthorizationTokens* be an empty set.  
+4. If input\["proxy\_authorization\_tokens"\] exists, append input\["proxy\_authorization\_tokens"\] to *proxyAuthorizationTokens*.
+
+### Processing model
+
+We will add the following items to the definition of a cross-origin prefetch IP anonymization policy struct:
```
domfarolino: Link to the struct.

referrer_provided_prefetch_proxies_draft.md
```diff
@@ -0,0 +1,328 @@
...
+   2. Set *referrerProvidedPrefetchProxyURL* to input\["referrer\_provided\_prefetch\_proxy"\].  
+3. Let *proxyAuthorizationTokens* be an empty set.  
+4. If input\["proxy\_authorization\_tokens"\] exists, append input\["proxy\_authorization\_tokens"\] to *proxyAuthorizationTokens*.
+
+### Processing model
+
+We will add the following items to the definition of a cross-origin prefetch IP anonymization policy struct:
+
+* Referrer-provided prefetch proxy, a URL  
+* Proxy authorization tokens, an ordered set of (ASCII strings?)
```
domfarolino, line 159: Are we unsure whether they'll be ASCII? What format are we accepting the tokens in? base64? something else?

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,328 @@
...
+   1. Set parameters in the prefetch record struct.  
+   2. Start a referrer-initiated navigational prefetch.
+
+### Creating an IP anonymized connection
+
+We will define an algorithm for browsers to create an IP anonymized connection. The algorithm will:
+
+* Require browsers to start an encrypted HTTP CONNECT tunnel session with the proxy server  
+* Specify headers for the CONNECT request, such as Host (of the origin) and Proxy-Authorization (for authorization tokens)  
+* **Require the proxy URL to be same-site to the referrer document**
```
domfarolino: Is this requirement special or surprising in some way? (Why the asterisks?)

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,328 @@
...
+* Formalize the ./well-known/traffic-advice file for site developers  
+* Require prefetch servers to respect ./well-known/traffic-advice files
+
+**See [Referrer-provided Prefetch Proxy Server Requirements](referrer_provided_prefetch_proxy_server_requirements.md) for more details.**
+
+## Detailed design discussion
+
+### Specifying a standard for privacy-preserving prefetch proxies
+
+* **Should the referrer-provided proxy take precedence over the UA IP anonymization implementation when all three speculation rule requirements are listed in a rule?**
```
domfarolino:
```suggestion
* **Should the referrer-provided proxy take precedence over the UA IP anonymization implementation (ACIWCO) when all three speculation rule requirements are listed in a rule?**  
```

referrer_provided_prefetch_proxies.md
```diff
@@ -0,0 +1,328 @@
...
+  Yes: the site is explicitly providing a resource for the UA.
+* **What network protocol should UAs use to connect to a proxy?**  
+  An encrypted version of HTTP CONNECT (HTTPS, H/2, H/3). Connection should be encrypted from the UA to the proxy and from the proxy to the origin server.  
+* **Privacy standards for proxy implementations**  
+  Additional privacy requirements are discussed in the attached document on proxy server requirements.
+
+### Token handling algorithm for CONNECT requests
+
+* **How do site developers control which tokens are used by a UA in a specific request?**  
+  Proposal: it shouldn’t matter exactly which token a UA uses, the site should provide as many tokens as there are prefetch candidates.
```
domfarolino: Wait, so if there are 100 resources on the site, the site should provide 100 proxy tokens? I thought the token was just for one-off auth to the proxy site?

referrer_provided_prefetch_proxies_draft.md
domfarolino: Ah, looks like I have the same question immediately above this. I don't have much to add here except that I agree that one token per request is unworkable for the web, so we'd need one to rule them all. Why is the token schema more complicated because of this?

referrer_provided_prefetch_proxies.md
rrliu: Updated to clarify that x-o prefetch has privacy concerns that create a need for client IP anonymization.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: That makes sense - updated to cross-origin.

referrer_provided_prefetch_proxies.md
rrliu: Changed to navigations in general.

referrer_provided_prefetch_proxies.md
rrliu: Yes, updated to "will".

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies_draft.md
rrliu: Updated to strings, with a link to https://infra.spec.whatwg.org/#strings.

referrer_provided_prefetch_proxies.md
rrliu: Removed the asterisks.

referrer_provided_prefetch_proxies.md
rrliu: Done.

referrer_provided_prefetch_proxies.md
rrliu: Updated this with some other options for token handling. I agree that token-per-CONNECT-request seems impractical for complex rules.

referrer_provided_prefetch_proxies_draft.md
rrliu: It's manageable complexity. The site/proxy would likely additionally track:
- What URLs the token is eligible to prefetch
- How many times the token has been used
- Client IP association with the token

Just some additional measures to prevent token reuse and abuse of the site's proxy.

referrer_provided_prefetch_proxies_draft.md
rrliu: Sounds good. I can file them after this is merged so that I can include a proper link to the explainer.

referrer_provided_prefetch_proxies.md
domfarolino: This does not seem done.

referrer_provided_prefetch_proxies_draft.md
domfarolino: Hmm, OK so what is the result of this thread? Are we keeping the one-token-per-request approach? You mentioned that the site would have to track "how many times the token has been used", which implies reuse to the tune that me and @tunetheweb were pushing for. But then below I see:

> What if a site developer wants to use the same token for every prefetch request?**
> Proposal: Call this something else to distinguish from the temporary token use case. Open to naming suggestions.

And the proposal still indicates that the set of tokens is monotonically decreasing with each request.

----

Finally, have we preflighted this design past any partners to see if the complexity is actually compatible with their design? Was Mozilla envisioning this kind of approach when they expressed interest? (Maybe @zcorpan knows).

referrer_provided_prefetch_proxies.md
domfarolino: Oops, I'm wrong my bad.

referrer_provided_prefetch_proxies_draft.md
rrliu: Updated the token handling proposal:
- If the rule has only one authorization token in its set:
  - Attach the token to each execution, do not consume the token
- Else if the rule has >1 authorization tokens in its set:
  - Select a token at random from the set and remove it from the set

I'll email Simon to ask about token handling and this explainer generally.

domfarolino (approved):
LGTM to merge now and iterate on the specific token discussion later.

domfarolino:
Alright, we've renamed the main file to `-draft.md` to indicate its draft status, because we'd like to merge it so we have an easy-to-pass-around link for this proposal at [BlinkOn 21](https://www.chromium.org/events/blinkon-21/) next week. @zcorpan we'd still love your feedback, either as comments on this PR after merging, or issues filed or wherever else is easiest for you!

PR #436, opened 2026-07-08
Fix link to "delivery type"

tidoust:
The spec links to the definition of "delivery type" in the "Resource Timing" spec but the fragment ID has changed from `dfn-delivery-type` to `performanceresourcetiming-delivery-type`.

Bikeshed will not auto-link to the definition because that definition is not exported but it still knows about it. This update fixes the link by using a link-default to tell Bikeshed that the link is intended (avoiding the need to hardcod a fragment ID which could perhaps change again).

### DIFF
--- a/prefetch.bs
+++ b/prefetch.bs
@@ -17,6 +17,7 @@ Boilerplate: omit conformance
 </pre>
 <pre class="link-defaults">
 spec:fetch; type:dfn; text:credentials
+spec:resource-timing; type:dfn; text: delivery type
 </pre>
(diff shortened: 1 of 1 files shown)


tunetheweb:
@yoavweiss is this presumably due to the move of Resource Timing in https://github.com/w3c/resource-timing/commit/61c55f4ec1f1618402c898677dc85422974982aa

We’ve had a few of these IIRC and I think we should we fix it there in case there are any other references rather than fixing it here. Do you agree?

PR #438, opened 2026-08-25
Describe nested iframe prerendering behavior

yoichio:
### Summary

This PR updates the [Prerendering cross-origin iframes](prerendering-cross-origin-iframes.md) explainer to specify the behavior and opt-in requirements for nested iframes (iframes embedded within another iframe).

### Details

- **Per-boundary evaluation**: Clarifies that the `Supports-Loading-Mode: prerender-cross-origin-frames` opt-in is evaluated per parent-child frame boundary.
- **Recursive rule**: A nested iframe is only prerendered if all its ancestor frames have been prerendered, and either:
  1. It is **same-origin** relative to its immediate parent frame; or
  2. Its immediate parent frame explicitly includes `Supports-Loading-Mode: prerender-cross-origin-frames` in its HTTP response.
- **Diagram**: Adds a Mermaid diagram illustrating a nested frame hierarchy (`Page A` → `Page B` → `Iframe C` → `D1` / `D2`).
- **Interaction with other features**: Clarifies that `credentialed-prerender` applies only to top-level navigations, and `<fencedframe>` is not prerendered.

PR #440, opened 2026-08-27
Update prerendering spec for nested cross-origin iframes

yoichio:
### Summary

Following the explainer update in #438, this PR updates [prerendering.bs](prerendering.bs) to formally specify the opt-in and deferral behavior for nested iframes during navigational prerendering.

### Details

- **Per-boundary check in `create navigation params by fetching`**: Updates the child navigable navigation check to compare the navigation URL origin against its immediate parent frame's active document origin (`navigable`'s `parent`'s `active document`'s `origin`). If cross-origin, it checks the parent document's `allow cross origin iframes navigation while prerendering` flag rather than the top-level traversable.
- **Recursive rule note**: Adds an explanatory note clarifying that nested iframe prerendering is evaluated at each parent-child boundary: all ancestor frames must have prerendered, and each child must be same-origin with its immediate parent or opted in via the parent's `Supports-Loading-Mode: prerender-cross-origin-frames` header.

### Related
- Explainer update: #438
- Explainer: [Prerendering cross-origin iframes](prerendering-cross-origin-iframes.md#nested-iframes)

### DIFF
--- a/prerendering.bs
+++ b/prerendering.bs
@@ -753,15 +753,17 @@ Patch the [=navigate=] algorithm to allow the [=prerendering traversable/activat
 
   Append the following steps after the first sub-step under "While true:":
 
-  1. If |navigable| is a [=prerendering navigable=] and <var ignore>currentURL</var>'s [=url/origin=] is not [=/same site=] with |initiatorOrigin|, then:
+  1. If |navigable| is a [=prerendering navigable=], then:
 
-    1. If |navigable| is a [=top-level traversable=], then return null.
+    1. If |navigable| is a [=top-level traversable=] and <var ignore>currentURL</var>'s [=url/origin=] is not [=/same site=] with |initiatorOrigin|, then return null.
 
-    1. Otherwise, if |navigable|'s [=navigable/top-level traversable=]'s [=navigable/active document=]'s [=Document/allow cross origin iframes navigation while prerendering=] is false,  the user agent must wait to continue this algorithm until |navigable|'s [=navigable/loading mode=] becomes "`normal`". At any point during this wait (including immediately), it may instead choose to [=destroy a top-level traversable|destroy=] |navigable|'s [=top-level traversable=] and return null from this algorithm.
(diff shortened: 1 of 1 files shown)


yoichio:
@domfarolino , could you take a look?
We need this change to specify the unclear situation and some of our customers want prerendering nesting.

prerendering.bs
```diff
@@ -753,15 +753,17 @@ Patch the [=navigate=] algorithm to allow the [=prerendering traversable/activat
 
   Append the following steps after the first sub-step under "While true:":
 
-  1. If |navigable| is a [=prerendering navigable=] and <var ignore>currentURL</var>'s [=url/origin=] is not [=/same site=] with |initiatorOrigin|, then:
+  1. If |navigable| is a [=prerendering navigable=], then:
+
+    1. If |navigable| is a [=top-level traversable=] and <var ignore>currentURL</var>'s [=url/origin=] is not [=/same site=] with |initiatorOrigin|, then return null.
 
-    1. If |navigable| is a [=top-level traversable=], then return null.
+    1. Otherwise, if <var ignore>currentURL</var>'s [=url/origin=] is not [=same origin=] with |navigable|'s [=navigable/parent=]'s [=navigable/active document=]'s [=Document/origin=] and |navigable|'s [=navigable/parent=]'s [=navigable/active document=]'s [=Document/allow cross origin iframes navigation while prerendering=] is false, the user agent must wait to continue this algorithm until |navigable|'s [=navigable/loading mode=] becomes "`normal`". At any point during this wait (including immediately), it may instead choose to [=destroy a top-level traversable|destroy=] |navigable|'s [=top-level traversable=] and return null from this algorithm.
```
domfarolino, line 760: We usually don't have explicit "wait here for a while" prose in the middle of algorithm steps. Instead, it's often better to register some continuation task somewhere, and then have the trigger (prerender activation in this case) hunt for the continuation and run it explicitly.

Can you share how this works in the implementation so we can find a cleaner way to map this to the spec? Where do we "suspend" and how do we "resume"?

yoichio: Spec writing:  
"must wait to continue this algorithm until"expression has existed before cross-origin prerendering.
And this "3.2. Navigation fetch changes" section states prerendering behavior by inserting additional algorithm within navigation as it stats "In create navigation params by fetching, add the following step near the top of the algorithm..".
As we see the original [create navigation params by fetching](https://html.spec.whatwg.org/multipage/browsing-the-web.html#create-navigation-params-by-fetching), it uses "While true" and "Wait until" terming.  
Then I think this follows the original navigation spec terming.

Implementation:  
In this section, we'd like to spec the behavior that nested iframe's prerendering will start after starting parent's frame prerendering. If the parent frame starts prererendering, then we start to decide if this nested frame should be prerendered.

BTW, [navigable-loading-mode](https://wicg.github.io/nav-speculation/prerendering.html#navigable-loading-mode) consits of only "default" or "prerender" now, we should use "`default`" instead of "`normal`".

